PR #2, opened 2019-05-22
Bluefors log agent

BrianJKoopman:
This Agent opens and follows the log output created by both the Bluefors LSCI 372 Reader and Valve Control software. This includes any LS372 channel output in R, T, and P, as well as heater values and ranges. It also includes the pressures, flow, valve states and system status. It currently does not include the Errors file, which logs errors from the LS372 in string format. 

This was tested and built against a custom build log simulator on a linux system. However, it will likely have to either run on or interact with a Windows filesystem. As such, some filepath bits will likely need adjustment as it's tested on a Windows system, I just wanted to get this version in its current state committed.


BrianJKoopman:
This has been tested in both the WSL and Docker on Windows. Merging.

PR #5, opened 2019-06-11
Package socs and move Lakeshore drivers to socs/

BrianJKoopman:
This PR makes an installable `socs` package. It follows the SO dev guide suggestion of using versioneer. I've only moved the two Lakeshore drivers, as those are what I'm familiar with (and I need them to be packaged for the thermometer calibration upload script to get fixed). Other drivers can be moved into the package later.

I remove some files in the Lakeshore 240 drivers directory that I don't believe are used any more.


setup.py
```diff
@@ -0,0 +1,10 @@
+from distutils.core import setup
```
jlashner: Right now `python3 setup.py develop` does not work on my system. After looking into it, it seems like the problem can be fixed by switching this line to:
```
from setuptools import setup
```
which works on my local system. Can you try it on your system, and then switch it if it works?

BrianJKoopman: Great, thanks for checking. While I was recommending calling setup.py directly I was actually still installing with pip. I carried over use of distutils from other packages I've put together, though see that setuptools is the preferred tool [1].

I've also added the use of `setuptools.find_packages`, else the import would fail under certain conditions, probably because I wasn't using the "packages" line appropriately in setup.py. I've tested installing in a virtual environment with both setup.py directly and with pip and both now work. (A side effect is I added some packages to the `requirements.txt` file.)

[1] - https://packaging.python.org/guides/tool-recommendations/

jlashner:
Great, just tested it out and its working for me. I think its good to merge.

PR #10, opened 2019-07-13
Lakeshore data acquisition autostart

BrianJKoopman:
This adds automatically starting data acquisition to both the 240 and 372 agents. By default they will initialize and start data acquisition on startup. Users can disable this behavior if they want with a flag or argument configured in their ocs config file.

I've tested with both Lakeshore models at Yale. This likely has some conflicts with the changes in #9, so merging should be handled carefully.

Also adds documentation for both Agents, as well as instructions for first time setup of the 240s, detailing driver installations an pitfalls there.

### DIFF
--- a/agents/lakeshore240/LS240_agent.py
+++ b/agents/lakeshore240/LS240_agent.py
@@ -39,16 +39,17 @@ class LS240_Agent:
 
     # Task functions.
     def init_lakeshore_task(self, session, params=None):
-        """Perform first time setup of the Lakeshore 240 Module.
+        """init_lakeshore_task(params=None)
 
-        Parameters
-        ----------
-        params : dict
-            Parameters dictionary for passing parameters to task. Supports the
-            following keys:
+        Perform first time setup of the Lakeshore 240 Module.
 
-            - 'auto_acquire' (bool, optional: Default is False. Starts data
-              acquisition after initialization if True.
+        Args:
+            params (dict): Parameters dictionary for passing parameters to
+                task.
+
+        Parameters:
+            auto_acquire (bool, optional): Default is False. Starts data
+                acquisition after initialization if True.
 
         """
 
--- a/agents/lakeshore240/LS240_agent.py
+++ b/agents/lakeshore240/LS240_agent.py
@@ -39,8 +39,17 @@ class LS240_Agent:
 
     # Task functions.
     def init_lakeshore_task(self, session, params=None):
-        """
-        Task to initialize Lakeshore 240 Module.
+        """Perform first time setup of the Lakeshore 240 Module.
+
+        Parameters
+        ----------
+        params : dict
+            Parameters dictionary for passing parameters to task. Supports the
+            following keys:
+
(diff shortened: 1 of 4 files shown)


jlashner:
Functionally I think this is good, and it does basically the same thing as I do in #9 so I'm fine with merging this first.

I think we should be more consistent with the documentation though. Most importantly doc strings for tasks and processes. For my tasks, I've been writing the values in the `params` dict as arguments in the docstring so it would appear nicely in the sphinx documentation. I personally like that better than just saying the argument for each task is `params` and then going into more detail about what params we want in the description, especially since we're not including `session` in the documentation.

I also think that we should be using the `argsparse` sphinx module to document the agent command line / site-config arguments, since this is easier to keep up to date.

agents/lakeshore240/LS240_agent.py
```diff
@@ -39,8 +39,17 @@ def __init__(self, agent,
 
     # Task functions.
     def init_lakeshore_task(self, session, params=None):
-        """
-        Task to initialize Lakeshore 240 Module.
+        """Perform first time setup of the Lakeshore 240 Module.
+
+        Parameters
```
jlashner: Why the switch to numpy docstrings?

BrianJKoopman: It's the recommendation in the SO Dev Guide, I've been trying to conform to that with new documentation I write. Consistency is probably more important, but with having such little documentation, I felt like writing anything new should conform to the chosen standard, rather than perpetuating the old one, else we'd never change later.

BrianJKoopman: For consistency I switched the couple of docstrings I wrote to Google's style like most of the rest of the docstrings. Some still remain in the Sphinx style.

BrianJKoopman:
> Functionally I think this is good, and it does basically the same thing as I do in #9 so I'm fine with merging this first.

Alright, yeah, sorry for the conflicts. I worked on this on the plane ride back from the meeting and didn't have a wifi connection.

> I think we should be more consistent with the documentation though. Most importantly doc strings for tasks and processes. For my tasks, I've been writing the values in the `params` dict as arguments in the docstring so it would appear nicely in the sphinx documentation. I personally like that better than just saying the argument for each task is `params` and then going into more detail about what params we want in the description, especially since we're not including `session` in the documentation.

I think putting the values into the "Args" part of the docstring is potentially misleading to users unfamiliar with the structure of an Agent, since the expected behavior then would be to pass things directly, rather than through params. In my recent commit I use "Args" to list params, but then use the "Parameters" [docstring section](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#docstring-sections) to list the elements of 'params'. This doesn't distinguish them in the sphinx page, but does if you're reading the docstring directly.
 
> I also think that we should be using the `argsparse` sphinx module to document the agent command line / site-config arguments, since this is easier to keep up to date.

I followed your example on the 240 and added this for the 372.

PR #14, opened 2019-08-10
Pysmurf file archiver

jlashner:
Here's a first go at the pysmurf file archiver. It's job is to run on a data server, and monitor a smurf_server with one or more pysmurf instances, copying over any files that pysmurf creates and adding the files to a local database. A more detailed description of the table, and where files are copied are in the agent's docstring. 

I have only been able to test it locally because I don't yet have access to a site with multiple hosts. If you see any problems with how it works definitely let me know, otherwise I'll test it with multiple hosts when we're setting up the server at UCSD in a week or so.

If this structure looks ok, I'll go through pysmurf and add `publish` statements wherever files are written so we can start to monitor it. Let me know if you have any problems or suggestions!


jlashner:
I changed the way the archiver works based on Matthew's suggestions. Now the pysmurf_controller creates the table, and adds files to it as soon as it learns about them from the UDP publisher. The pysmurf_archiver then checks the database to see if there are any un-copied files written by one of the target agents, and attempts to copy it over, incrementing a failed copy if something goes wrong.

I still haven't been able to test this on a multi-host network. Does anyone know the best way to test this if I don't have multiple hosts? Do you think I could test it on grumpy's network? Or should I wait until I can test it on the princeton system?

BrianJKoopman:
> I still haven't been able to test this on a multi-host network. Does anyone know the best way to test this if I don't have multiple hosts? Do you think I could test it on grumpy's network? Or should I wait until I can test it on the princeton system?

Actually, since we had the problem with the NUC freezing I temporarily reverted data collection back to being solely on grumpy, so that's no longer a multi-node system, not right now anyway. Before I had deployed a multi-node setup I had tested it in a small set of virtual machines. You could try that if you didn't want to wait for Princeton.

jlashner:
So we should lock the table whenever we want to write or update it? Should agents block until the lock is released, or should there be a main loop where the pysmurf_controller checks every so often to see if the table is free, and then writes all cached files to it?

mhasself:
I don't understand your comments... the agents don't need to lock down the database, I don't think.  The problem is that your API calls might block, regardless, because the database is on the network and because databases are slow sometimes.  You need to be cautious here for the same reason that you need to be cautious when doing disk I/O.  Slack me if we're still not on the same page.

mhasself:
Great, this is much more twisted now.  The only other thing to do (for this nice first version) is to check for success/fail on your self.dbpool.runInteraction calls.  As currently written, your calls will immediately return a Deferred object but the ultimate success or failure of the DB actions will be ignored.  You can capture the result/exception with ```rows = yield self.dbpool.runInteraction(...)``` (which will suspend the thread until the db action completes), or by adding callback + error handlers.

jlashner:
Added some more changes, including putting the table creation into an init task, and adding a command line option to enable/disable it, and adding an errback to the db insertion which prints an error if it fails. I didn't want to use `yield` because I think twisted does some weird stuff with exceptions where they don't always print tracebacks, and they recommend using errbacks instead. 

If this looks good I can rebase and merge.

PR #15, opened 2019-08-12
Replaced job locking with TimeoutLock

jlashner:
This PR replaces job locking in the LS372 agent with TimeoutLock, so people shouldn't get locked out of functions after they fail. Since the fake_data option doesn't really work with all tasks, I haven't been able to test most of the tasks, so if you want maybe you should test them out with hardware at yale. I have tested the acquisition process, and verified that if it raises an exception you can still call other tasks.

### DIFF
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -332,7 +332,6 @@ class LS372_Agent:
 
             # Make sure we aren't servoing too high in temperature.
             if params["temperature"] > 1:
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
I've built a test image on the private registry called `ocs-lakeshore372-agent-dev`. I'm running with it now and have tested acquisition, but no other tasks/processes.

I just looked over your ls240 agent tests and would like to take this opportunity to write something similar for the 372.

PR #19, opened 2019-09-03
Bluefors Agent data feed bugfix

BrianJKoopman:
This fixes a bug in the Bluefors log Agent that causes the aggregator to crash by accidentally passing a string to the OCS Feed. This issue still needs to be address in the HK Aggregator (see https://github.com/simonsobs/ocs/issues/72), but this solves the problem in the Agent.

We also add some things that were useful in debugging, such as logging, the ability to set the frame length via environment variable, and the ability to set the log level via environment variable.

### DIFF
--- a/agents/bluefors/bluefors_log_tracker.py
+++ b/agents/bluefors/bluefors_log_tracker.py
@@ -171,7 +171,7 @@ class LogParser:
             'data': {}
         }
 
-        data['data'][log_name] = data_value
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
This is a pretty important bug fix, and I've tested it on the system at Yale (perhaps the only system setup to actually test this), self-merging so the official Docker image will get built.

PR #21, opened 2019-09-06
Ljsaunders

laurensaunders:
not sure if this was the right place to put this

### DIFF
--- a/socs/smurf_g3_reader/smurf_g3_reader.py
+++ b/socs/smurf_g3_reader/smurf_g3_reader.py
@@ -2,17 +2,22 @@ import so3g
 from spt3g import core
 import numpy as np
 import pickle
-import datetime
+import datetime, time
 import sys, os
 
 def g3_to_array(g3file):
-    '''
+    """
     Takes a G3 file output from the SMuRF archiver and reads to a numpy array.
 
-    Arguments
-    ---------
+    Parameters
+    ----------
     g3file : full path to the G3 file
-    '''
+    
+    Returns
+    -------
+    times : array of G3Time objects
+    data : array of arrays, where each internal array is a SMuRF channel
+    """
     frames = [fr for fr in core.G3File(g3file)]
     
     data=[]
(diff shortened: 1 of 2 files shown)


mhasself:
Great, thanks for this.  W.r.t. where this should live: here is fine for now, if @BrianJKoopman and @jlashner think it's convenient for our tests next week.  We'll probably want to move it into sodetlib or something, once that exists.

A few notes for the next iteration:

1) I see you're looping over individual samples from the loaded frame.  This could be pretty slow.  Please look into the numpy function "hstack" -- it's good for taking lists of lists and assembling them into a single 1-d array.  For example numpy.hstack([ [1,2,3], [4,5] ]) ----> array([1,2,3,4,5]).

2) Can you add a docstring to the top of your function?  It should say what the argument is expected to be (in this case, the name of a g3 file) and what the outputs are.  It's ok to be terse and informal, for now, but if you want to get fancy you can try to make it a docstring in the "google style".

3) The name of your function references "dict", but there's no dict involved yet... you are actually returning arrays.  Perhaps g3_to_array?

4) Looks like this PR includes a "test.txt" file... which we can live without :)  Use "git rm" to remove the file, commit that change, and push to this branch.  Then we'll "squash" when we merge it to master and it will be as though it never existed.

BrianJKoopman:
I think here's fine for now. I'm going to copy the function into an ipynb for the demo. A couple of small comments:

- Can you add the Returns section to your docstring as well? This'll make clear that we're returning the same data structures as the pysmurf function we're replacing.
- The SO dev guide formally recommends the "numpy style" of docstring. We currently have some of both everywhere, which isn't great, but we should use one. "Arguments" with the underlined header mixes the two. Here's an example to work off: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html#example-numpy
- For consistency docstrings should always use """triple double quotes""".

BrianJKoopman:
Some more comments related to functionality:

- Not all frames are going to contain the 'data' key, so we should do some type checking. We'll be looking for `core.G3FrameType.Scan`. This currently raises an exception when it encounters an Observation frame, which will typically be the first frame.
- We should convert the `spt3g.core.G3Time` objects returned in the `times` array to something sensible, like ctimes. This could be accomplished with the datetime module's strptime functionality. Though currently, when opening data generated at Princeton I'm getting some strange timestamps, i.e. `26-Mar--953:02:07:12.547758080`

mhasself:
Before anyone goes down a strptime rabbit hole, I think the best way to convert to unix timestamp is, e.g.:
```
>>> t = core.G3Time('2019-01-02T09:12:12')
>>> t.time / core.G3Units.s
1546420332.0
```

BrianJKoopman:
Closing, we've moved the code to so3g. See https://github.com/simonsobs/so3g/pull/31.

PR #22, opened 2019-09-12
Pysmurf monitor

jlashner:
This PR separates the pysmurf_monitor and pysmurf_controller into  two different agents, so the monitor can be run during non-ocs pysmurf usage. It also separates the `pysmurf_files` database structure and creation to an external script in socs which can be run without using OCS.


mhasself:
The docs aren't required to merge this, I guess...

Also let me know if you want me to start writing that doc -- I like writing...

BrianJKoopman:
> The docs aren't required to merge this, I guess...

It does ensure it won't be forgotten though. It would be great to see before merging.

jlashner:
Alright, I fixed the sql queries and added lots of documentation on how to set up the pysmurf controller, monitor, and archiver. Let me know if it looks good!

PR #23, opened 2019-09-13
Timestream Aggregator Agent improvements

BrianJKoopman:
This PR contains a handful of improvements for the Timestream Aggregator Agent. In no particular order:

- Rotated files based on file duration share same basename with incremented suffix, i.e. `2019-01-01-12-30-00_000.g3`, `2019-01-01-12-30-00_001.g3`, `2019-01-01-12-30-00_002.g3`, etc.
- File basename changed when acquisition stopped and restarted (currently determined by 5 second or larger gap, should eventually be notified by Streamer in some way when the stream starts/stops.)
- Safely startup before connection is available (meaning order of startup between streamer/aggregator does not matter.)
- Use new spt3g G3Reader timeout functionality, enabling the Process call to not block. File is closed and an attempt is made to then reestablish the connection.
- Only starts writing a file if frames are actually coming in.

I've added a written description of the aggregator logic in the docs as well.


agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -24,28 +30,319 @@ def _create_file_path(start_time, data_dir):
...
+        Full filepath of file for output
+
+    """
+    filename = "{}{}.{}".format(basename, suffix, extension)
+    filepath = os.path.join(dirname, filename)
 
     return filepath
 
 
+class FrameRecorder:
```
mhasself: All file system operations and I/O are now isolated in this class; that's great.  I think you should emphasize somewhere that this should not be used in the reactor thread, for that reason.  Similary it would be good to annotate the relevant Process to indicate that it runs as a worker thread.  (Perhaps it is already -- I'm looking at a diff here...)

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -24,28 +30,319 @@ def _create_file_path(start_time, data_dir):
...
+    filename_suffix : int
+        For enumerating files within an acquisition. We want to split files in
+        10 minute intervals, but the names should remain the same with a
+        "_0001" style suffix, which we track with this attribute.
+    dirname : str
+        The directory path currently in use.
+    basename : str
+        The basename of the file currently being used. The actual file will
+        also contain the tracked suffix.
+
```
mhasself: Might be worth adding start_time here, and pointing out that it's the timestamp of the current file, not the whole fileset (i.e. it will only agree with basename when filename_suffix=0).

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -114,40 +410,19 @@ def start_aggregation(self, session, params=None):
...
-            for f in frames:
-                if f.type == core.G3FrameType.Observation:
-                    last_meta = f
-                writer(f)
-                writer.Flush()
-
-            if (time.time() - start_time) > self.time_per_file:
-                writer(core.G3Frame(core.G3FrameType.EndProcessing))
-                writer = None
+            recorder.run()
```
mhasself: I think this loop won't always exit in a timely fashion when self.is_streaming is set False.  Consider the case where no G3NetworkSender is present -- the read_frames() function will block... because _establish_reader_connection() will block.

And _establish_reader_connection() can block in the FrameRecorder constructor, which is equally problematic..

I think the solution is to remove the _establish... call from the constructor -- it will get called on first run(), anyway.  Then rewrite the _establish_reader_connection() and run() functions a little so that run() will always return within 1 second or so (probably 5 seconds sometimes...), even if all it has done is to fail to make a connection.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -114,40 +410,19 @@ def start_aggregation(self, session, params=None):
...
 
         # Once we're done writing close out the file
-        if writer is not None:
-            writer(core.G3Frame(core.G3FrameType.EndProcessing))
-            writer = None
+        if recorder.writer is not None:
+            recorder.close_file()
```
mhasself: Instead of this "invasive" check/call that needs to know internal rules about FrameRecorder, consider defining a FrameRecorder.finalize() function that, by definition, does all the necessary things to properly terminate a run.  Alternately you could put this code into FrameRecorder.__del__ (the destructor), and just do "del recorder" to clean things up explicitly (though the same should happen when recorder goes out of scope, including if there's some kind of exception).

jlashner:
Yep, I pretty much agree with everything Matthew said. I think to prevent the main process from blocking, read_frames and _establish_reader_connection need to return after timing out instead of looping. I think this also requires changing create_new_file to check if frames exist before creating a new file, or else it'll create files with just a metaframe and no data.

Other than that, this looks great and is very well documented.

Can confirm, writing an EndProcessing frame resets the boost stream and does not actually write anything to disk.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -24,28 +30,319 @@ def _create_file_path(start_time, data_dir):
...
+            self.writer = None
+            self.filename_suffix += 1
+
+    def close_file(self):
+        """Close a file. Write the EndProcessing frame, set self.writer to
+        None, rezero suffix.
+
+        """
+        if self.writer is not None:
+            self.writer(core.G3Frame(core.G3FrameType.EndProcessing))
```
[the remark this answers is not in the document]
BrianJKoopman: Jack clarified, but also to quote the spt3g documentation:

> EndProcessing is a special-purpose frame emitted implicitly by G3Pipeline at the end of processing. No further frames will occur after this and reception of an EndProcessing frame is intended as a signal to modules that they should clear any internally buffered data and clean up.
> 
> EndProcessing frames should, in general, contain no data.

So this doesn't write a frame to file. Clarified in docstring of `close_file` and added comments directly about EndProcessing frame writes.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -24,28 +30,319 @@ def _create_file_path(start_time, data_dir):
...
+            self.writer = core.G3Writer(filename=filepath)
+
+            # Write the last metadata frame to the start of the new file
+            if self.last_meta is not None:
+                self.writer(self.last_meta)
+
+    def write_frames_to_file(self):
+        """Write all frames to file.
+
+        Note: Assumes file writer is already instatiated.
```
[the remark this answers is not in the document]
BrianJKoopman: Fixed, thanks!

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Thanks for pointing that, I missed documenting that attribute. Added now.

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: I like the destructor suggestion, so I did that and now explicitly call `del recorder`.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -24,28 +30,319 @@ def _create_file_path(start_time, data_dir):
...
+
+        """
+        # Don't even try to rotate files if writer not active.
+        if self.writer is None:
+            self.log.debug("Writer not active, not checking frame gap.")
+            return
+
+        if self.last_frame_write_time is not None:
+            if (time.time() - self.last_frame_write_time) > gap_size:
+                self.log.debug("Last frame written more than {g} seconds " +
```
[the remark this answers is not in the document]
BrianJKoopman: Good point, changed to report the difference instead.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -24,28 +30,319 @@ def _create_file_path(start_time, data_dir):
...
+
+        Closes out the current file with an EndProcessing frame if the reader
+        timesout or has otherwise lost its connection.
+
+        """
+        if not self.frames:
+            while not self.frames:
+                self.frames = self.reader.Process(None)
+                if self.frames:
+                    continue
```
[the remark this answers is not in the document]
BrianJKoopman: I suppose we save one evaluation of `if self.frames` that way. Changed.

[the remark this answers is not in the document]
mhasself: Let's save another -- the "if not self.frames" is not needed if all the block does is loop "while not self.frames".

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Hmm, right. I'm going to have to give this some thought to avoid breaking desired functionality in other scenarios. If I remove the _establish call in the constructor then the first `self.reader.Process(None)` call in `read_frames()` won't be available. If I establish first in that call then I'd be reestablishing the connection every time I read, which I don't think we want.

Even once we make that non-blocking then `create_new_file` would start making new files even if frames weren't coming in (we were relying on `read_frames()` to block until frames came in to prevent this.) I think most of the other methods need adjustment then too to check if reader is instantiated. Alternatively I suppose the check could happen in run(), where we skip these other things if we were unable to `read_frames()`.

jlashner: This is kind of why I like having the frame reading in a separate thread. If you want to do something like the HKAggregator, you could start a thread that continuously reads frames to a queue and tries to re-establish connections while streaming is true (just make sure that there's no way it can block forever).

Then in the run function you can do something like:
```
if self.frames:
    if writer is not None:
        create new file
    write frames to file

if (time.time() - last_write_time > file_limit):
    close file and set writer to None
```

mhasself: The threads idea is interesting, if you want to go that route.  But that's extra, because you still need to handle doing a clean exit from run() in cases that you can't make a reader or you can't get frames.  (Unless you use Exceptions.)

I propose this minimal fix:
- _establish_reader_connection() tries only once and then returns.
- read_frames() should check self.reader is None, and try (once) to _establish_reader_connection().  If that fails, it returns.  Otherwise, read_frames() should read some frames.  But if that fails, it should clean up (as it does now) and return.
- In run(), after read_frames(), check if self.frames.  If not, return immediately instead of calling the other 4 functions.

Also, two other things:
- Currently you avoid rapid reconnection attempts by sleep(1) in the _establish_reader_connection loop.  If not looping any more, one way to limit the rate of reconnections is to compare the present time to some stored "last_reader_connect_time", and then sleep a bit if that time was too recent.
- In close_file(), the log message "File closed" is either potentially misleading should be in the "if" block.

BrianJKoopman: Thanks, this was helpful, but now I run into the situation where I need to distinguish between a timeout and the streamer being off.

After making the suggested changes I realized that if `read_frames()` checks for `self.reader is None` to reestablish the connection, then we'll want to set `self.reader = None` if the connection times out, so I started to do this if no frames came in. However, this would then cause the connection to break and reestablish if the streamer isn't gone, but simply not sending frames. After several of these iterations the aggregator locks up on a connection reestablishment (is there some limit to the number of connections a G3NetworkSender will make?)

I got caught up prepping a cooldown here, so got to this later than I thought. I'll get back to it in the morning.

jlashner: Do we have to distinguish them? I think that once we have the ability to detect when rogue has actually stopped streaming, and close the connection on the NetworkSender side, we won't need the timeout functionality at all.

mhasself: I have looked into this a bit today... I think it's a real problem.  The G3NetworkSender currently only handles the accepting of new connections, and the termination of closed connections, in Process().  So if no frames are being written, the connection requests (or dead connections) just pile up.

I looked into a few approaches to this but I think the only quick fix is to have the smurf-streamer send, on ~1 second intervals, some kind of keep-alive frame (that will be simply discarded in stream-processor).  The structure of these frames should fit into a general schema for "flow control" frames that will also be used to announce start / stop of data acquisition.

The potential problem with simply closing the Sender down whenever there's no frame data to process is that it means a whole bunch of things need to get set up in the short time after the first couple of frames are received from Rogue.  The Sender needs to set up the server while also queuing frames; the Reader needs to connect quickly enough that those queued frames don't start falling on the floor.  That kind of rush is asking for trouble.

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Added a note about not running in the reactor to the docstring for the `FrameRecorder` class and for the `start_aggregation` process.

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Alright, all aspects of your minimal fix have been addressed @mhasself. Some additional changes were needed once those were made, as we've discussed in Slack. Details are in 
aa1e28a.

We also now handle control frames, but currently only keep alive frames, not start/end frames.

BrianJKoopman:
In addition to addressing the comments made here I've also made some updates to the SMuRF Stream Simulator, which I've been using for testing, including documentation for configuring and running a simulator. This is a bit beyond the initial PR, but I feel is related enough that I've included it.

agents/smurf_stream_simulator/smurf_stream_simulator.py
```diff
@@ -101,48 +105,87 @@ def start_stream(self, session, params=None):
...
 
-            self.writer.Process(f)
-            self.log.info("Writing frame...")
-            frame_num += 1
+        """
+        self.running_in_background = False
+        return True, "Stopping stream"
 
-        return True, "Finished streaming"
+    def start_data_stream(self, session, params=None):
```
mhasself: Docstring should point out that this is a Task.  Same with stop_data_stream.

mhasself:
Actually I'm a little uncomfortable with the Operation naming in the simulator.  Typically a Process has a pair of functions "start_X" and "stop_X".  And tasks do not.  Can we maintain that policy here?  How about:
```
  start_background_streamer
  stop_background_streamer
  set_stream_on
  set_stream_off
```

BrianJKoopman:
Good catch, updated.

PR #25, opened 2019-09-30
Reorganize simulators and containerize ls240 simulator

BrianJKoopman:
This just moves the simulators directory out of `socs/` and creates a Dockerfile for the ls240 simulator. This'll help test containerized agents in a bit more of a self-contained manner.


BrianJKoopman:
Yup, that's what I was doing with the changes here. I'm realizing documentation for running the simulator in a container is missing, I'll add that before merging.

BrianJKoopman:
Added documentation for running the simulator in a container. Merging.

PR #26, opened 2019-09-30
Add heartbeat control task to LS240 agent

BrianJKoopman:
This commit adds a single task to the LS240 Agent for testing the disconnection
of an Agent heartbeat. We use the 240 Agent since it is an easy example that
has an accompanying software simulator.

We also adjust the acq method name to conform to naming conventions and update
its documentation.

Finally, we add txaio logging and a LOGLEVEL environment variable.

### DIFF
--- a/agents/lakeshore240/LS240_agent.py
+++ b/agents/lakeshore240/LS240_agent.py
@@ -2,6 +2,7 @@ import time
 import os
 import argparse
 import warnings
+import txaio
 
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


jlashner:
Is this still something that we want to be testing? Is there any reason that this can't go into the fake-data agent so that we don't mix testing operations with agents that people actually use? Since I don't think we actually want to give people the option to disable the heartbeat of the LS240 agent...

BrianJKoopman:
I'm not sure. This was written back when we were having the problem at Penn where the registry would unregister a bunch of Agents mysteriously. Since we removed the need for the registry it's probably not important any more.

It also would depend on this getting merged, else it can't flip the bit it's trying to flip: https://github.com/simonsobs/ocs/pull/73

That said, the acq method name change and logging stuff would be worth adding still probably.

BrianJKoopman:
@jlashner I moved the heartbeat control task to the FakeDataAgent instead (see https://github.com/simonsobs/ocs/pull/73). Will merge by the end of the day.

PR #27, opened 2019-10-13
updated cryomech agent file

JakeSpisak:
Changes made:
- Set connection to ptc to remain open from the initialization task, and only closed when the PTC instance is destroyed, rather than opened and closed every time data was published to the feed. 
- Added doc strings
- Removed debugging print statements


agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
```diff
@@ -35,7 +35,73 @@ def signal_handler(signum, frame):
...
+                  "Motor Current": [58, 57, 60, 59]}
         
+        #Iterate through all keys and return the data in a usable format
+        data = {}
+        for key in keyloc.keys():
+            locs = keyloc[key] 
+            wkrBytes = bytes([rawdata[loc] for loc in locs])
+            
+            #three different data formats to unpack
+            if key in list(keyloc.keys())[0:2]:
```
jlashner: dict keys are not ordered, so this is an unreliable way of selecting keys. Instead, I would do something like:
```
if key in ["Coolant In", "Coolant Out", "Oil"]:
   do something
```

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
```diff
@@ -35,7 +35,73 @@ def signal_handler(signum, frame):
...
+        return query
+
+    def breakdownReplyData(self, rawdata):
+        """
+        Take in raw ptc data, and return a dictionary. The dictionary keys are the data labels, 
+        the dictionary values are the data in floats or ints. 
+        """
+        
+        #Associations between keys and their location in rawData
+        keyloc = {"Operating State": [9,10], "Pump State": [11, 12], "Warnings": [15, 16, 13, 14], "Alarms":[19, 20, 17, 18],
```
jlashner: I think using struct.unpack for the whole data blob seems simpler than extracting the bytes in this way. There's an example [here](https://docs.python.org/3.7/library/struct.html#examples) on how to extract bytes into a namedtuple.

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
```diff
@@ -35,7 +35,73 @@ def signal_handler(signum, frame):
...
+        Gets the raw data from the ptc and returns it in a usable format. 
+        """
+        self.comm.sendall(self.buildRegistersQuery()) 
+        data = self.comm.recv(1024)
+        brd = self.breakdownReplyData(data)
+            
+        return brd    
+    
+    def buildRegistersQuery(self):
+        query = bytes([0x09, 0x99,  # Message ID
```
jlashner: I think using [struct.pack](https://docs.python.org/3.7/library/struct.html#struct.pack) here is clearer and more maintainable than setting each individual byte manually.

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Made this change

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Didn't make this change after mutual agreement; not knowing the cryomech byte order made this difficult

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Didn't make this change after mutual agreement

PR #28, opened 2019-11-05
Add delay in data collection to account for pause

BrianJKoopman:
The 372 inputs have a user set parameter called "change pause", which has a
minimum of 3 seconds. This is time after switching channels that the 372 waits
before collecting measurements. If the channel is queried in this time it'll
report the same value repeatedly. This means we end up collecting 3 seconds 
(or whatever the pause value is set to) of identical, stale, data points. This patch
addresses that by tracking when channels change, querying the pause setting
and sleeping for that time.

There's also some other minor cleanup that were simple things I noticed while
linting.

### DIFF
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -2,15 +2,19 @@ import os
 import random
 import argparse
 import time
-import threading
 import numpy as np
+import txaio
 
 from socs.Lakeshore.Lakeshore372 import LS372
 
-on_rtd = os.environ.get('READTHEDOCS') == 'True'
-if not on_rtd:
+ON_RTD = os.environ.get('READTHEDOCS') == 'True'
+if not ON_RTD:
     from ocs import ocs_agent, site_config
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
There's been one more related feature request, which I hope to get to by the end of the day. So maybe if you haven't looked at this yet @jlashner you could wait for that commit.

BrianJKoopman:
Alright @jlashner, all set for review. I added a flag to allow users to set a wait time after the initial pause before data is collected. Penn requested that feature, as they want to avoid even seeing the transient spike that occurs after a channel change. Since this has the potential to drop all (well most, given the way it's written, I think we'd still get a single data point...) of the data on the floor if set improperly (i.e. longer than the actual channel dwell time), I've included a check on the channel dwell time and will limit the delay such that we at least get a second of data at the end of a channel scan.

BrianJKoopman:
Going to merge this so it can be used at Penn. It's been tested on the cryostat at Yale.

PR #33, opened 2019-12-09
Update agent.py file to handle bad compressor output strings

jseibert575:
If the compressor spits out a bad string, the agent skips that data block and prints the output into logs


agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
```diff
@@ -78,24 +78,33 @@ def breakdownReplyData(self, rawdata):
                   "High Pressure": [46,45,48,47], "High Pressure Average": [50, 49, 52, 51], "Delta Pressure": [54, 53, 56, 55],
```
mhasself: Please reformat so that lines are less than 78 columns in less absolutely necessary.  That includes comments, and string literals, which you can break into multiple lines like:
```
print(f"value of a is {a} and "
      f"value of b is {b}")
```

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
```diff
@@ -78,24 +78,33 @@ def breakdownReplyData(self, rawdata):
...
+                locs = keyloc[key] 
+                wkrBytes = bytes([rawdata[loc] for loc in locs])
+
+                #three different data formats to unpack
+                if key in ["Operating State", "Pump State"]:
+                    state = int.from_bytes(wkrBytes, byteorder='big')
+                    data[key] = state
+
+                if key in ["Warnings", "Alarms"]:
+                    data[key] = int(''.join('{:02x}'.format(x) for x in wkrBytes))
```
mhasself: You need to pass a radix of 16 to the `int` cast, so it decodes the string as hexadecimal.  Currently this will either crash (if the string contains a-f) or decode incorrectly (because it's interpreting digits as decimal).  Just add a second argument `int(..., 16)`.

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
```diff
@@ -175,13 +184,15 @@ def start_acq(self, session, params=None):
...
             self.take_data = True
             
-            #Publish data, waiting 1/f_sample seconds in between calls. 
+            #Publish data, waiting 1/f_sample seconds in between calls.
             while self.take_data:
                 pub_data = {'timestamp':time.time(), 'block_name':'ptc_status'}
-                data = self.ptc.get_data()
+                data_flag, data = self.ptc.get_data()
                 pub_data['data'] = data
                 time.sleep(1./self.f_sample)
```
mhasself: The sleep is before the publish... it would be better to have it after, right?

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Got it, I reduced the column size in those lines and a few others as well

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Ah, this is probably the source of the crashes we have been seeing!  When there is a warning, the string is something like this: 'c9000000', which currently the system can't handle. Cryomech directly gave us this code, so I'll forward this correction to them after verifying this is really the solution.

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Yep it's cleaner that way, I made the change

JakeSpisak:
Per Brian's request, I also added the folder I used to test that the agent properly handles bad strings. I assume that shouldn't be released to master

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
BrianJKoopman: This should include comments. Recent commits look good except for one comment which is wrapped at 128 characters. (line 83)

BrianJKoopman:
> Per Brian's request, I also added the folder I used to test that the agent properly handles bad strings. I assume that shouldn't be released to master

That's great to have, thanks for adding it. I think it's fine to put hardware specific tests in master, but ideally it would:
* be written to import the code they're testing, rather than duplicate that code
* not depend on jupyter notebooks
* be structured for running via a test runner like pytest (see `tests/hardware/test_ls372.py`)
* live in `tests/hardware/`

agents/cryomech-cpa1114/cryomech_cpa1114_agent.py
JakeSpisak: Fixed in the latest commit

JakeSpisak:
> > Per Brian's request, I also added the folder I used to test that the agent properly handles bad strings. I assume that shouldn't be released to master
> 
> That's great to have, thanks for adding it. I think it's fine to put hardware specific tests in master, but ideally it would:
> 
> * be written to import the code they're testing, rather than duplicate that code
> * not depend on jupyter notebooks
> * be structured for running via a test runner like pytest (see `tests/hardware/test_ls372.py`)
> * live in `tests/hardware/`

I implemented this in the latest commit. Now there is a script in tests/hardware that starts data acquisition, with instructions to add a 'fake-errors' argument to the site config compressor block. Then the agent injects fake errors in the form of bad strings 50% of the time, which you can see printed to logs.

BrianJKoopman:
Great, thanks for the changes!

PR #37, opened 2020-01-27
Added docs on how to setup .ssh permissions for pysmurf archiver

jlashner:
Not much here to review, but I thought this would be a good place to discuss/finalize how to setup ssh permissions for the pysmurf-archiver so it can copy over files from the smurf-server. The docs describe how I currently set it up at UCSD, and it seems to work and not require much effort. If you have any problems with it let me know, otherwise I'll go ahead and merge it and use this method whenever I set up the archiver.

### DIFF
--- a/docs/agents/pysmurf/pysmurf-archiver.rst
+++ b/docs/agents/pysmurf/pysmurf-archiver.rst
@@ -38,7 +38,8 @@ Example site-config entry::
 
 Setting up SSH Permissions
 --------------------------
-This still needs work...
+For instructions on how to setup ssh-permissions for the pysmurf-archiver,
+see the following SO-wiki page: http://simonsobservatory.wikidot.com/daq:smurf-ssh-permissions
 
 Docker Configuration
 --------------------
@@ -55,6 +56,7 @@ The docker-compose entry is similar to that of the pysmurf-monitor. For example:
             MYSQL_PASSWORD: ${DB_PW}
         volumes:
             - ${OCS_CONFIG_DIR}:/config
+            - /home/ocs:/home/ocs
             - /data:/data
         depends_on:
             - "sisock-crossbar"


mhasself:
I have a few comments on this:
- In the pysmurf-archiver docs, it's not clear where that Agent needs to run.  Close to the storage or close to the pysmurf?  I have figured out that it must be close to the storage...
- There are security implications for using ssh keys without password protection.  That needs to be mentioned.  Under what circumstances should a user be more cautious about this?
- Mention: If a single archiver is used for multiple smurf-servers, the same ssh key can/should be re-used.
- Is OCS_CONFIG_DIR the directory that we commit to ocs-site-configs?  If so then there's a greater risk that people will add and commit their ssh keys to the repo.  Normally one does not want to do that.  What's our stance on this?
- Am I alone in finding it annoying when directories are hidden, when they don't need to be?  The smurf-server directory could be called "dot_ssh", or "ssh_config", and mounted into .ssh in the docker.  This is a minor gripe, but is related to the previous item.

BrianJKoopman:
> In the pysmurf-archiver docs, it's not clear where that Agent needs to run.  Close to the storage or close to the pysmurf?  I have figured out that it must be close to the storage...

I'm going to second this. A diagram in the docs might help? I thought I knew where it should run, but am confused by the `ssh-copy-id` command indicating the smurf-user and smurf-hostname. I thought it was moving files off the smurf-server?

> There are security implications for using ssh keys without password protection.  That needs to be mentioned.  Under what circumstances should a user be more cautious about this?

Definitely this. Is smurf-user in the docker group? If so, we probably should even change the user on the storage end to a more restrictive one.

> Is OCS_CONFIG_DIR the directory that we commit to ocs-site-configs?  If so then there's a greater risk that people will add and commit their ssh keys to the repo.  Normally one does not want to do that.  What's our stance on this?

Yeah, it's in ocs-site-configs. Can we just add `.ssh/*` to `.gitignore`?

> Am I alone in finding it annoying when directories are hidden, when they don't need to be?  The smurf-server directory could be called "dot_ssh", or "ssh_config", and mounted into .ssh in the docker.  This is a minor gripe, but is related to the previous item.

I'm more in favor of `ssh_config/` or even just `ssh/`. I don't find it bothersome, but not having it hidden will make it clearer for users probably.

### Misc
* Usually systems are unhappy if `.ssh/` doesn't have 700 permissions, is this addressed?

jlashner:
All really good feedback

- The archiver does run close to storage. I'll put that into the documentation. The reason the ssh-key is needed in the first place is so that rsync can copy files from the smurf-server without password verification each time. This requires the OCS user on the storage node to have an ssh-id that is put in the smurf-server's `authorized users` file.

- If you password encrypt the ssh key, do you have to enter that password whenever you use it? cryo:smurf does have docker permissions on the smurf-srv. What do you mean by change the storage user to be more specific? Does this mean to replace the `ocs` user with a new user? Could this be solved by setting the ssh dir permissions to 700, and not allowing the `ocs` group access to the file?

- I support renaming the directory to `ssh_config` and adding it to the git ignore. Do you think this is safe enough or do you think people will still find a way to put it on github?

- UCSD's ssh file does not have 700 permissions but that should be changed. I'll add it to the docs.

BrianJKoopman:
> If you password encrypt the ssh key, do you have to enter that password whenever you use it? cryo:smurf does have docker permissions on the smurf-srv. What do you mean by change the storage user to be more specific? Does this mean to replace the `ocs` user with a new user? Could this be solved by setting the ssh dir permissions to 700, and not allowing the `ocs` group access to the file?

Yeah, unless a keyring is used. But then you'd need to unlock the keyring...

My concern is that a compromised key then effectively gives root access on the smurf-srv, since an account with docker permissions is root equivalent. I meant only that a different user with permissions to read the files you want to copy could be used (i.e. one that doesn't have docker permissions, so that a compromised key, while bad, isn't as bad...)

> I support renaming the directory to `ssh_config` and adding it to the git ignore. Do you think this is safe enough or do you think people will still find a way to put it on github?

Not sure, people could always force add it. We should maybe keep a closer eye on that repo if we want to track this.

jlashner:
It seems like if we're really concerned about this it makes the most sense to just remove the cryo user from the docker group. I think people (including myself) are already using ssh keys to connect to smurf-servers...

jlashner:
Merging cause matthew said so.

PR #38, opened 2020-01-30
Organize and document cryomech Agent

BrianJKoopman:
Added an old page for documentation that I had started working on a while ago.
I also move the agent directory, this is so that sphinx can import the code for
auto-documentation purposes. We should avoid '-' in directory names for this
reason.

I also linted the cryomech agent, there were a lot of PEP08 related spacing
issues.

@JakeSpisak can you add any documentation you might find helpful to the .rst 
file I've started here?


docs/agents/cryomech_cpa1114.rst
```diff
@@ -0,0 +1,68 @@
+.. highlight:: rst
+
+.. _cryomech_cpa1114:
+
+======================
+Cryomech CPA1114 Agent
```
JakeSpisak: Should we consider renaming this before it gets distributed too widely? It works fine with our CPA1110, so the name should probably be more general. It's a minor thing, so maybe not worth the hassle.

BrianJKoopman: Yeah, I sort of agree it should be more generally named. Are the CPA1114 and CPA1110 the only compressors in SO?

Renaming isn't too bad within the repo/ocs. Renaming the dockerhub image doesn't seem possible. However a new one can be created and the old one can be updated to explain that there's a newer image elsewhere (and eventually deleted.)

JakeSpisak: Hmm, I'm not sure about the compressors used in SO. I'll ping the cryo channel. The old/new image names sounds a bit confusing, but if the image is really only in use at a few institutions right now maybe it won't be a huge issue.

docs/agents/cryomech_cpa1114.rst
```diff
@@ -0,0 +1,68 @@
...
+ethernet to monitor the health of the unit, including stats such as Helium
+temperature and pressure, oil temperature, and more. Control is not yet
+implemented.
+
+.. argparse::
+    :filename: ../agents/cryomech_cpa1114/cryomech_cpa1114_agent.py
+    :func: make_parser
+    :prog: python3 crypmech_cpa1114_agent.py
+
+Description
```
JakeSpisak: Was there supposed to be text underneath this header?

BrianJKoopman: Yeah, I think I copied the header structure from the [timestream aggregator page](https://socs.readthedocs.io/en/latest/agents/timestream_aggregator.html). If there isn't any description to add beyond the initial one we could probably remove this header (no other pages have it.)

JakeSpisak:
I added an example client to collect data acq: I like having these in there so that people new to ocs can just copy-paste and modify setup-specific stuff like the instance ID.

BrianJKoopman:
Thanks for the changes @JakeSpisak. Merging now.

PR #40, opened 2020-02-06
Fixed pysmurf monitor numpy bug

jlashner:
Quick fix to make sure numpy files have correct extension. This should fix bug with copying numpy files over.


jlashner:
If nobody has any objections I'll merge this this weekend

PR #43, opened 2020-02-16
modified archiver path creation

jlashner:
I mostly wanted to start this PR so we can decide if we want to change the pysmurf_archiver path based on what we talked about 1.5 weeks ago before the pton meeting. Based on that discussion, the path should be something like 
```
data_dir/<5 ctime digits>/<pub_id>/<time_action>/outputs/fname
```
or
```
data_dir/<5 ctime digits>/<pub_id>/<time_action>/plots/fname
```

For example: `/data/pysmurf/15058/crate1_slot2/1580857504_find_freq/outputs/1580857504_amp_sweep_resp.txt`

The only thing I'm not quite sure about is how to get the <time_action> value. The time I can almost always get from the first part of the filename, and then they will all be grouped together, but the only thing I can think of for the action is submitting the function name whenever a new file is registered. Can anyone think of a better idea for this? If that's the case then it might be hard to have this by the pton meeting since it'll require a pysmurf pull-request.

In this branch there's kind of a stop-gap that has that structure, but without the <time_action> section that we can use for now, or we can just stick with the old format for this meeting and mention we want to change it soon after. Any thoughts?

### DIFF
--- a/agents/pysmurf_archiver/pysmurf_archiver_agent.py
+++ b/agents/pysmurf_archiver/pysmurf_archiver_agent.py
@@ -21,11 +21,13 @@ def create_local_path(file, data_dir):
 
         The file path will be:
 
-            data_dir/<5 ctime digits>/<file_type>/<file_name>
+            data_dir/<5 ctime digits>/<pub_id>/<time_action>/outputs/fname
 
-        E.g.
+        or 
 
-            /data/pysmurf/15647/tuning/1564799250_tuning_b1.txt
+            data_dir/<5 ctime digits>/<pub_id>/<time_action>/plots/fname
+        
+        depending on whether the file is a plot
 
         Arguments
         ---------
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Are types matched to only a single action? Could they be used to determine the action that generated them? (I may have answered my own question. All `register_file` calls have types that match to only one action -- if action is just iv, noise, tune, util -- except for 'daq' and 'adc'.)

That's the only idea I have right now, and I do think it would only serve as a stop-gap. I agree adding the action to the file registration is probably the right way forward.

mhasself:
I think @BrianJKoopman has the right idea ... the 'action' should be part of the announcement, going forward.  But if the 'action' is missing, for now, figure it out from type using a hard-coded map ('amp_sweep' -> 'find_freq' and so on).  We'll remove the map later, when it's no longer needed.

I think it would be good to have this for the meeting, because it's a key part of how the data will look in the archive.  And it should be 100% forwards compatible.

jlashner:
That sounds good, right now the types are semi-random since I didn't really know what was important when I made them, but I think if we assign them in a smarter way (or even just make them the same as the name of the function that they're in) then we can use them as the action tag. If we want something that can handle files created in functions being called by other functions (i.e. the `find_freqs` function is also called by `S.tune`) then maybe we can pass the action string through as an argument, but that might be unnecessary.

For now though I'll just create a map from the type to the action name and get the group timestamp from the filename.

jlashner:
Alright, put in a map from file "type" to corresponding "action" which should be used in path. Some types were in multiple functions, so I had to choose something that made sense. I also decided to group similar types together, such as all the noise measurements. Let me know if this looks reasonable or if you have any suggestions.

PR #46, opened 2020-02-24
Add flow control handling to timestream-aggregator

BrianJKoopman:
This PR introduces flow control frame handling to the Timestream Aggregator. The `run` function is mostly unchanged, the only differences being that if we see a "START" frame, we close the currently file and create a new one, and if we see an "END" frame, we close the current file.

All flow control frames (frames with type `core.G3FrameType.none`) are dropped, so that we don't write them to file.

As a way to test the timestream aggregator without a full system we also add the flow control frames to the smurf stream simulator, however the current implementation isn't perfect, as a frame could still be send after an "END" frame.

I'm hoping to write some tests for this before merging, but if you see anything obviously wrong with the logic, please comment.


BrianJKoopman:
Alright, tests are going to take some time, since some testing infrastructure still needs to be setup for this repo. It's on the todo list, but probably shouldn't hold up this PR, since the timestream aggregator should start seeing use at test institutions.

I did add a catch for rapidly started files. This would cause a naming collision if files were started within a second of each other, which is currently happening due to the Observation Frames. I started incrementing filename suffix if the last file timestamp matches a new one (to within a truncated second.) This'll ensure we don't overwrite files. We should handle the Observation Frames in some way though...

Also updated the docs to discuss the flow control logic.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -243,6 +248,18 @@ def read_frames(self):
 
         self.frames = self.reader.Process(None)
         if self.frames:
+            # Handle flow control frames
+            for f in self.frames:
+                if f.type in [core.G3FrameType.none]:
+                    # START; create_new_file
+                    if f.get('sostream_flowcontrol') == 1:
```
mhasself: Please define constants for these flow control values (0,1,2,3), and use those instead of literals.  In the simulator, too.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -243,6 +248,18 @@ def read_frames(self):
...
+                        self.close_file()
+                        self.create_new_file()
+
+                    # END; close_file
+                    if f.get('sostream_flowcontrol') == 2:
+                        self.close_file()
+
             # Discard flow control frames
             self.frames = [x for x in self.frames if x.type != core.G3FrameType.none]
```
mhasself: So all "none"-type frames are flow control?  Then we should print an error or warning if a badly formatted "none" frame is encountered (I.e. one with no sostream_flowcontrol information).

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Mhm, great suggestion! Defined an Enum for it in both.

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Yeah, I believe right now all "none" type frames are for flow control. I added a warning to the logs if one doesn't contain the flowcontrol key.

BrianJKoopman:
I want to note that this still doesn't handle the new CLEANSE control frames. Maybe wait to re-review until I've added those.

agents/smurf_stream_simulator/smurf_stream_simulator.py
```diff
@@ -153,17 +167,36 @@ def start_background_streamer(self, session, params=None):
...
+            self.writer.Process(f)
+
     def stop_background_streamer(self, session, params=None):
         """stop_background_streamer(params=None)
 
         Stop method associated with start_background_streamer process.
 
         """
+        if self.is_streaming:
+            self._send_end_flowcontrol_frame()
```
jlashner: Is this run in a separate thread than the "start_background_streamer" function? If so it seems its possible to send frames in the wrong order here. Like if you call `_send_end_flowcontrol_frame` before the Process call on line 161 then you'll get end_frame->dataframe->exit. Not sure if it matters so much but might be worth trying to keep the G3NetworkSender only writing from a single thread.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -243,6 +255,24 @@ def read_frames(self):
...
+            for f in self.frames:
+                if f.type in [core.G3FrameType.none]:
+                    if 'sostream_flowcontrol' not in f:
+                        self.log.warn("Improperly formatted flow control " +
+                                      "frame encountered.")
+
+                    flow = f.get('sostream_flowcontrol')
+
+                    # START; create_new_file
+                    if FlowControl(flow) is FlowControl.START:
```
jlashner: Won't this break here if f.get('sostream_flowcontrol') returned None? Probably don't have to use a get call, since it would break anyway if the key doesn't exist.

jlashner:
I think the way that you are adding start and stop frames in the smurf_stream_simulator can lead to data frames between stop and start frames. Since you are now publishing `self.writer` from multiple threads, it's possible that the 'stop' function is called and the stop frame is sent before the data frame is processed in the `running_in_background`  loop. I think to be accurate we should only write frames from a single thread to prevent these kinds of issues. 

Also the real data streamer won't always be sending ALIVE frames while it's streaming, only when no data exists. Now might be a good time to make the simulator behave the same way if that's something we care about.

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Ah, yeah, good catch. I added a continue after the check and warning about improperly formatted flow control frames.

agents/smurf_stream_simulator/smurf_stream_simulator.py
BrianJKoopman: Yeah, definitely it's already sending frames in the wrong order. One frame typically makes it after an "end" frame, as you pointed out. It worked quickly for testing prior to the meeting, but I agree it should be fixed.

BrianJKoopman: I've updated the simulator so that all write operations happen from the main streamer process, all the frame sending should happen from a single thread now.

BrianJKoopman:
I've made the following changes:
* timestream-aggregator now handles CLEANSE frames
* Stream simulator does all write operations from a single thread
* Fixed a bug in the timestream-aggregator that would cause the 1-second files at the start of an acquisition under certain conditions (see 10fca45)
* Added the ability to force the sending of START frames from the simulator
* Added CLEANSE wiring and observation frames after END frames

I'm still a little unsure about the CLEANSE frames. Am I sending them in the right place? After END frames?

I ask because I'm seeing behavior I wouldn't have expected, but maybe I misunderstand the resent Observation frames on reconnect. What I'm seeing is that I'll start the simulator and aggregator, it'll start streaming and a file will start writing. I then stop the data stream by sending an END frame and both CLEANSE frames. I then restart the simulator container, which brings down and back up the G3NetworkSender. Frames start coming in again. The new file that starts writing will have two Observation frames at the start, the first one with a time matching the one I got at the start of the previous file, the second one being at the current time.
```
$ python3 -i /home/koopman/data/open.py 2020-02-28-20-05-51_000.g3 
>>> frames[0]['start_time']
1582920350.297343
>>> 
$ python3 -i /home/koopman/data/open.py 2020-02-28-20-06-43_000.g3 
>>> frames[0]['start_time']
1582920350.297343
>>> frames[1]['start_time']
1582920401.9369667
```

Maybe this is because the writer doesn't get closed? I'm just stopping the flow of data, but keeping the G3NetworkSender active. I still would have thought the cached "most recent" Observation frame would have been the empty CLEANSE frame.

BrianJKoopman:
Alright, as @jlashner pointed out in Slack the issue was related to the cached last_meta frame. Addressed in fcad583. I think this settles out the timestream aggregator work I had left. I still want to add tests, and will, but it's going to take some time. I think if people are up for reviewing this again now, that'd be good.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -305,10 +353,10 @@ def write_frames_to_file(self):
         for f in self.frames:
             # Do not record flowcontrol frames
             if f.type == core.G3FrameType.none:
-                self.log.warn("Received flow control frame with value {v}." +
+                self.log.warn("Received flow control frame with value {v}. " +
```
mhasself: Just flagging this as another place where frame type is used to determine whether a frame is flow control or not.

agents/timestream_aggregator/timestream_aggregator.py
```diff
@@ -217,10 +229,37 @@ def read_frames(self):
...
+
         self.frames = self.reader.Process(None)
         if self.frames:
-            # Discard flow control frames
-            self.frames = [x for x in self.frames if x.type != core.G3FrameType.none]
+            # Handle flow control frames
+            for f in self.frames:
+                if f.type in flowcontrol_types:
+                    if 'sostream_flowcontrol' not in f:
+                        self.log.warn("Improperly formatted flow control " +
```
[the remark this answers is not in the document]
BrianJKoopman: A standard observation frame wouldn't get discarded since the discard only occurs if 'sostream_flowcontrol' is in the frame. I was just working on this though, since that would mean none type frames that are poorly formatted would make it into the timestream, which we don't want, right? I was going to also add in the previous line that dropped flowcontrol frames:
```
self.frames = [x for x in self.frames if x.type != core.G3FrameType.none]
```

It is a good point that standard Observation and Wiring frames would generate this warning though.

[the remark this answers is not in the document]
BrianJKoopman: I think some type checking will still be required if we want to warn about poorly formatted none type frames, as suggested earlier: https://github.com/simonsobs/socs/pull/46#discussion_r383522174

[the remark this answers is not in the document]
mhasself: Right, I mis-read the handling of standard Obs+Wiring.

My suggestion in comment #46 was predicated on a question I posed to which I sort of hoped the answer was "no".  My suggestion is: since we now can't use type as the sole means of separating flowcontrol from non-flowcontrol frames (because of CLEANSE), then perhaps we should use the 'sostream_flowcontrol' key as a single criterion.  This would simplify all the logic.  Would it mean that we now consider None-type frames without sostream_flowcontrol to be valid data that should be written into the output stream?  Yes.  Since we don't currently use those, this has no real effect.  But it leaves us the option of using None frames in our real data stream, later.  And more usefully, eliminating the flowcontrol_types list also allows us to add CLEANSE frames of other types, later, without modifying the smurf-recorder.  In case we some day realize we need to incorporate, and thus also to CLEANSE, InstrumentStatus or Calibration or Housekeeping type frames.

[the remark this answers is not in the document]
mhasself: But... I blocked this merge based on my mis-reading of Obs+Wiring handling.  So I'll "approve" and leave the above as a suggestion / opportunity to say "I knew this would happen!" in 18 months :)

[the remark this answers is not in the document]
BrianJKoopman: Heh, well I took your advice and removed the type checking and check only for the 'sostream_flowcontrol' key.

I fixed it over here, https://github.com/simonsobs/socs/pull/50/commits/9d8b74c536ca5d70be11ab6bf86eb595fd0b5a18, since the tests are present there and that involved moving the location of the code in question (so working on this branch would cause some difficulties.) I'm going to merge that in to this branch now anyway. I kept it separate so that the diff here wouldn't change dramatically while people were reviewing.

agents/timestream_aggregator/timestream_aggregator.py
BrianJKoopman: Removed the type check here, https://github.com/simonsobs/socs/pull/50/commits/9d8b74c536ca5d70be11ab6bf86eb595fd0b5a18, and now check for the flow control key instead. Even though this extra check shouldn't be necessary, it's been helpful to have.

BrianJKoopman:
Thanks for all the helpful feedback! Merging now. Of course, feel free to comment/open new issues for related items.

PR #48, opened 2020-02-25
Update basename convention to use ctime

BrianJKoopman:
Similar to https://github.com/simonsobs/ocs/pull/110, we discussed changing the naming convention of .g3 files to use ctime, rather than the `"%Y-%m-%d-%H-%M-%S"` format. This PR changes the naming scheme in the timestream aggregator.

Edit: Just to be more clear, this means that files will look like "ctime_000.g3", "ctime_001.g3", etc. for sets of observations, broken up into chunks of a configurable amount of time (defaults to 10 minutes.)


BrianJKoopman:
Talked on the phone about this. No objections to merging, will do so now.

PR #50, opened 2020-03-03
Timestream Aggregator Unit Tests

BrianJKoopman:
I wrote some unit tests for the timestream aggregator. You can run them with:
```
$ python3 -m pytest -p no:wampy --cov-report html --cov socs ./tests/
```

The only method untested is the `run()` method. Some kind of integration testing that tests the flow of this would be ideal, though more complicated to setup. Still not sure the best way to do that. These unit tests can be run automatically on Travis though, within a container, much like tests for so3g and ocs are performed.


BrianJKoopman:
These tests have been very helpful in writing/making changes to the timestream aggregator. Merging into the flow control branch now.

PR #51, opened 2020-03-06
Add test stage to Travis CI Pipeline

BrianJKoopman:
After writing all those tests for the timestream aggregator I wanted to get this running on travis. This setup matches the setup we have for both the ocs and so3g repositories, where we run the tests within a docker container that has all the required dependencies.

The only potentially conflicting change is the small adjustments I made to the Dockerfile. I did this just to be consistent with how the ocs Dockerfile is organized, allowing the commands for running things to remain consistent across repos. Let me know if the changed WORKDIR is going to disrupt anything.

We also add test coverage reporting via coveralls like we have for ocs and so3g.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
   - "3.6"
 
 stages:
+  - test
   - name: dockerize
     if: |
       branch = master AND \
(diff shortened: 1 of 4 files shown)


jlashner:
Looks good! I'm pretty sure changing the workdir is safe, so I'd say go ahead and merge.

PR #53, opened 2020-03-16
Adds pysmurf actions to the pysmurf archiver path

jlashner:
This fixes the archiver path based on what we decided at the pysmurf meeting, using "pysmurf actions" implemented [here](https://github.com/slaclab/pysmurf/pull/317). With this, the OCS archived path looks something like:
```
base/<5 ctime digits>/<smurfpub_id>/<grouptime_action>/<outputs or plots>/filename
```
Where the grouptime is the pysmurf timestamp at which the top-level function is called, and the action name is set in pysmurf (but right now is usually the smurf function name). For example:
```
/data/pysmurf/15844/smurf_emulator/1584400934_take_stream_data/outputs/1584400934.dat
```

This is backwards compatible with versions of pysmurf without the publisher actions, but we should probably wait on this until the pysmurf branch is merged.

### DIFF
--- a/agents/pysmurf_archiver/pysmurf_archiver_agent.py
+++ b/agents/pysmurf_archiver/pysmurf_archiver_agent.py
@@ -22,11 +22,11 @@ def create_local_path(file, data_dir):
 
         The file path will be:
 
-            data_dir/<5 ctime digits>/<file_type>/<file_name>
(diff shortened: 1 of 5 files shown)


jlashner:
Finished resolving conflicts, so this is ready to review and merge! Let me know if you think there's anything else I should add.

jlashner:
Looked over and updated some of the pysmurf archiver and monitor docs. Let me know if you think its ok to merge!

PR #54, opened 2020-03-17
session_data pathway for pysmurf-controller

jlashner:
This PR adds a pathway to communicate data from a running pysmurf script to the pysmurf-controller's session data variable.

If you call `S.pub.publish(data, msgtype="session_data")` for some dict `data`, then they pysmurf-monitor will publish `data` to an ocs_feed. The pysmurf-controller will then receive that data, then update the current session's `session_data` variable. The pysmurf-controller needs to be told which pysmurf-monitor to listen to, which can be done using the `--monitor-id` site-arg.

Since there will only ever be one script running at a time, I took Matthew's suggestion and made the controller's tasks all run in the reactor thread. I've tested it and it seems to work fine, but Matthew if you could look at it and verify that it's correct that would be awesome.

### DIFF
--- a/agents/pysmurf_controller/pysmurf_controller.py
+++ b/agents/pysmurf_controller/pysmurf_controller.py
@@ -1,7 +1,7 @@
 from twisted.internet import reactor, protocol
 from twisted.python.failure import Failure
 from twisted.internet.error import ProcessDone, ProcessTerminated
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import inlineCallbacks, Deferred
 from autobahn.twisted.util import sleep as dsleep
 from twisted.logger import Logger, FileLogObserver
 
@@ -61,10 +61,12 @@ class PysmurfScriptProtocol(protocol.ProcessProtocol):
 
     def processExited(self, status: Failure):
         """Called when process has exited."""
-        if self.log is not None:
-            self.log.info("Process ended: {reason}", reason=status)
 
-        self.end_status = status
+        rc = status.value.exitCode
+        if self.log is not None:
+            self.log.info("Process ended with exit code {rc}", rc=rc)
+
+        self.deferred.callback(rc)
 
 
 class PysmurfController:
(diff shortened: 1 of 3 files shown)


docs/agents/pysmurf/pysmurf-controller.rst
```diff
@@ -115,6 +115,35 @@ OCS client from my host computer::
...
+    controller = MatchedClient('pysmurf-controller', args=[])
+
+    script_path = '/config/scripts/pysmurf/tune.py'
+    controller.run.start(script=script_path))
+
+    ok, msg, sess = controller.run.wait()
+    print(sess['data'])
+
+    >> {'datafile': '/data/smurf_data/20200316/1584401673/outputs/1584402020.dat'}
+
```
mhasself: Can you add a second example here, where actual numerical data is passed instead of just a filename?  I think a strong use case for this feature is to communicate useful stats that are immediately available and informative for debugging / monitoring.  How about {'resonator_peak_count': [490, 340, 134, 230]}?

agents/pysmurf_controller/pysmurf_controller.py
```diff
@@ -93,7 +96,26 @@ def __init__(self, agent, args):
...
+                'observatory.{}.feeds.pysmurf_session_data'.format(args.monitor_id),
+            )
+
+    def _on_session_data(self, _data):
+        data, feed = _data
+
+        if self.current_session is not None:
+            if data['id'] == os.environ.get("SMURFPUB_ID"):
+                if isinstance(data, dict):
+                    self.current_session.data.update(data['payload'])
```
mhasself: There's a timing / race condition issue with current_session.  Currently it's not cleared, which means that any messages coming in will be assigned to the most recent session.  But a rule for session.data is that it must not change once the session is closed (when state -> done).  That could be fixed by clearing current_session before returning from the Task/Process.  But then you might lose messages, if they are issued just before the Task/Process ends.  So I think you also need a short buffer time, after the script has ended but before you close the session, to allow any such messages to propagate into session before you close the session.

docs/agents/pysmurf/pysmurf-controller.rst
mhasself: Another good use case is for progress indicators...  {'status': 'Starting TES bias ramp.'} at one point and then {'status': 'Analyzing bias ramp data.'} later.  These would show up in .status() calls, be useful for keeping an eye on things (in a web widget or client).

agents/pysmurf_controller/pysmurf_controller.py
jlashner: `current_session` is currently closed out at the end of the _run_script function (line 171). I agree with adding a small buffer before closing it though.

jlashner:
Thanks Matthew, I implemented most of your suggestions. I am leaving the file logger in for now because I'm not really sure where else to put it, but if it causes problems I can remove it. Do you have any suggestions how we can log pysmurf stdout to a file otherwise? Should I just always call log functions with "CallInThread"?

mhasself:
> Do you have any suggestions how we can log pysmurf stdout to a file otherwise? Should I just always call log functions with "CallInThread"?

The problem, as I see it, is only with the "open()" call... because I assume the twisted logging functions do all the actual file writes in a reasonably twisted way.

The solution might be the one-liner:
```
   log_file = yield deferToThread(open, log, 'a')
```

PR #69, opened 2020-03-30
lakeshore372: allow Tasks to run while acq Process is running

mhasself:
This is accomplished by adding a lock to guard the acq Process from other
runs of the acq Process, then using the usual job lock to guard only
hardware i/o in the acq Process.

Added some helper functions to return the right ContextManager with an
appropriate timeout... these could instead be provided by OCS.

Also removed misleading working in the lock warnings.

### DIFF
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -4,6 +4,8 @@ import argparse
 import time
 import numpy as np
 import txaio
+import threading
+from contextlib import contextmanager
 
 from socs.Lakeshore.Lakeshore372 import LS372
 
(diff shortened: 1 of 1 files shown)


mhasself:
This is intended to resolve issue https://github.com/simonsobs/ocs/issues/102, for the specific case of the LS372 Agent.

This is tested on Yale's system, but I don't know much about the hardware so someone should confirm that there aren't gotchas in the timing.  The default timeout values are not clear to me, for example... how often, as this is currently written, will the acq Process release the lock?  That should guide the default timeout in self._lock_acquire().

jlashner:
This is my fault for not documenting, but the timeout values in TimeoutLock work as specified here: https://docs.python.org/3.8/library/threading.html#lock-objects.

That is they set how long lock.acquire will block for before giving up, with -1 waiting indefinitely, 0 giving up immediately, and any positive number being the number of seconds before timing out.

So I think the code you have now means that it will drop and re-acquire the lock after every sample, but it'll wait for 10 seconds to acquire a lock before timing out, which is probably not what you want. I'll submit another PR with TimeoutLock documentation.

agents/lakeshore372/LS372_agent.py
```diff
@@ -129,16 +153,24 @@ def start_acq(self, session, params=None):
...
-                    # The 372 reports the last updated measurement repeatedly
-                    # during the "pause change time", this results in several
-                    # stale datapoints being recorded. To get around this we
-                    # query the pause time and skip data collection during it
-                    # if the channel has changed (as it would if autoscan is
-                    # enabled.)
-                    if previous_channel != active_channel:
-                        if previous_channel is not None:
+
+                    with self._lock_acquire('acq', 10.) as acquired:
```
jlashner: I'm not sure about what the behavior is for when multiple threads are waiting for the same lock. Right now there is no sleep time between when the cm drops the lock and attempts to re-acquire it. If another thread is waiting for that lock, are we sure it'll always be able to take it? Or is it random whether the lock will go to the waiting vs the acq process trying to re-acquire? Maybe we should add a sleep statement here just to be sure...

mhasself:
@jlashner  is right on two counts... it drops the lock on every sample, and it's not guaranteed that some thread waiting for the lock will get it if you drop it and immediately reacquire.  I will fix both of those now, standby...

The 10 second time-out for acq thread to reacquire the lock is not that unreasonable.  It wouldn't _always_ wait for 10 seconds -- it would wait for _up to_ 10 seconds.  This should only occur if some other Task does something (with lock held) that takes a long time, or if some client has blasted in a whole bunch of commands that are hogging the lock.  We can protect from the latter case more carefully if that's what's needed (by starting to reject Task start requests completely if the acq thread hasn't sampled recently enough).

mhasself:
@jlashner @BrianJKoopman The latest version only sleeps once per second.  I created a YieldingLock class to support this; we should consider moving this into ocs library.

agents/lakeshore372/LS372_agent.py
```diff
@@ -104,20 +176,35 @@ def init_lakeshore_task(self, session, params=None):
...
-                self.log.warn(f"Could not start init because "
-                              f"{self.lock.job} is already running")
+        with self._acq_proc_lock.acquire_timeout(timeout=0, job='acq') \
+             as acq_acquired, \
+             self._lock.acquire_timeout(job='acq') as acquired:
+            if not acq_acquired:
+                self.log.warn(f"Could not start Process because "
+                              f"{self._acq_proc_lock.job} is already running")
+                return False, "Could not acquire lock"
+            if not acq_acquired:
```
jlashner: should check `acquired` instead of `acq_acquired`

mhasself: good catch... fixed.

mhasself:
Right... this is not a general solution if one wants to manage the order in which threads get activated.  Here all that is guaranteed is that if a thread A releases the lock, some other thread will get to have it before thread A can acquire it again.

Let's think of a more general solution for inclusion in OCS, then.  I have a few ideas.

Regardless, in the present case this doesn't bother me much.  If someone wants to totally reconfigure, they can stop acq and then restart it after.

PR #70, opened 2020-04-02
Added table versioning and failed copy limit

jlashner:
This PR does two things: 

1. It versions the `pysmurf_files` table in the ocs database so that we can make changes to column structure without getting rid of previous data.

2. Puts a limit (of 5) on the number of failed copy attempts before the pysmurf-archiver gives up on copying a file.

This is to solve an issue where pysmurf will write and publish files that don't have unique names, like `mask.txt`. The table versioning was necessary for me to get rid of the Unique requirement on pysmurf file paths in the database. 

I also realized that we are getting an md5sum error sometimes because pysmurf will write and publish a file, and then that file might be changed after the md5sum is already in the database. I'm not sure what the correct thing to do is here for on the OCS side, but the limit on failed copy attempts prevents the archiver from continuously trying to copy this file (and compare its md5sum) forever.

If there are any suggestions on how OCS should handle this situation they are welcome, but I'd like to get some version of this PR merged relatively soon so that the archiver isn't completely broken when this happens. Thanks!

### DIFF
--- a/agents/pysmurf_archiver/pysmurf_archiver_agent.py
+++ b/agents/pysmurf_archiver/pysmurf_archiver_agent.py
@@ -29,8 +29,9 @@ def create_local_path(file, data_dir):
             /data/pysmurf/15647/tuning/1564799250_tuning_b1.txt
 
         In the case of duplicate datafiles being registered, the duplicates
-        will still be copied over but put into a `duplicates` sub_directory
-        with a timestamp before the filename, so that data isn't overwritten.
+        will still be copied over to the location `new_path_name.{i}` where
+        `i` is the next unique index.
+
 
         Arguments
         ---------
@@ -55,17 +56,19 @@ def create_local_path(file, data_dir):
                 f"{file['type']}"
              )
     new_path = os.path.join(subdir, filename)
-    if os.path.exists(new_path):
-        print(f"Warning! Trying to archive duplicate of {new_path}. ")
-        subdir = os.path.join(subdir, 'duplicates')
-        new_path = os.path.join(subdir,
-                                f'{int(time.time())}_{filename}')
-        print(f"Saving duplicate file to {new_path}")
+    unique_path = new_path
+    i = 1
+    while os.path.exists(unique_path):
+        unique_path = new_path + f'.{i}'
+        i += 1
+    if unique_path != new_path:
+        print(f"Warning! Trying to archive duplicate of {new_path}! "
+              f"Will be archived as {unique_path} instead.")
 
     if not os.path.exists(subdir):
         os.makedirs(subdir)
 
-    return new_path
+    return unique_path
 
 
 class PysmurfArchiverAgent:
--- a/agents/pysmurf_archiver/pysmurf_archiver_agent.py
+++ b/agents/pysmurf_archiver/pysmurf_archiver_agent.py
@@ -28,6 +28,10 @@ def create_local_path(file, data_dir):
 
             /data/pysmurf/15647/tuning/1564799250_tuning_b1.txt
 
+        In the case of duplicate datafiles being registered, the duplicates
+        will still be copied over but put into a `duplicates` sub_directory
+        with a timestamp before the filename, so that data isn't overwritten.
+
         Arguments
         ---------
         file: dict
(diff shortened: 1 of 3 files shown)


BrianJKoopman:
> I also realized that we are getting an md5sum error sometimes because pysmurf will write and publish a file, and then that file might be changed after the md5sum is already in the database. I'm not sure what the correct thing to do is here for on the OCS side, but the limit on failed copy attempts prevents the archiver from continuously trying to copy this file (and compare its md5sum) forever.
> 
> If there are any suggestions on how OCS should handle this situation they are welcome, but I'd like to get some version of this PR merged relatively soon so that the archiver isn't completely broken when this happens. Thanks!

What causes a file to change after it's written and published?

jlashner:
On the UCSD system I think it was the fact that they write the channel mask file to the same location every time, something like `/data/smurf2mce_config/mask.txt`: https://github.com/slaclab/pysmurf/blob/046bd8d41b9db9d133e72a0c0735a9311a1e3083/python/pysmurf/client/util/smurf_util.py#L3185

I guess what must've happened was that they wrote the file, the publisher added it to the database, and they re-wrote the file before the archiver copied it over, changing the md5sum from the one stored in the database. The publisher required the path-name to be unique so it didn't allow the new one to be put into the database and then the archiver was stuck in an infinite loop where it kept seeing an uncopied file but whenever it tried to copy it the md5sum was wrong.

Do you have any suggestions for how we should handle something like that? Or are we fine just limiting the failed copy attempts?

BrianJKoopman:
Is each copy of `mask.txt` important? i.e. should we be saving them all? It looks like right now if the md5sums don't match you delete the local file that's different. Am I understanding that right?

It looks like you record the timestamp from when the file was published, can you compare to that to see if the file has been purposefully changed since you recorded the md5sum?

If the path-name has to be unique, does a new `mask.txt` file ever make it into the archiver copy even with the limited failed copy attempts?

jlashner:
Ok, just checked with Ed and it looks like the mask.txt that I'm registering is old and not used by SO even though they must still be updating the file for some reason. They do indeed make a timestamped mask file for each run which I'm not registering but should be, so I'll just add that and remove the registration from the old mask file. 

I still think this PR is necessary and should be merged, but I can probably keep the path uniqueness constraint.

jlashner:
I added a commit addressing @mhasself's concerns about not having a well-defined behavior if there are duplicate files. I've kept it so that duplicate paths are allowed in the files database, but it will be copied to the location `sub_dir/duplicates/<ctime>_<filename>` with a clear warning message about there being duplicate files. This way data won't be over-written, but will still be copied. 

Is there anything else that should be addressed before merging this?

agents/pysmurf_archiver/pysmurf_archiver_agent.py
```diff
@@ -50,11 +54,18 @@ def create_local_path(file, data_dir):
                 f"{str(dt.timestamp()):.5}",
                 f"{file['type']}"
              )
+    new_path = os.path.join(subdir, filename)
+    if os.path.exists(new_path):
+        print(f"Warning! Trying to archive duplicate of {new_path}. ")
+        subdir = os.path.join(subdir, 'duplicates')
+        new_path = os.path.join(subdir,
+                                f'{int(time.time())}_{filename}')
```
mhasself, line 62: You do not check if this exists.  And it's not guaranteed to be a new name.  One could imagine a failure pattern where a given file is registered multiple times in rapid succession (e.g. due to a save+reg being in a loop when it should be after the loop).  These entries would appear in the database together, and be processed together.  Consider starting with index=1, and incrementing index until f'{filename}.{index}' does not exist?

socs/db/pysmurf_files_manager.py
```diff
@@ -71,49 +71,28 @@ def create_table(cur, update=True):
...
-    existing_cols = set([c[0] for c in cur.fetchall()])
-
-    try:
-        for c in columns:
-            if c.name not in existing_cols:
-                cur.execute("ALTER TABLE pysmurf_files ADD {}".format(str(c)))
-                print("Added column {}".format(c.name))
-    except mysql.connector.errors.ProgrammingError as e:
-        print(e)
+        print(f"Found table {table} table and not updating")
```
mhasself, line 85: table table table

jlashner:
Great, thanks for the suggestion, that makes more sense than using a timestamp. Implemented and will merge as soon as checks pass.

PR #72, opened 2020-04-07
Update acq process and add documentation

JakeSpisak:
Add session data to acq process to allow users to query current psu status, and create documentation


docs/agents/keithley-psu.rst
```diff
@@ -1,4 +1,94 @@
...
+.. highlight:: rst
 
+.. _keithly_psu:
+
+==================
+Keithly PSU Agent
+==================
+
+This agent works for many power supplies, including the Keithly 2230G 
+and BK Precision 9130. It connects to the PSU over ethernet, and allows
```
BrianJKoopman, line 10: If this isn't Keithly specific, is there a better, more general name we can rename this Agent to? Perhaps something related to the protocol/interface used to communicate with the PSUs?

agents/keithley2230G-psu/keithley_agent.py
```diff
@@ -70,7 +70,7 @@ def monitor_output(self, session, params=None):
                     data = {
                         'timestamp': time.time(),
                         'block_name': 'output',
-                        'data': {}
+                        'data': data
```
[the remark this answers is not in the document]
JakeSpisak: Good catch, I have no idea why that change was made. Fixed it in the latest commit.

docs/agents/keithley-psu.rst
JakeSpisak: Yeah this is probably a good time to do that: both power supplies run off of the SCPI (standard commands for programmable interfaces) formalism, which is why the agent works for both. How about something like psu_scpi? I'll wait to make the change everywhere until we agree on a name.

BrianJKoopman: That name sounds good, but I'd prefer we flip the words, so scpi_psu, describing the protocol first, then a description of the device it works for.

JakeSpisak: Just did in the latest commit

JakeSpisak:
> Thanks for fixing that nesting problem. I've made a suggestion on name in the inline comments.
> 
> I'm not sure why this is true now, but wasn't before. This is now failing to build due to an import error:
> 
> ```
> $ make html
> Running Sphinx v3.0.3
> building [mo]: targets for 0 po files that are out of date
> building [html]: targets for 20 source files that are out of date
> updating environment: [new config] 20 added, 0 changed, 0 removed
> reading sources... [ 15%] agents/keithley-psu                                                                                                                 
> Exception occurred:
>   File "../agents/keithley2230G-psu/keithley_agent.py", line 5, in <module>
>     from keithley_driver import psuInterface
> ModuleNotFoundError: No module named 'keithley_driver'
> The full traceback has been saved in /tmp/sphinx-err-t7q0mejw.log, if you want to report the issue to the developers.
> Please also report this if it was a user error, so that a better error message can be provided next time.
> A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
> make: *** [Makefile:20: html] Error 2
> ```
> 
> This could be solved my moving the driver code into the actual socs library, and would then match the structure I'm slowly working towards. Please move `keithley_driver.py` into `socs/socs/agent/` and import it from socs like: `from socs.agent.keithley_driver import psuInterface`.

I moved `keithley_driver.py` in the latest commit, as well as its dependency `prologixInterface.py`

JakeSpisak:
Looks like the build failed due to the name change: "build path /home/travis/build/simonsobs/socs/agents/keithley2230G-psu either does not exist, is not accessible, or is not a valid URL."

docs/agents/scpi_psu.rst
```diff
@@ -0,0 +1,95 @@
...
+
+This agent uses Standard Commands for Programmable Instruments (SCPI)
+It works for many power supplies, including the Keithley 2230G 
+and BK Precision 9130. It connects to the PSU over ethernet, and allows
+users to set current, voltage, and turn channels on/off. It also allows for 
+live monitoring of the PSU output. 
+
+.. argparse::
+    :filename: ../agents/scpi_psu/scpi_psu_agent.py
+    :func: make_parser
```
BrianJKoopman, line 17: This is causing a build error in the documentation. This particular line is searching the file specified in `:filename:` for a function called `make_parser`, which doesn't exist. The `make_parser` function should return the parser. You can see an example for how to move your existing parser to a `make_parser` function in several other Agents, for example the cryomech Agent.

docs/agents/scpi_psu.rst
```diff
@@ -0,0 +1,95 @@
...
+configuration file. The IP address is then associated with the converter.
+
+Docker
+``````
+The SCPI PSU Agent should be configured to run in a Docker container.
+An example docker-compose service configuration is shown here::
+
+  ocs-psuK:
+    image: simonsobs/ocs-scpi-psu-agent:latest
+    hostname: manny-docker
```
BrianJKoopman, line 51: Can we change the SAT1 specific `manny-docker` to `ocs-docker`? This is consistent with other examples.

JakeSpisak:
Addressed latest comments in the current commit, plus fixed some whitespace violations in the agent file.

PR #80, opened 2020-06-24
New CHWP encoder readout agent

harukinishino:
This is for a new agent for the CHWP encoder DAQ to receive and publish the data sent from Beaglebone Black reading the encoder and timing signals. This agent has three different data feeds for IRIG-B and the encoder counter and quadrature because the data rates are different.
Note that this is a bit newer than the one running currently with the real CHWP at UCSD, but this version of the agents is also tested with the other beaglebone test setup using a function generator.

### DIFF
--- a/agents/chwp/Dockerfile
+++ b/agents/chwp/Dockerfile
@@ -1,5 +1,4 @@
-# comment here...
-# 
+# CHWP Agent
 
 # Use socs base image
 FROM socs:latest
@@ -11,10 +10,10 @@ WORKDIR /app/socs/agents/chwp/
 COPY . .
 
 # Run registry on container startup
-ENTRYPOINT ["python3", "-u", "hwpbbb_agent.py"]
+ENTRYPOINT ["dumb-init", "python3", "-u", "hwpbbb_agent.py"]
 
 # Sensible default arguments
-CMD ["--site-hub=ws://sisock-crossbar:8001/ws", \
-     "--site-http=http://sisock-crossbar:8001/call"]
+CMD ["--site-hub=ws://crossbar:8001/ws", \
+     "--site-http=http://crossbar:8001/call"]
 
 EXPOSE 8080/udp
--- a/agents/chwp/Dockerfile
+++ b/agents/chwp/Dockerfile
@@ -1,0 +1,20 @@
+# comment here...
+# 
+
+# Use socs base image
+FROM socs:latest
+
+# Set the working directory to registry directory
+WORKDIR /app/socs/agents/chwp/
+
+# Copy this agent into the app/agents directory
+COPY . .
+
+# Run registry on container startup
+ENTRYPOINT ["python3", "-u", "hwpbbb_agent.py"]
+
+# Sensible default arguments
+CMD ["--site-hub=ws://sisock-crossbar:8001/ws", \
+     "--site-http=http://sisock-crossbar:8001/call"]
+
+EXPOSE 8080/udp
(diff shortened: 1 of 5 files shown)


BrianJKoopman:
Here is a summary of the action items I recorded this morning on our call.

* Use the IRIG information in the `start_acq` Process to make actual timestamps for the `data_counter` and `data_quad` data sets. Currently these are all given identical timestamps in one iteration of the loop. (This will only store the last data point in the live monitor database.)
* Combine separate feeds into single feed with different blocks. (The Lakeshore372 Agent was mentioned as one that publishes separate blocks, if only serially. Could be an example to look at.)
* Store the raw IRIG bit field.
* Add docstrings to supporting functions, particularly `grab_and_parse_data`. Information that would be valuable to incorporate includes how shared data structures like the queues are changing in each function.
* Currently, `flush_buffer` is called on every iteration of the acquisition loop. These flushes should be moved outside of the loop so that they are flushed when acquisition is stopped, but otherwise allow the feed to buffer the data.

I think I captured the major ones here, but certainly @jlashner and @mhasself should add any that I might have missed.

Since this agent, as is, has been in use for over a month I would like to get some recent version formally merged into the main branch, establishing a sort of version 0 for this agent. Perhaps the best path forward is I submit a review with some mundane things like requests for documentation and code style comments, and the above action items, which might take some more time to implement, are addressed in a later PR? Let me know if this doesn't make sense/you think it is unnecessary.

One other comment. In testing this with Peter during the DAQ breakout session of the FF2F meeting it seemed that the functionality here was somewhat dependent on the code uploaded to the beaglebone. Do we have some way of tracking which version of that code (is that code kept somewhere on github?) is the version we should be using with the latest version of this Agent?

harukinishino:
Sorry for taking too much time on this. I think we have addressed every item. We have used the latest version for a while in the setup at UTokyo without a major problem.

> Use the IRIG information in the start_acq Process to make actual timestamps for the data_counter and data_quad data sets. Currently these are all given identical timestamps in one iteration of the loop. (This will only store the last data point in the live monitor database.)

No more identical timestamps in a loop. The timestamps of the data feeds are still using computer system time because we want to look at data in grafana even when we have some issues in IRIG decoding or IRIG itself. We need some offline timing reconstruction anyway for precise analyses.

> Combine separate feeds into single feed with different blocks. 

Done. Instead, we separated them into two feeds, to avoid sending relatively high-rate raw encoder data to the influxdb publisher using the 'exclude_influx' flag.

> Store the raw IRIG bit field.

Now it is stored in irig_info in the HWPEncoder_irig_raw block.

> Add docstrings to supporting functions, particularly grab_and_parse_data. Information that would be valuable to incorporate includes how shared data structures like the queues are changing in each function.

I have added docstrings as much as possible.

> Currently, flush_buffer is called on every iteration of the acquisition loop. These flushes should be moved outside of the loop so that they are flushed when acquisition is stopped, but otherwise allow the feed to buffer the data.

Done.

> We should add a documentation page to the docs/agents/ directory for the CHWP agent. More details are always welcome, but the essential things to capture are example configuration blocks for the ocs-site-config file and the docker-compose file. See other agent documentation files for examples.

Added a minimum description about an example configuration.

> The Agent should be added to the socs/docker-compose.yml file. This will cause a Docker image to be automatically built and published to Docker Hub. Other agents are in that file and can be used as an example.

Added. I hope I did it correctly.

> There are some PEP08 whitespace violations, like mixed tab/spaces and trailing whitespace that should be addressed. I'd recommend running a linter like flake8 or pylint (tends to be more verbose) to catch these.

I have corrected those violations using pylint. Please let me know if you can still find any.

BrianJKoopman (commented):
This is looking great, thanks for all the changes Haruki. I took another quick look over the Agent and only had some very small suggestions. Since they were so minor, I made some updates while I was looking at the code, and PR'd them onto this branch in https://github.com/simonsobs/socs/pull/151. Details are over there, if they look good to you, we can merge onto chwp-agent, and then merge the result into develop.

BrianJKoopman (approved):

PR #82, opened 2020-07-01
Weather monitor agent

RemingtonGerras:
Weather Monitor Agent and its support code are written, as well as a Dockerfile. I have also added a weather monitor entry to the ocs_plugin_so.py file


agents/weather_agent/weather_monitor_agent.py
```diff
@@ -0,0 +1,189 @@
...
+def make_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+
+    pgroup = parser.add_argument_group('Agent Options')
+    pgroup.add_argument('--port', type=str,
+                        help="Path to USB node for the weather monitor")
+    pgroup.add_argument('--mode', type=str, choices=['idle', 'init', 'acq'],
+                        help="Starting action for the agent.")
+    pgroup.add_argument('--loops', type=int,
```
jlashner, line 159: Remove

agents/weather_agent/weather_monitor_agent.py
```diff
@@ -0,0 +1,189 @@
+
+import time
+import os
+import argparse
+import warnings
+import txaio
+
+from typing import Optional
+
+from weather_monitor import WeatherMonitor
```
jlashner, line 10: Move weather_monitor.py to `socs/socs/agents/weather_monitor/`, and use a package import instead of relative import. Something like: `from socs.agents.weather_monitor.weather_monitor import WeatherMonitor`

agents/weather_agent/weather_monitor_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    agent, runner = ocs_agent.init_site_agent(args)
+
+    path = '/dev'
+    for fname in os.listdir(path):
+        if fname[0:6] == 'ttyUSB':
+            usb = os.path.join(path, fname)
+            path += usb
+
+    print(path)
+
```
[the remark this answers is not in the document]
RemingtonGerras: how to define agent, runner if I don't initialize via ocs_agent.init_site_agent(args)? 

agent and runner are used later on, lines 183 & 184

agents/weather_agent/weather_monitor_agent.py
```diff
@@ -0,0 +1,189 @@
...
+    wMonitor = WeatherMonitorAgent(agent, args.port, args.loops)
+    agent.register_task('init_weathermonitor', wMonitor.init_weather_monitor_task,
```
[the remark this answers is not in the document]
RemingtonGerras: can delete line 184 no? Because the code block for init_weather_monitor_task is deleted as per 

> You can delete this part of the docstring. I realize it's in the LS240 agent but I probably should remove it there too.

agents/weather_agent/weather_monitor_agent.py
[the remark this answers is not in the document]
jlashner: I was just talking about removing the path determination stuff, since the `path` variable is not used. You should keep `ocs_agent.init_site_agent`

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,166 @@
...
+            sample_freq (int):
+                Frquency at which weather data is sampled.
+                Defaults to sample/2 seconds, the highest frequency possible.
+
+        """
+        time.sleep(2)
+        if params is None:
+            params = {}
+
+        sample_freq = params.get('freq')
```
jlashner, line 92: The key you get here needs to match the name of the arg in the docstring. I suggest changing to `params.get('sample_freq')`

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,166 @@
...
+            while self.take_data:
+                data = {
+                    'timestamp': time.time(),
+                    'block_name': 'weather',
+                    'data': {}
+                }
+
+                data['data'] = self.module.weather_daq()
+                self.agent.publish_to_feed('weather_data', data)
+                time.sleep(sample_freq)
```
jlashner, line 116: sleep time should be `1/sample_freq`.

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,166 @@
...
+            if not acquired:
+                self.log.warn("Could not start acq because {} is already running"
+                              .format(self.lock.job))
+                return False, "Could not acquire lock."
+
+            session.set_status('running')
+
+            self.take_data = True
+
+            while self.take_data:
```
jlashner, line 107: I suggest using the [Pacemaker](https://ocs.readthedocs.io/en/latest/api.html?highlight=Pacemaker#ocs.ocs_twisted.Pacemaker) class to have even quantized sampling.

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,166 @@
...
+    return parser
+
+
+if __name__ == "__main__":
+    parser = make_parser()
+    args = site_config.parse_args(
+        agent_class='VantagePro2Agent', parser=parser)
+    startup = False
+
+    if args.mode == 'acq':
```
jlashner, line 155: For an example of how to make the auto-acquire work, see the ls240 agent: https://github.com/simonsobs/socs/blob/015f25e9ad9b93b610971bfe93d6a1c4252df83e/agents/lakeshore240/LS240_agent.py#L322-L327 

If `startup` is a dict, it will be passed as `params` to the task on start.

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+    :prog: python3 vantage_pro2_agent.py
+
+Description
+-----------
+Out of the box, one just needs to connect the weather station to the 
+Vantage Pro2 monitor, and the monitor to the lab computer via usb. As long as
+the Vantage Pro2 monitor is on, the intitialize and data acquisition tasks 
+should work "out of the box", as it where.
+
+The Vantage Pro2 monitor and weather station records
```
jlashner, line 25: Seems like you were going to give a description of the data fields recorded by the weather station

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+    :filename: ../agents/vantagePro2_agent/vantage_pro2_agent.py
+    :func: make_parser
+    :prog: python3 vantage_pro2_agent.py
+
+Description
+-----------
+Out of the box, one just needs to connect the weather station to the 
+Vantage Pro2 monitor, and the monitor to the lab computer via usb. As long as
+the Vantage Pro2 monitor is on, the intitialize and data acquisition tasks 
+should work "out of the box", as it where.
```
jlashner, line 23: I think you can delete the sentence starting with "As long as"

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+``````````
+To configure the Vantage Pro2 Agent we need to add a VantagePro2Agent block to our ocs
+configuration file. Here is an example configuration block using all of the
+available arguments::
+
+     {'agent-class': 'VantagePro2Agent',
+       'instance-id': 'vantagepro2agent',
+       'arguments': [['--port', '/dev/ttyUSB0'],
+                     ['--mode', 'acq'],
+                     ['--freq', '2']]},
```
jlashner, line 42: change to `sample_freq`

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+
+ocs-config
+``````````
+To configure the Vantage Pro2 Agent we need to add a VantagePro2Agent block to our ocs
+configuration file. Here is an example configuration block using all of the
+available arguments::
+
+     {'agent-class': 'VantagePro2Agent',
+       'instance-id': 'vantagepro2agent',
+       'arguments': [['--port', '/dev/ttyUSB0'],
```
jlashner, line 40: This `/dev/ttyUSB0` port will change depending on how many other usb devices are plugged in. There may be udev rules that assigned a fixed device port based on the serial number. See the LS240 agent as an example: https://github.com/simonsobs/socs/blob/015f25e9ad9b93b610971bfe93d6a1c4252df83e/agents/lakeshore240/LS240_agent.py#L339-L345

I'd suggest looking on the system you're testing with and figure out a fixed way to determine the device port based on the hardware info.

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+                     ['--freq', '2']]},
+   
+You should know the port, or usb connection, that the Vantage Pro2 is connected to.
+Assign that location to the '--port' argument.
+
+Note, the '--freq' argument specifies the sample frequency that the Vantage Pro2 
+Monitor collects weather data. The Vantage Pro2 and weather station can sample weather
+data at a maximum sample frequency of 1/2 Hz. The user can define lower 
+sample frequencies if they so desire.  
+
```
jlashner, line 51: Seems like in you're code you are actually using the freq parameter to set the sampling period (1/freq) which is confusing. You should clarify the units here (probably should be hz), and if the user specifies 2 Hz you should sleep for 1/2 sec.

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+example docker-compose service configuration is shown here::
+
+  ocs-vantage-pro2:
+    build: /socs/agents/vantagePro2_agent
+
+    environment:
+      TARGET: VantagePro2Agent
+      NAME: 'Vantage Pro2'
+      DESCRIPTION: "Vantage Pro2 Weather Station Monitor"
+      FEED: "weather_data"
```
jlashner, line 64: Are these environment variables used anywhere in your code?

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+    volumes:
+      - ./:/config:ro
```
jlashner, line 66: Avoid using relative directory paths. You can use `${OCS_CONFIG_DIR}` which will pull the config path from she bash environment.

docs/agents/vantage_pro2.rst
```diff
@@ -0,0 +1,80 @@
...
+    command:
+      - "--instance-id=vantagepro2agent"
+      - "--site-hub=ws://sisock-crossbar:8001/ws"
+      - "--site-http=http://sisock-crossbar:8001/call"
+    devices:
+      - "/dev/ttyUSB0:/dev/ttyUSB0"
```
jlashner, line 72: Replace with fixed device path.

socs/agent/vantage_pro2/vantage_pro2.py
```diff
@@ -0,0 +1,293 @@
...
+        loop_data['soil_leaf_alarm1'] = byte_data[68]
+        loop_data['soil_leaf_alarm2'] = byte_data[69]
+        loop_data['soil_leaf_alarm3'] = byte_data[70]
+        loop_data['transmitter_battery_status'] = byte_data[71]
+        loop_data['console_battery_voltage'] = ((byte_data[72]*300)/512)/100.0
+        loop_data['forecast_icons'] = byte_data[73]
+        loop_data['forecast_rule_num'] = byte_data[74]
+        loop_data['time_sunrise'] = byte_data[75]
+        loop_data['time_sunset'] = byte_data[76]
+
```
jlashner, line 259: This is fine, but you should document what all of these values mean and the units of each. This should probably go in the docstring of the function.

socs/agent/vantage_pro2/vantage_pro2.py
```diff
@@ -0,0 +1,293 @@
...
+        # CRC check, data must be sent byte by byte
+        pure_data = struct.unpack('=99b', info)
+        self.crc_check(pure_data)
```
jlashner, line 262: Is there any reason you don't pass `info` here?

socs/agent/vantage_pro2/vantage_pro2.py
```diff
@@ -0,0 +1,293 @@
...
+        del loop_data['L']
+        del loop_data['O']
+        del loop_data['O1']
```
jlashner, line 265: I would probably just not set these fields in the first place

socs/agent/vantage_pro2/vantage_pro2.py
```diff
@@ -0,0 +1,293 @@
...
+        
+        return loop_data
+
+    def weather_daq(self):
+        """
+            Issues "LOOP <loops>" command to weather station, and unpacks
+            the data to daq{}.
```
jlashner, line 272: I think you can remove the words "to daq{}" because it isn't clear what you mean.
Also you should update this to be consistent with the fact that we are only sending "LOOP 1"

jlashner (changes requested):
Hey Remy, looks much better! I still have some suggestions inline. Let me know on slack if you want to work through them together.

socs/agent/vantage_pro2/vantage_pro2.py
RemingtonGerras: I loop through bytes in crc_check/calc_crc, info is a string

socs/agent/vantage_pro2/vantage_pro2.py
RemingtonGerras: need them for the crc check :/

RemingtonGerras: Ah I get what you meant here by the way. But! Addressed all the comments with my recent commit!

agents/vantagePro2_agent/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+WORKDIR /app/socs/agents/vantage_pro2/
+
+# Copy this agent into the app/agents directory
+COPY . .
+
+# Run registry on container startup
+ENTRYPOINT ["dumb-init", "python3", "-u", "vantage_pro2_agent.py"]
+
+CMD ["--site-hub=ws://sisock-crossbar:8001/ws", \
+     "--site-http=http://sisock-crossbar:8001/call"]
```
jlashner, line 17: Change "sisock-crossbar" to "crossbar" here

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        Method to start data acquisition process.
+
+        Args:
+            sample_freq (double):
+                Frequency at which weather data is sampled.
+                Defaults to 0.5 Hz.
+
+        """
+        time.sleep(2)
```
jlashner, line 99: Why do we need to sleep here? If a sleep is required after initialization this should go at the end of the init task.

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,213 @@
...
+                already running"""
+                              .format(self.lock.job))
+                return False, "Could not acquire lock."
+
+            session.set_status('running')
+            # use pacemaker class to take data at regular intervals
+            if sample_freq % 1 == 0:
+                pm = Pacemaker(sample_freq, True)
+            else:
+                pm = Pacemaker(sample_freq)
```
jlashner, line 121: Oops looking back now this is definitely a problem with the Pacemaker class and quantize should also work with sample frequencies that are 1/n Hz... my bad

agents/vantagePro2_agent/vantage_pro2_agent.py
```diff
@@ -0,0 +1,213 @@
...
+if not on_rtd:
+    from ocs import ocs_agent, site_config
+    from ocs.ocs_twisted import Pacemaker
+    from ocs.ocs_twisted import TimeoutLock
+
+
+class VantagePro2Agent:
+    """Agent to connect to single VantagePro2 Weather Monitor Device.
+
+    Args:
```
jlashner, line 20: add field for port arg

jlashner (changes requested):
Hey Remy, this is looking really good. There are a couple more small comments that I swear will be quick fixes and then I think we can merge!

jlashner:
Looks good! Lets get this thing merged

jlashner (approved):

PR #86, opened 2020-07-08
Fixing errors on the Pfeiffer agent

tanaybhandarkar:
The line querying gauge states was causing errors when trying to read channels that were off. I commented out that line so that the agent would stop crashing.


agents/pfeiffer_tpg366/pfeiffer_tpg366_agent.py
```diff
@@ -53,7 +53,6 @@ def read_pressure(self, ch_no):
         self.comm.send(ENQ.encode())
         read_str = self.comm.recv(BUFF_SIZE).decode()
         pressure_str = read_str.split(',')[-1].split('\r')[0]
-        pressure = float(pressure_str)
         return pressure
```
BrianJKoopman, line 56: `pressure` no longer exists for you to return.

tanaybhandarkar: Fixed, don't know how that deletion happened.

PR #74, opened 2020-04-25
Cryomech CPA Agent Cleanup

gdevenyi:
Implement a number of changes to Cryomech agent
- uniformize conversion using struct.unpack for big-endian data
- add additional outputs
- add lookup table for status codes -- note, "-2147483648" is returned from my device under no warning/error condition, rather than 0
- renamed fields according to the cryomech manual

I took cryomech_cpa_agent.py and wrote a standalone version for my monitoring systems. This is a backport of the changes I made during my quest to understand the data returned from the device. These changes have not been tested in the context of "socs"

PR #87, opened 2020-07-09
Rename invalid field names

BrianJKoopman:
This PR adds the `rename_fields` script, which attempts to correct all invalid field names that might have been previously recorded in .g3 files across all OCS/SOCS Agents. Since we want to eliminate any invalid field names in current Agents, we also simultaneously make updates to all SOCS Agents with invalid field names.

I've tested this on single days worth of data at Yale, and on single files from UCSD's system. It seems to work, but a review of the functionality of the renamer and additional verification from someone else would be helpful. To help with testing I've added some functionality to the `checkdata` script in ocs (see https://github.com/simonsobs/ocs/pull/147) to identify invalid field names. The sort of ultimate test will be in converting the Yale data set and re-uploading to InfluxDB. This will occur before we do so at any other institution.

@JakeSpisak - I attempted to cover some of the changes mentioned recently in #74 and #88. Can you take a look at the changes to that Agent that I've made and comment, particularly at the questions raised in 42a27cf?

### DIFF
--- a/agents/cryomech_cpa/cryomech_cpa_agent.py
+++ b/agents/cryomech_cpa/cryomech_cpa_agent.py
@@ -50,6 +50,7 @@ class PTC:
 
         self.model = None
         self.serial = None
+        self.software_revision = None
 
         self.comm = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.comm.connect((self.ip_address, self.port))  # connects to the PTC
@@ -95,16 +96,16 @@ class PTC:
                   "Compressor_State": [11, 12],
                   "Warning_State": [15, 16, 13, 14],
                   "Alarm_State": [19, 20, 17, 18],
-                  "Coolant_In_Temp": [22, 21, 24, 23],
-                  "Coolant_Out_Temp": [26, 25, 28, 27],
-                  "Oil_Temp": [30, 29, 32, 31],
-                  "Helium_Temp": [34, 33, 36, 35],
-                  "Low_Pressure": [38, 37, 40, 39],
-                  "Low_Pressure_Average": [42, 41, 44, 43],
-                  "High_Pressure": [46, 45, 48, 47],
-                  "High_Pressure_Average": [50, 49, 52, 51],
-                  "Delta_Pressure_Average": [54, 53, 56, 55],
-                  "Motor_Current": [58, 57, 60, 59],
+                  "Coolant_In_Temp": [23, 24, 21, 22],
+                  "Coolant_Out_Temp": [27, 28, 25, 26],
+                  "Oil_Temp": [31, 32, 29, 30],
+                  "Helium_Temp": [35, 36, 33, 34],
+                  "Low_Pressure": [39, 40, 37, 38],
+                  "Low_Pressure_Average": [43, 44, 41, 42],
+                  "High_Pressure": [47, 48, 45, 46],
+                  "High_Pressure_Average": [51, 52, 49, 50],
+                  "Delta_Pressure_Average": [55, 56, 53, 54],
+                  "Motor_Current": [59, 60, 57, 58],
                   "Hours_of_Operation": [63, 64, 61, 62],
                   "Pressure_Unit": [65, 66],
                   "Temperature_Unit": [67, 68],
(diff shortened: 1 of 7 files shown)


BrianJKoopman:
This is ready for review now. I've done some testing, but review of the HKRenamer and of changes to each Agent would be helpful. I made all changes to field names just reading through Agents, I haven't directly run any Agents on a system which would reject invalid field names yet, but intend to do so before publishing new Agents for users to deploy.

agents/cryomech_cpa/cryomech_cpa_agent.py
```diff
@@ -80,20 +80,27 @@ def breakdownReplyData(self, rawdata):
...
+                  "Coolant_Out_Temp": [26, 25, 28, 27],
+                  "Oil_Temp": [30, 29, 32, 31],
+                  "Helium_Temp": [34, 33, 36, 35],
+                  "Low_Pressure": [38, 37, 40, 39],
+                  "Low_Pressure_Average": [42, 41, 44, 43],
+                  "High_Pressure": [46, 45, 48, 47],
+                  "High_Pressure_Average": [50, 49, 52, 51],
+                  "Delta_Pressure_Average": [54, 53, 56, 55],
+                  "Motor_Current": [58, 57, 60, 59],
+                  "Hours_of_Opperation": [63, 64, 61, 62],
```
JakeSpisak, line 98: small typo: two p's in "Opperation"

JakeSpisak:
All my comments (aside a small typo) are in the answer to the commit 42a27cf. Once we agree there, I'm happy to push working compressor code with the necessary edits to this branch.

agents/cryomech_cpa/cryomech_cpa_agent.py
BrianJKoopman: Good catch, fixed, thanks!

BrianJKoopman:
> 1. The short answer is I'd recommend keeping "Hours of Operation", "Pressure Unit", "Temperature Unit", and "Software Revision" in the database, and move "Serial Number" and "Model" to a separate function. The code as written doesn't work, it needs more of the updates from #74. I can push validated code if you want.
>
> Since 'Serial Number' and 'Model' are more compressor properties than actual data, I'd recommend not storing them in the database and instead adding them in as another class method, something like "get_compressor_properties". I can add this in as part of my general compressor update, #88. 'Software Revision', 'Pressure Unit', and 'Temperature Unit' aren't really data so much as the current system state (the latter two are physical units, e.g. zeros mean Kelvin and Psi). But they might be good to store so that users can go back and see what their values were at a particular time.

Alright, thanks for the quick testing and info! I've made a few more small changes in 67f3487. My intention is to still decode the Model and Serial Numbers from the raw data, but just store them in the `self.ptc` object attributes. Upon initialization we'll issue a `get_data` call and display the model and serial numbers in the Agent logs.

Can you make any needed changes here to `breakdownReplyData` to get this working? I think the proper workflow would be to branch off of this field-rename branch and PR onto it. Let me know if you have questions. I'm going to continue to work on this branch to address comments from Matthew.

>     2. No, the order doesn't matter. I reversed the order and it worked fine.

Good to know, thanks!

BrianJKoopman:
> The claims that it changes invalid names to be valid seem a little overstated. It does so for several cases that we know of. But other stuff (e.g. "000a") would sneak through unaltered. This is an opportunity to check for invalid fields that we didn't know about in advance. Perhaps warn/fail on that?

Good point! Added a check via `Feed.verify_data_field_string` that will raise an exception when it encounters an invalid field name after our attempted renaming. Warning might have been good, but I don't want the warning to pass by ignored, so we'll fail instead. Urged users to open an issue with the invalid name encountered. (Even though probably this'll just be run by me on each system...)

> This rename_fields script is going to be useful for a couple of months, then not. Will we plan to drop it from the repo when v0 has been eradicated (along with Translator)? If not then the next paragraph is more important.

I hadn't really thought ahead to whether we'd get rid of this eventually. If other format changes come up later it could be adapted. It doesn't quite work that v0 is the only HK version that could have gotten some invalid field names, else I'd say we should just have a check that this runs on v0, then it would only be useful until v0 is gone. It's highly unlikely any invalid field names got written to a v1 format though.

Anyway, we should revisit this in a few months.

BrianJKoopman:
I'd like to sneak one more thing into this PR, and that is to bump the ocs version we're basing the socs images off of to the latest (`v0.6.0-67-g5993b6a-dev`). This'll ensure any invalid field names that we didn't catch here don't actually get published to the Feeds, as the latest ocs version will catch them at that level.

If nothing else comes up, we could soon make a release on ocs, and update this version here accordingly (following that up with official release of this repo, likely after adopting the same release model recently implemented in ocs). This dev version will be useful as we begin testing the images at test institutions and applying the new `rename_fields` script.

PR #91, opened 2020-07-23
Track processed files and their output status

BrianJKoopman:
This adds some tracking functionality to the field renaming script so we don't process all files every time we run the script. It's similar to the tracking in ocs' g32influx script, using sqlite to record the state of processed files. We also add logging, sensible renaming for the M1000 Agent, and progress bars.

### DIFF
--- a/bin/rename_fields
+++ b/bin/rename_fields
@@ -376,4 +376,12 @@ if __name__ == '__main__':
 
             c.close()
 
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
Self-merging. These changes were used in the HK v1 translation at Yale, and are needed as we prep to update to v2 at other institutions.

PR #99, opened 2020-08-04
New Tektronix3021C AWG agent.

mjrand:
Created an agent for the Tektronix3021c AWG to be used for vibration testing.

## Description
New agent for highbay AWG.

## Motivation and Context
Agentify vibration Tests.

## How Has This Been Tested?
I have used this agent to run vibration tests on SAT1.

## Types of changes

## Checklist:

### DIFF
--- a/agents/tektronix3021c/tektronix_agent.py
+++ b/agents/tektronix3021c/tektronix_agent.py
@@ -1,0 +1,212 @@
+"""Michael Randall
+mrandall@ucsd.edu"""
+
+import time
+import os
+import socket
+import argparse
+
+from socs.agent.tektronix3021c_driver import tektronixInterface
+
+on_rtd = os.environ.get('READTHEDOCS') == 'True'
+if not on_rtd:
+    from ocs import ocs_agent, site_config
+    from ocs.ocs_twisted import TimeoutLock
+
+
+class TektronixAWGAgent:
+    def __init__(self, agent, ip_address, gpib_slot):
+        """
+        Tektronix3021c Agent Initializer
+
+        params:
+            ip_address (string): the IP address of the gpib to ethernet
+            controller connected to the function generator.
+
+            gpib_slot (int): the gpib address currently set
+            on the function generator.
+        """
+
+        self.agent = agent
+        self.log = agent.log
+        self.lock = TimeoutLock()
+
+        self.job = None
+
+        self.ip_address = ip_address
+        self.gpib_slot = gpib_slot
+        self.monitor = False
+
+        self.awg = None
+        # Registers data feeds
+        agg_params = {
+            'frame_length': 60,
+        }
+        self.agent.register_feed('awg',
+                                 record=True,
+                                 agg_params=agg_params)
+
+    def init_awg(self, session, params=None):
+        """ Task to connect to Tektronix AWG """
+
+        with self.lock.acquire_timeout(0) as acquired:
+            if not acquired:
+                return False, "Could not acquire lock"
+
+            try:
+                self.awg = tektronixInterface(self.ip_address, self.gpib_slot)
+                self.idn = self.awg.identify()
+
+            except socket.timeout as e:
+                self.log.error("""Tektronix AWG
+                               timed out during connect -> {}""".format(e))
+                return False, "Timeout"
+
+            self.log.info("Connected to AWG: {}".format(self.idn))
+
+        return True, 'Initialized AWG.'
+
+    def set_frequency(self, session, params=None):
+        """
+        Sets frequency of function generator:
+
+        Args:
+            frequency (float): Frequency to set in Hz.
+            Must be between 0 and 25,000,000.
+        """
+
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                freq = params.get("frequency")
+
+                try:
+                    float(freq)
+
+                except ValueError as e:
+                    return False, """Frequency must
+                                    be a float or int -> {}""".format(e)
+
+                except TypeError as e:
+                    return False, """Frequency must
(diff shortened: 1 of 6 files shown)


agents/tektronix3021c/tektronix_agent.py
```diff
@@ -0,0 +1,146 @@
...
+    # Parse comand line.
+    args = parser.parse_args()
+    # Interpret options in the context of site_config.
+    site_config.reparse_args(args, 'Tektronix AWG')
+
+    agent, runner = ocs_agent.init_site_agent(args)
+
+    p = TektronixAWGAgent(agent, args.ip_address, int(args.gpib_slot))
+
+    agent.register_task('init', p.init_awg)
```
BrianJKoopman, line 141: This init task could be run at startup if it's convenient with the `startup=True` argument. See the ocs docs for more details: https://ocs.readthedocs.io/en/latest/developer/agents.html#example-agent

agents/tektronix3021c/tektronix_agent.py
```diff
@@ -0,0 +1,146 @@
...
+    def set_output(self, session, params=None):
+        """
+        Task to turn channel on or off.
+
+        Args:
+            state (bool): True for on, False for off.
+        """
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                state = params['state']
```
BrianJKoopman, line 110: It would be safer to use a `params.get('state')` with a default value or perhaps `None` and some error handling for if a user doesn't pass in a valid value (or no value at all) to avoid a `KeyError` or the like.

agents/tektronix3021c/tektronix_agent.py
```diff
@@ -0,0 +1,146 @@
...
+    def set_amplitude(self, session, params=None):
+        """
+        Sets current of power supply:
+
+        Args:
+            Amplitude (float): Peak to Peak voltage to set. Must be between 0 and 10.
+        """
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                amp = params['amplitude']
```
BrianJKoopman, line 87: See the below `params.get()` suggestion.

agents/tektronix3021c/tektronix_agent.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        Sets frequency of function generator:
+
+        Args:
+            frequency (float): Frequency to set. Must be between 0 and 25,000,000
+        """
+
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                freq = params['frequency']
```
BrianJKoopman, line 64: Again, `params.get()`.

BrianJKoopman (changes requested):
This now includes a bunch of changes that have occurred since the last time you committed. This branch should be rebased on the `develop` branch before proceeding, that'll ensure that the PR only contains the changes relevant to the new Agent.

Happy to help with those steps if you aren't familiar with rebasing, feel free to reach out on Slack.

jlashner:
Hey Brian, looks like Michael addressed most of your comments. I did a few things to clean up the git history and make sure docs build, but didn't really change any code.

I think the one thing that looks like it's not addressed is the following comment.
> In the design of the agent you publish the values you set on the device to the OCS feed. This assumes the setting takes to the device, i.e. there isn't some error or restriction on the device that causes the setting to not work. It also would not catch if a user manually changes the device, nor would it capture the value setting if the Agent is restarted for some reason during a run (without resetting the value as well anyway.) In many other Agents there is a process that will query relevant values and publish those on the ocs Feed. I'm not saying you must do this, just trying to point out some edge cases here. You know the hardware device and its use cases better than I do. This could always be added later too, certainly the Agent as is has been working for you.

It looks like this'll require additional functions which @mjrand  will need to add to the driver to read back the amplitude and frequency, but I'm tempted to just merge. What do you think?

mhasself:
> > In the design of the agent you publish the values you set on the device to the OCS feed. This assumes the setting takes to the device, i.e. there isn't some error or restriction on the device that causes the setting to not work. It also would not catch if a user manually changes the device, nor would it capture the value setting if the Agent is restarted for some reason during a run (without resetting the value as well anyway.) In many other Agents there is a process that will query relevant values and publish those on the ocs Feed. I'm not saying you must do this, just trying to point out some edge cases here. You know the hardware device and its use cases better than I do. This could always be added later too, certainly the Agent as is has been working for you.
> 
> It looks like this'll require additional functions which @mjrand will need to add to the driver to read back the amplitude and frequency, but I'm tempted to just merge. What do you think?

Perhaps a quick interim fix would be to rename the fields to, e.g. "awg_amp_cmd", which (a) clarifies they are commanded values, not measured / readback and (b) leaves namespace to later record the actual "awg_amp" that is read back.

docs/agents/tektronix3021c.rst
```diff
@@ -0,0 +1,85 @@
+.. highlight:: rst
+
+.. _tektronix3021c:
+
+===================
+Tektronix AWG Agent
+===================
+
+This agent uses Standard Commands for Programmable Instruments (SCPI)
+It works for many function generators, including the Tektronix3021c.
```
BrianJKoopman, line 10: If this works for many function generators, not just the Tektronix3021c, should this Agent be more generally named? Like we did with the SCPI PSU Agent? (We can perhaps do this renaming later if it becomes clear we're using more than just the Tektronix device, I know this is time sensitive.)

BrianJKoopman:
This is looking good. I do like Matthew's suggestion, you should considering doing that before merging. I also fixed a small docstring thing in #143. (I know you're in a hurry, so I just made the change instead of telling you about it.)

BrianJKoopman (approved):

Issue #94, opened 2020-07-29
SNMP Trap Agent


BrianJKoopman:
After working with SNMP on the M1000 Agent, it has become evident that it might make sense to have a single SNMP Trap agent, meant to catch all SNMP traps, rather than introduce trap functionality to each Agent that uses SNMP to interface with a piece of hardware. This might eventually make sense to move upstream to the ocs repo as a core OCS Agent, but let's start here.

Issue #30, opened 2019-11-06
Experiment with SNMP


mhasself:
At some point we will want Agents that speak SNMP to some hardware devices.  It would be useful to take a few steps in that direction, perhaps with a simple ibootbar control agent.

SNMP is the best way to communicate with:
- remote power switches such as ibootbars
- UPSes such as Eaton UPSes

UPS and ibootbar communication over SNMP has been implemented for ACT; that code can be extracted and shared.

A cursory reading of http://snmplabs.com/pysnmp/examples/v3arch/twisted/contents.html suggests that twisted + SNMP is going to work great.

BrianJKoopman:
SNMP support added by #101, closing.

PR #101, opened 2020-08-07
Create M1000 agent and add SNMP support to socs library

BrianJKoopman:
## Description

This PR introduces SNMP support to socs and adds an Agent for the M1000, which uses SNMP.

This adds the Meinberg M1000 Agent, for monitoring parts of the M1000 via SNMP. To do so we also introduce SNMP support to socs. This uses pysnmp, which supports twisted.

## Motivation and Context

We need an agent for the M1000, and SNMP is a great interface for monitoring the state of the device. #30 motivated wide support for SNMP, which we attempt to provide here with the snmp module in socs. It probably makes sense to eventually move this upstream to ocs.

The one thing this doesn't solve yet is SNMP Trap support, which might be useful as noted in #94.

## How Has This Been Tested?

This has been run with the M1000 in an isolated OCS setup (separate from the main Yale setup.) Typically this has only been for short durations of time, but things are working so far.

## Types of changes

## Checklist:


BrianJKoopman:
This should be ready for review now.

docs/developer/snmp.rst
```diff
@@ -0,0 +1,88 @@
...
+    from socs.snmp import SNMPTwister
+
+    snmp = SNMPTwister(address, port)
+
+    get_list = [ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgRefclockState', 1)),
+                ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgSysPsStatus', 1)),
+                ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgSysPsStatus', 2))]
+
+    result = yield snmp.get(get_list)
+
```
mhasself, line 78: I think this example is a little too sketchy.  Consider fleshing out to a full working script -- would only ~double the number of lines.  Need some imports (where is ObjectType from?) and a ```@inlineCallbacks``` function and a reactor run.  Would still be helpful (the pysnmp examples are pretty intimidating).  Is it possible to run without an MIB?  Such a bare bones +twisted starting point could be really useful for people working on new Agents.

socs/snmp.py
```diff
@@ -0,0 +1,113 @@
...
+                      ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB',
+                                                'mbgLtNgRefclockLeapSecondDate',
+                                                1))])
+
+        Parameters
+        ----------
+        oid_list : list
+            List of high-level MIB Object OIDs. See `Specifying MIB Objects`_ for
+            more info
+
```
mhasself, line 100: How would you feel about accepting either
```
ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgRefclockState', 1))
```
or simply
```
('MBG-SNMP-LTNG-MIB', 'mbgLtNgRefclockState', 1)
```
for each element of the list?  I think that would save a lot of needless imports and typing in Agent code.  You could just promote each tuple to a ObjectType if it's not already an Object Type.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    if parser is None:
+        parser = argparse.ArgumentParser()
+
+    # Add options specific to this agent.
+    pgroup = parser.add_argument_group("Agent Options")
+    pgroup.add_argument("--auto-start", default=True, type=bool,
+                        help="Automatically start listening for data at " +
+                        "Agent startup.")
+    pgroup.add_argument("--address", help="Address to listen to.")
```
mhasself, line 171: "Listening" is left over from the template agent I think... we are polling, not listening.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+
+                    # Update lastGet time
+                    for mib in self.mib_timings:
+                        if mib['oid'] in get_list:
+                            mib['lastGet'] = read_time
+
+                    self.log.debug("{msg}", msg=message)
+                    session.app.publish_to_feed('m1000', message)
+
+                result = []
```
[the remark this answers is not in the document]
BrianJKoopman: Yeah, good point, moved to top of the loop.

socs/snmp.py
BrianJKoopman: I like the suggestion of accepting the tuple. I probably wrote it this way originally just sticking to the structure that was in the pysnmp doc examples. I would sort of lean towards just accepting the tuples, as it likely would make the documentation more clear. However, I've written it so that it currently accepts both, promoting the tuples to ObjectType if they aren't already like you suggested.

agents/meinberg_m1000/meinberg_m1000_agent.py
BrianJKoopman: Yup, good catch, fixed.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                                 agg_params=agg_params,
+                                 buffer_time=1)
+
+        self.mib_timings = [{"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgRefclockState', 1)), "interval": 60, "lastGet": None},
+                            {"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgSysPsStatus', 1)), "interval": 60, "lastGet": None},
+                            {"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgSysPsStatus', 2)), "interval": 60, "lastGet": None},
+                            {"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgEthPortLinkState', 1)), "interval": 60, "lastGet": None},
+                            {"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgRefclockLeapSecondDate', 1)), "interval": 60*60, "lastGet": None},
+                            {"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgNtpCurrentState', 0)), "interval": 64, "lastGet": None},
+                            {"oid": ObjectType(ObjectIdentity('MBG-SNMP-LTNG-MIB', 'mbgLtNgPtpPortState', 1)), "interval": 3, "lastGet": None}]
```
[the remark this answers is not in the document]
BrianJKoopman: You mean the intervals here?

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                            # look like 'mbgLtNgRefclockState_1'
+                            valid_field = oid.split("::")[1].replace('.', '_')
+
+                            self.log.debug("{o} {value}",
+                                           o=oid,
+                                           value=int(item[1]))
+                            message['data'][valid_field] = int(item[1])
+                        except ValueError:
+                            self.log.warn('{oid} is of type {_type}, not int',
+                                          oid=item[1], _type=type(item[1]))
```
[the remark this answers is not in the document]
BrianJKoopman: The violation you have in mind is that the number of fields returned would change? OID types are fixed, so I think if you come across an unsupported type it'll always be that type and always raise this exception, so I don't think there's risk of feed rule violation.

That said, this was written before we supported strings, and that's the instance this catches in this specific example. I'll update to allow those.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+
+                    # Update lastGet time
+                    for mib in self.mib_timings:
+                        if mib['oid'] in get_list:
+                            mib['lastGet'] = read_time
+
+                    self.log.debug("{msg}", msg=message)
+                    session.app.publish_to_feed('m1000', message)
+
```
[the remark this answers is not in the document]
BrianJKoopman: Alright, this is added now. We store the OIDs, their latest value, the corresponding description string, and their last GET time in session.data.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                            self.log.warn('{oid} is of type {_type}, not int',
+                                          oid=item[1], _type=type(item[1]))
+
+                    # Update lastGet time
+                    for mib in self.mib_timings:
+                        if mib['oid'] in get_list:
+                            mib['lastGet'] = read_time
+
+                    self.log.debug("{msg}", msg=message)
+                    session.app.publish_to_feed('m1000', message)
```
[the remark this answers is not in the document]
BrianJKoopman: Is having the string saved needed? Now with it in session.data the latest string description is visible to clients. Grafana can be configured to display the corresponding string, but through a value mapping setup in a dashboard, rather than storing the strings themselves in InfluxDB. What other applications did you have in mind?

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,201 @@
...
+                        get_list.append(mib["oid"])
+                    elif time.time() - mib["lastGet"] > mib["interval"]:
+                        get_list.append(mib["oid"])
+
+                if get_list:
+                    # Issue SNMP GET command
+                    result = yield self.snmp.get(get_list)
+                    read_time = time.time()
+
+                if result:
```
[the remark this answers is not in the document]
BrianJKoopman: This has been significantly restructured. This loop moved to `MeinbergSNMP.run_snmp_get()` and no longer has many of those if checks. I did use your continue suggestion for when we don't have any OIDs to GET yet (i.e when get_list is empty).

docs/developer/snmp.rst
BrianJKoopman: I would love to make this a standalone script. I do not know how to get this to work though. Would be happy to discuss it on Slack sometime.

ObjectType is no longer needed (it's from pysnmp), since SNMPTwister now accepts tuples per your suggestion above. This example is extremely specific to the M1000, so no, it's not possible to run this without the MIB. Some demo SNMP target could probably be made though...

BrianJKoopman: Alright, in typical fashion I have figured it out immediately after saying otherwise. Will update in a moment...

BrianJKoopman: So the example I've come up with that works looks like this:
```
    from twisted.internet import reactor
    from socs.snmp import SNMPTwister
    
    # Setup communication with M1000
    snmp = SNMPTwister('10.10.10.186', 161)
    
    # Define OIDs to query
    get_list = [('MBG-SNMP-LTNG-MIB', 'mbgLtNgRefclockState', 1),
                ('MBG-SNMP-LTNG-MIB', 'mbgLtNgSysPsStatus', 1),
                ('MBG-SNMP-LTNG-MIB', 'mbgLtNgSysPsStatus', 2)]
    
    def printResult(x):
        print(x)
    
    # Run the twisted SNMP GET and print the results
    d = snmp.get(get_list)
    d.addCallback(printResult)
    
    # Shut things down by asking the reactor to stop after 3 seconds
    reactor.callLater(3, reactor.stop)
    reactor.run()
```

My problem with this is that it doesn't look like what you'll end up writing from within an agent, which is what I was trying to go for with the other example, as I don't expect people to be using SNMPTwister much outside of Agents, at least, that's outside the scope of this guide.

BrianJKoopman:
I think this is ready for re-review. To describe the changes a bit in words, since the diff is going to be large -- I've created a helper class much like other Agents have that handles all of what was previously going on in the record process (now called "acq" like many other Agents, "record" was carried over from the smurf recorder agent that I started from when making this Agent.) This helper Agent has a method called `run_snmp_get` that issues the GET commands for each interval, the main inner loop previously within the "while is_streaming" loop of the Agent.

This `run_snmp_get` method is pretty readable, each main step is broken out into a method within this helper class -- building the "get_list", updating the cache (which is passed to session.data), and building the OCS Feed message.

docs/developer/snmp.rst
mhasself: How about:
```
@inlineCallbacks
def query_snmp():
    x = yield snmp.get(get_list)
    print(x)
    reactor.stop()
```

socs/snmp.py
```diff
@@ -0,0 +1,120 @@
...
+        """Failure Errback for getCmd.
+
+        Taken from Twisted example for SNMPv1 from pySNMP documentation:
+        http://snmplabs.com/pysnmp/examples/hlapi/twisted/contents.html
+
+        """
+        self.log.error('%s failure: %s' % (self.address, error_indication))
+
+    def get(self, oid_list):
+        """Issue a getCmd to get SNMP OID states.
```
mhasself, line 83: This docstring could be a touch more helpful... although it gives examples of calling get(), it does not tell me what get() returns, nor more importantly how I can extract the gotten values.

Perhaps be a little more pointed on this docstring (what, exactly, does get() return?  A Deferred, which will callback with the var_binds list from self._success...) and then show a more complete usage example in the class docstring.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,482 @@
...
+        self.address = address
+        self.port = port
+        self.snmp = SNMPTwister(address, port)
+
+        # Decode OID States
+        self.decoder = {'mbgLtNgRefclockState': REFCLOCKSTATE,
+                        'mbgLtNgNtpCurrentState': NTPCURRENTSTATE,
+                        'mbgLtNgSysPsStatus': SYSPSSTATUS,
+                        'mbgLtNgEthPortLinkState': ETHPORTLINKSTATE,
+                        'mbgLtNgPtpPortState': PTPPORTSTATE}
```
mhasself, line 92: This map from OID to enumerator could be at global (module) scope.  That would be more useful than only having REFCLOCKSTATE and friends at global scope.  In fact you wouldn't need REFCLOCKSTATE and friends at all.

This is a useful data structure (for interpreting / displaying data downstream, perhaps far in the future), and it doesn't need to be hidden in an instance.

docs/developer/snmp.rst
BrianJKoopman: Mhm, I had tried something like this, it hangs and doesn't print anything.

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,482 @@
...
+
+# Mapping of integer OID values to meaningful strings from M1000 manual
+REFCLOCKSTATE = {0: "refclock is not available",
+                 1: "synchronized",
+                 2: "not synchronized"}
+
+NTPCURRENTSTATE = {0: "not available",
+                   1: "not synchronized",
+                   2: "synchronized"}
+
```
mhasself, line 28: The meaningful strings above this point and below this point are somewhat differently constructed... I know they all come from the manual but I think it will look better if we adopt the approach of the strings below for the strings above... i.e. make each string a single word ("notAvailable", "notSynchronized"), and drop "refclock is ".

agents/meinberg_m1000/meinberg_m1000_agent.py
```diff
@@ -0,0 +1,482 @@
...
+                 "description": "up",
+                 "lastGet": 1598543359.6326838},
+            "mbgLtNgSysPsStatus_2":
+                {"status": 2,
+                 "description": "up",
+                 "lastGet": 1598543359.6326838},
+            "mbgLtNgEthPortLinkState_1":
+                {"status": 1,
+                 "description": "up",
+                 "lastGet": 1598543359.6326838}}
```
mhasself, line 412: What's the format of LeapSecondDate?  Since you tell me all the other ones...

docs/developer/snmp.rst
BrianJKoopman: Debugged offline, example updated.

agents/meinberg_m1000/meinberg_m1000_agent.py
BrianJKoopman: Yeah, I just copied each description verbatim from the manual. Updated to using all single word strings.

agents/meinberg_m1000/meinberg_m1000_agent.py
BrianJKoopman: Right, that's actually a str, which differed from other "strings" which were stored as bytes within the returned ObjectType. The types within ObjectType were pretty awkward to handle and I'm mostly empirically figuring out what works. I was initially assuming all strings were initially `bytes`, but now am letting `str` through uncaught as well. I added LeapSecondDate to the docstring. Good catch.

agents/meinberg_m1000/meinberg_m1000_agent.py
BrianJKoopman: Moved the decoder to global scope and got rid of REFCLOCKSTATE and friends, consolidating to one dict.

socs/snmp.py
BrianJKoopman: Attempted to make more clear in 0be0a90. The returned objects from getCmd I find awkward to work with. I'm reaching into a hidden value `._value` to get the integer the corresponds to the description string in the Enum, as there doesn't seem to be a good way to grab it otherwise, just the actual description string (which I did start saving to the .g3 files as well in b8676d5.) They seemed well designed to print, as seen in the example.

BrianJKoopman:
Just wanted to note that the patch in https://github.com/simonsobs/ocs/pull/164 needs to happen along with an ocs version bump here in order for the data published by this Agent to work properly with the OCS Feeds.

PR #103, opened 2020-08-25
Create synacc agent

JakeSpisak:
Create synacc agent

## Description
Added the synaccess power strip agent, with on/off and status query capabilities.

## How Has This Been Tested?
All agent functions were tested on a power strip at UCSD.

## Types of changes

## Checklist:


agents/synacc/synacc.py
```diff
@@ -0,0 +1,115 @@
...
+        self.ip_address = ip_address
+        self.user = username
+        self.passw = password
+
+    def get_status(self, session, params=None):
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                req = "http://" + self.user + ":" + self.passw + "@" +\
+                    self.ip_address+"/cmd.cgi?$A5"
+                r = requests.get(req)
```
BrianJKoopman, line 25: I don't think it should hold up this PR, but this could be made to support twisted by using [twisted's HTTP client](https://twistedmatrix.com/documents/current/web/howto/client.html) instead of `requests`.

agents/synacc/synacc.py
```diff
@@ -0,0 +1,115 @@
...
+
+class SynAccAgent:
+    def __init__(self, agent, ip_address, username, password):
+        self.agent = agent
+        self.lock = TimeoutLock()
+        self.ip_address = ip_address
+        self.user = username
+        self.passw = password
+
+    def get_status(self, session, params=None):
```
BrianJKoopman, line 20: It would be helpful on the Client end to have the outlet status saved in `session.data`. This now has documentation here: https://ocs.readthedocs.io/en/develop/developer/session-data.html

agents/synacc/synacc.py
```diff
@@ -0,0 +1,115 @@
...
+            if acquired:
+                req = "http://"+self.user + ":" + \
+                    self.passw + "@" + self.ip_address + \
+                    "/cmd.cgi?$A4" + " " + str(params['outlet'])
+                requests.get(req)
+                return True, 'Rebooted outlet {}'.format(params['outlet'])
+            else:
+                return False, "Could not acquire lock"
+
+    def set_outlet(self, session, params=None):
```
BrianJKoopman, line 48: This could use a docstring describing the required params.

agents/synacc/synacc.py
```diff
@@ -0,0 +1,115 @@
...
+                req = "http://" + self.user + ":" + self.passw + "@" + \
+                    self.ip_address+"/cmd.cgi?$A3" + " " + \
+                    str(params['outlet']) + " " + on
+                requests.get(req)
+                return True, 'Set outlet {} to {}'.\
+                    format(params['outlet'], params['on'])
+            else:
+                return False, "Could not acquire lock"
+
+    def set_all(self, session, params=None):
```
BrianJKoopman, line 64: Similarly this could use a docstring describing the required params.

agents/synacc/synacc.py
```diff
@@ -0,0 +1,115 @@
...
+import argparse
+
+on_rtd = os.environ.get('READTHEDOCS') == 'True'
+if not on_rtd:
+    from ocs import ocs_agent, site_config
+    from ocs.ocs_twisted import TimeoutLock
+
+
+class SynAccAgent:
+    def __init__(self, agent, ip_address, username, password):
```
BrianJKoopman, line 13: The class could use a docstring as well, since it is autoclass'd in the docs.

docs/agents/synacc.rst
```diff
@@ -0,0 +1,82 @@
...
+To configure the SynAcc Agent we need to add a SynAcc Agent block to our ocs 
+configuration file. Here is an example configuration block using all of the 
+available arguments::
+
+       {'agent-class': 'SynAccAgent',
+        'instance-id': 'synacc',
+        'arguments':[
+          ['--ip-address', '10.10.10.8'],
+          ['--username', 'admin'],
+          ['--password', 'admin'],
```
BrianJKoopman, line 33: How do these username/password get setup initially? I take it admin/admin is default?

BrianJKoopman (changes requested):
This is looking good, couple of comments inline, mostly documentation requests.

I might also generally suggest we publish the power states to an OCS feed. It doesn't have to be regularly like acq processes in other Agents, but it would give us record of when power was cycled, etc.

You will also need to add a block to the main [`docker-compose.yml` file ](https://github.com/simonsobs/socs/blob/create_synacc_agent/docker-compose.yml)for the image to be built automatically.

agents/synacc/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# SOCS LabJack Agent
+# socs Agent container for running the LabJack Agent.
```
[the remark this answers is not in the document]
JakeSpisak: fixed

docs/agents/synacc.rst
```diff
@@ -0,0 +1,82 @@
...
+          ]}
+
+Docker
+``````
+The SynAcc Agent should be configured to run in a Docker container.
+An example docker-compose service configuration is shown here::
+
+  ocs-synacc:
+    image: synacc-img
```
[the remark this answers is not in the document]
JakeSpisak: done

agents/synacc/synacc.py
```diff
@@ -0,0 +1,115 @@
...
+import requests
+import argparse
+
+on_rtd = os.environ.get('READTHEDOCS') == 'True'
+if not on_rtd:
+    from ocs import ocs_agent, site_config
+    from ocs.ocs_twisted import TimeoutLock
+
+
+class SynAccAgent:
```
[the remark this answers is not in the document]
JakeSpisak: Done, and changed in all other files that SynAcc was used

agents/synacc/synacc.py
JakeSpisak: I will add this in in the next iteration that includes the acquisition proccess

docs/agents/synacc.rst
JakeSpisak: Yeah admin/admin is the default. You can change it in the web browser (it has a typical IP address site)

agents/synacc/synacc.py
JakeSpisak: Will change in the next iteration: see comment below

JakeSpisak:
We didn't make changes 1) adding session.data  2) adding the acquisition process and 3) the 'request' function, in the interest of getting this first version pulled ASAP, and also because I have a question for you on 1) and 2). My understanding of session.data is that it completely duplicates the 'get_status' task currently in the agent. So my proposal is to delete that task, and create a new process that updates a feed as well as session.data every time an outlet changes. I was thinking that the process could continually (every 10s or so) check the outlet status, and if it has changed, publish to the feed and update session.data. What do you think?

docker-compose.yml
```diff
@@ -94,6 +94,13 @@ services:
   ocs-pfeiffer-tpg366-agent:
     image: "ocs-pfeiffer-tpg366-agent"
     build: ./agents/pfeiffer_tpg366/
+    
+  # --------------------------------------------------------------------------
+  # Synaccess power strip agent
+  # --------------------------------------------------------------------------
+  ocs-synaccess-agent:
+    image: "ocs-synaccess-agent:"
```
BrianJKoopman, line 102: There's a syntax error here, remove that trailing `:`.

BrianJKoopman:
> We didn't make changes 1) adding session.data 2) adding the acquisition process and 3) the 'request' function, in the interest of getting this first version pulled ASAP, and also because I have a question for you on 1) and 2). 

Sounds good, we can go ahead and merge this and add these later.

> My understanding of session.data is that it completely duplicates the 'get_status' task currently in the agent. So my proposal is to delete that task, and create a new process that updates a feed as well as session.data every time an outlet changes. I was thinking that the process could continually (every 10s or so) check the outlet status, and if it has changed, publish to the feed and update session.data. What do you think?

While you have access to the outlet information in the message string you write out in the `get_status` return `return True, 'status outlet [1,2,3,4,5] is {}'.format(ret_str)`, that's meant for human reading in the logs. Putting that information in a dictionary within `session.data` would allow for say, easily getting that return value in a Client, be that a python script, or the Javascript web interface.

I'm not against the long running process, but was just suggesting you publish to a feed from the tasks that change outlet state. The advantage a long running process might bring is if someone changes an outlet state not through OCS, the change would still get logged.

One last thing, `session.data` is unique per task, so `session.data` in `get_status` is not the same `session.data` in `set_outlet`.

docker-compose.yml
JakeSpisak: fixed

JakeSpisak:
Ah okay I understand your point about session.data being a storable value in get_status. And yeah I can just add to a feed for each of the tasks that change the output: the use case of people changing the outlet without OCS is probably pretty minimal, at least once people start using the agent

BrianJKoopman (approved):
Great, thanks. Merging, I'll open issues for features to include in the next version.

PR #104, opened 2020-08-27
Labjack ljm: add ljm module, downloaded from remote source

JakeSpisak:
## Description
The labjack LJM module is used to query the labjack. This is much faster than the previous method, enabling readout tested up to 700Hz. It requires downloading custom labjack software. Fortunately, labjack created their own ljm-enabled image, so I was able to copy their dockerfile commands to use the software in this image.

## Motivation and Context
We needed faster labjack readout for vibration testing.

## How Has This Been Tested?
I tested readout functionality on multiple channels at several frequencies. The client start/stop acq processes work. I also confirmed that the channel conversion functions still work. Due to channel name standardization, the labjack function .yaml file now requires underscores in the channel names (that's why this is a 'breaking change').

## Types of changes

## Checklist:


agents/labjack/Dockerfile
```diff
@@ -9,6 +9,18 @@ WORKDIR /app/socs/agents/labjack/
 
 ## Copy this agent into the app/agents directory
 COPY . .
+RUN pip3 install -r requirements.txt
+
+# Install ljm module from labjack
```
BrianJKoopman, line 14: Can we add a comment to link to the labjack/ljm image on Dockerhub for reference? It's this one, yeah?
https://hub.docker.com/r/labjack/ljm

That gives context for the Dockerfile additions and provides a good reference for if we need future updates that they might implement.

BrianJKoopman (approved):
This is great, thanks Jake. Glad to see LabJack has a Docker image for this. One minor suggestion to link to that image, then this can be squashed and merged.

agents/labjack/Dockerfile
JakeSpisak: Sure, just did it in the latest commit

PR #105, opened 2020-08-27
Adds stream_id component to SmurfRecorder file path

jlashner:
This PR modifies the smurf-recorder file path to include the stream-id.

## Description

Based on [this](http://simonsobservatory.wdfiles.com/local--files/smurfdaq-user-interface-2020/readout_data_storage.pdf?ukey=4b10d8a49a529d110dc49cb2377d031215082f6d) presentation that Marius gave at the last Smurf-DAQ workshop, detector timestreams should be organized according to 
```
<data_dir>/<date-code>/<stream-id>/files
```
Currently data is just recorded as `<data_dir>/<date-code>/files`, so this must be updated before we can operate with multiple slots. 

This PR adds the `stream_id` variable to the path, which the SmurfRecorder will either pull from a G3Frame if possible, or use a default provided as a command line argument. In the next release of the `smurf-streamer` all frames will be tagged with the stream_id, however when running with the current iteration of the software, a default will need provided in the site-config.

This PR also includes modifications to the StreamSimulator to tag frames with a stream_id.

## Motivation and Context

Modifies SmurfRecorder path to accommodate multiple slots and crates.

## How Has This Been Tested?

I tested these changes by running the modified versions of the StreamSimulator, and the SmurfRecorder and verified that the new filepaths were correct.

## Types of changes

## Checklist:

### DIFF
--- a/agents/smurf_stream_simulator/smurf_stream_simulator.py
+++ b/agents/smurf_stream_simulator/smurf_stream_simulator.py
@@ -72,6 +72,8 @@ class SmurfStreamSimulator:
         Port to send data over
     num_chans : int
         Number of channels to simulate
+    stream_id : str
+        Stream ID to put into G3Frames. Defaults to "stream_sim"
 
     Attributes
     ----------
@@ -95,12 +97,15 @@ class SmurfStreamSimulator:
         List of simulated channels to stream
 
     """
-    def __init__(self, agent, target_host="*", port=4536, num_chans=528):
+    def __init__(self, agent, target_host="*", port=4536, num_chans=528,
+                 stream_id='stream_sim'):
         self.agent = agent
         self.log = agent.log
         self.target_host = target_host
 
         self.port = port
+
+        self.stream_id = stream_id
 
         self.writer = None
         self.is_streaming = False
@@ -164,6 +169,7 @@ class SmurfStreamSimulator:
                 f = core.G3Frame(core.G3FrameType.Scan)
                 f['session_id'] = 0
                 f['frame_num'] = frame_num
+                f['sostream_id'] = self.stream_id
                 f['data'] = core.G3TimestreamMap()
 
                 for i, chan in enumerate(self.channels):
@@ -211,6 +217,7 @@ class SmurfStreamSimulator:
             f = core.G3Frame(core.G3FrameType.Observation)
             f['session_id'] = 0
             f['start_time'] = time.time()
+            f['sostream_id'] = self.stream_id
             self.writer.Process(f)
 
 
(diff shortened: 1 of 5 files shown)


agents/smurf_stream_simulator/smurf_stream_simulator.py
```diff
@@ -164,6 +169,7 @@ def start_background_streamer(self, session, params=None):
                 f = core.G3Frame(core.G3FrameType.Scan)
                 f['session_id'] = 0
                 f['frame_num'] = frame_num
+                f['sostream_id'] = self.stream_id
```
BrianJKoopman, line 172: How about 'stream_id' instead of 'sostream_id'? (Unless this is already established somewhere like the TOD format document?) That would keep it consistent with "stream-id", etc. elsewhere and "so" feels redundant.

socs/agent/smurf_recorder.py
```diff
@@ -31,6 +31,8 @@ def _create_dirname(start_time, data_dir):
         Timestamp for start of data collection
     data_dir : str
         Top level data directory for output
+    stream_id : str
+        Stream id of colleciton
```
BrianJKoopman, line 35: *collection

BrianJKoopman (approved):
This looks good to me. My one suggestion was related to the naming of the 'sostream_id' key within the Frames, but I see that's widely used in the smurf-streamer. Does the SMuRF have some separate 'stream_id' that needs to be distinguished from the "so" stream_id?

Some misc questions -- Are there batches of old files that don't have stream_id's that'll need to be handled? That'll just be some directory creating and file moving? Is the stream_id going to be used for anything besides this directory creation? Marius' presentation calls this the 'system-id' and gives the example 'crate0slot2'. Will stream_id's always be of this form?

agents/smurf_stream_simulator/smurf_stream_simulator.py
jlashner: I chose 'sostream_id' to match the 'sostream_version' and 'sostream_flowcontrol' keys described here: https://github.com/simonsobs/tod2maps_docs/pull/4. I'd be fine switching but the key is already written into a lot of G3Files since it is set by the smurf-streamer... If this is really important to you we can figure out how to change it but it will be much easier to keep it as is unfortunately.

BrianJKoopman: Ah, makes sense. We should update that tod_format document with this addition.

jlashner:
Thank for the comments Brian! 

Yeah, existing archives will need to be dealt with in some way. The simplest way might be to not touch the archives, and to just give the reader the ability to set the stream_id to some default if none exists in the frames. If we want to be proper about it, we'd have to add the `stream_id` subdirectory to the archive, and tag each frame with the `sostream_id` key.

PR #110, opened 2020-10-05
Lakeshore session.data

BrianJKoopman:
## Description

This PR adds the latest measurements from the acq Processes for each Lakeshore Agent to the Process' `session.data` object.

## Motivation and Context

Allows most recent data to be queried easily in OCS clients.

## How Has This Been Tested?

Built and ran with either a 240 simulator or actual 372 in the lab at Yale and viewed with ocs-web.

## Types of changes

## Checklist:

### DIFF
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -212,8 +212,7 @@ class LS372_Agent:
             previous_channel = None
             last_release = time.time()
 
-            # Returned in session.data
-            data_cache = {}
+            session.data = {"fields": {}}
 
             self.take_data = True
             while self.take_data:
@@ -293,15 +292,17 @@ class LS372_Agent:
                                                         chan=active_channel.channel_num)
                     res_reading = self.module.get_temp(unit='ohms',
                                                        chan=active_channel.channel_num)
+
+                    # For data feed
                     data['data'][channel_str + '_T'] = temp_reading
                     data['data'][channel_str + '_R'] = res_reading
 
+                    # For session.data
                     field_dict = {channel_str: {"T": temp_reading,
                                                 "R": res_reading,
                                                 "timestamp": current_time}}
 
-                    data_cache.update(field_dict)
-                    session.data = {'fields': data_cache}
+                    session.data['fields'].update(field_dict)
 
                 session.app.publish_to_feed('temperatures', data)
 
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -291,6 +291,7 @@ class LS372_Agent:
 
                 session.data = {'fields': data['data'],
                                 'timestamp': current_time}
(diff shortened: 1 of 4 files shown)


agents/lakeshore372/LS372_agent.py
```diff
@@ -175,6 +175,21 @@ def init_lakeshore_task(self, session, params=None):
...
+        Method to start data acquisition process.
+
+        The most recent data collected is stored in session.data in the
+        structure::
+
+            >>> session.data
+            {"fields":
+                {"Channel_01_T": 1.34,
+                 "Channel_01_R": 21234.02},
+             "timestamp": 1601919441.9610307}
```
jlashner, line 189: Can we make it so `session.data` contains the most recent data point taken for each channel? So like:
```
 {"fields": {
       "Channel_01": {"T": 1.34, "R": 21234.02, "timestamp": 1601919441.9610307},
       "Channel_02": {"T": 4.34, "R": 12345, "timestamp": 1601919480},
        ....
}}
```

jlashner (changes requested):
This looks great, however I think it would make sense for the 372 to return the most recent temperature for each channel, instead of just the most recent channel that it's read. That way clients can get an estimate of all channel temperatures without waiting for a complete scan cycle.

agents/lakeshore372/LS372_agent.py
BrianJKoopman: Those are identical, as they're sampled at the same time.

BrianJKoopman: Yeah, can do. This strays a bit from the actual field names as stored in the .g3 files, but I think makes the most sense.

BrianJKoopman:
Updated with the suggestion structure, and adapted the 240 Agent to have a similar one so that usage is uniform across Lakeshore devices.

agents/lakeshore372/LS372_agent.py
```diff
@@ -258,21 +280,33 @@ def start_acq(self, session, params=None):
...
+                                                       chan=active_channel.channel_num)
+                    data['data'][channel_str + '_T'] = temp_reading
+                    data['data'][channel_str + '_R'] = res_reading
+
+                    field_dict = {channel_str: {"T": temp_reading,
+                                                "R": res_reading,
+                                                "timestamp": current_time}}
+
+                    data_cache.update(field_dict)
+                    session.data = {'fields': data_cache}
```
jlashner, line 304: I think this probably works, but since `session.data` is persistent between loops, you could also just initialize session.data with 
```session.data = {'fields': {}}```
and update with:
```
session.data['fields']['channel_str'] = {
    "T": temp_reading,
    "R": res_reading,
    "timestamp": current_time
}
```
instead of keeping a `data_cache`.

jlashner (approved):

jlashner:
Looks great, thanks brian!

PR #113, opened 2020-10-21
Addes stream monitor to the smurf_recorder agent

jlashner:
This PR adds functionality to the smurf_recorder agent to monitor specified readout channels, downsample incoming data for those channels, and sends them to grafana.

## Description

This adds two new tasks to the smurf recorder agent. `set_monitored_channels` will set update the list of channels to stream, and `set_target_rate` sets the sample rate that the detector data will be downsampled to.

It also adds a function `read_stream_data` to the `FrameRecorder` class that loops through incoming frames and records downsampled data for the monitored channels so it can be published by the agent.

This PR also changes the G3TimestreamMap keys in the smurf_stream_simulator to be the same as the data currently passed by the real smurf-streamer (e.g. 'r0012' instead of '12').

## Motivation and Context

This will allow detector data to be viewed in grafana, which will be very useful for RF and optical tests.

## How Has This Been Tested?

I've tested by manually running the smurf_recorder modified smurf_stream_simulator locally and was able to view the monitored timestreams on grafana, and verified that `set_monitored_channels` and `set_target_rate` both work as expected. 

## Types of changes

## Checklist:

### DIFF
--- a/agents/smurf_recorder/smurf_recorder.py
+++ b/agents/smurf_recorder/smurf_recorder.py
@@ -84,9 +84,9 @@ class SmurfRecorder:
 
         Args
         -----
-        channels : list, required
-            List of channel numbers to be monitored.
-            A maximum of 6 channels can be monitored at a time.
+        channels : List[Int], required
+            List of channel numbers (integers between 0 and 2048) to be
+            monitored. A maximum of 6 channels can be monitored at a time.
         """
         if params is None:
             params = {}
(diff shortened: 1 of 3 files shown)


BrianJKoopman (commented):
This is looking good, thanks Jack. My only non-documentation comment relates to the new tasks and whether they should be run in the reactor, rather than separate threads.

jlashner:
Hey Brian, good point about putting the tasks in the reactor thread. Thanks! Made the changes and re-tested.

BrianJKoopman (approved):
Great, thanks the changes, approved!

Issue #122, opened 2020-11-25
smurf_recorder index step size out of range


jlashner:
When the smurf-recorder is monitoring a channel and there is a short data file, the record process crashes with this error: 
```
ocs-smurf-recorder-s2_1   | 2020-11-25T02:26:18+0000 Writing to file /data/timestreams/16062/crate1slot2/1606271178_000.g3
ocs-smurf-recorder-s2_1   | FATAL (G3Timestream): Step index 19 out of range (G3Timestream.cxx:824 in G3TimestreamPtr {anonymous}::G3Timestream_getslice(const G3Timestream&, boost::python::slice))
ocs-smurf-recorder-s2_1   | 2020-11-25T02:26:36+0000 Unable to format event {'log_logger': <Logger 'ocs.ocs_agent.OCSAgent'>, 'log_level': <LogLevel=info>, 'log_namespace': 'ocs.ocs_agent.OCSAgent', 'log_source': None, 'log_format': "record:4 Crash in thread: [Failure instance: Traceback: <class 'RuntimeError'>: Step index 19 out of range (in G3TimestreamPtr {anonymous}::G3Timestream_getslice(const G3Timestream&, boost::python::slice))\n/usr/lib/python3.6/threading.py:916:_bootstrap_inner\n/usr/lib/python3.6/threading.py:864:run\n/usr/local/lib/python3.6/dist-packages/twisted/_threads/_threadworker.py:46:work\n/usr/local/lib/python3.6/dist-packages/twisted/_threads/_team.py:190:doWork\n--- <exception caught here> ---\n/usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:250:inContext\n/usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:266:<lambda>\n/usr/local/lib/python3.6/dist-packages/twisted/python/context.py:122:callWithContext\n/usr/local/lib/python3.6/dist-packages/twisted/python/context.py:85:callWithContext\nsmurf_recorder.py:142:start_record\n/app/socs/socs/agent/smurf_recorder.py:442:run\n/app/socs/socs/agent/smurf_recorder.py:473:read_stream_data\n/app_lib/spt3g_software/build/spt3g/core/timestreamextensions.py:13:G3Timestream_getitem\n]", 'log_time': 1606271196.0203595}: 'anonymous'.
```

Digging into spt3g, this points to this segment:
```
  808     if (step >= a.size() || step <= 0)
[...]

PR #123, opened 2020-11-25
Clamp ds_factor in smurf recorder

jlashner:
## Description

Clamps the downsample factor in smurf_recorder's stream monitoring function, solving #122.

## Motivation and Context

Slicing with step size larger than the length of the would cause spt3g to log a fatal error.

## How Has This Been Tested?

Tested running this locally with the smurf-recorder with `target_rate=0.1` and short file sizes.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agent/smurf_recorder.py
+++ b/socs/agent/smurf_recorder.py
@@ -461,6 +461,9 @@ class FrameRecorder:
             ds_factor = (frame['data'].sample_rate/core.G3Units.Hz) \
                         // self.target_rate
             ds_factor = max(int(ds_factor), 1)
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
This looks alright to me. You'll always be grabbing the last point in the array where you wouldn't have previously, but I think that's okay. I do agree with @mhasself in his comment on https://github.com/simonsobs/socs/issues/122#issuecomment-733684238 that we should probably fix this upstream so that behavior is consistent with what one would expect in Python.

jlashner:
Thanks Brian, I'll merge and post an issue to spt3g_software.

Issue #126, opened 2020-12-15
Smurf Recorder crash if sample_rate is nan


jlashner:
Got the following error from the k2so smurf-recorder:
```
ocs-smurf-recorder-s2_1   | 2020-12-15T19:49:31+0000 record:0 Crash in thread: [Failure instance: Traceback: <class 'ValueError'>: cannot convert float NaN to integer
ocs-smurf-recorder-s2_1   | /usr/lib/python3.6/threading.py:916:_bootstrap_inner
ocs-smurf-recorder-s2_1   | /usr/lib/python3.6/threading.py:864:run
ocs-smurf-recorder-s2_1   | /usr/local/lib/python3.6/dist-packages/twisted/_threads/_threadworker.py:46:work
ocs-smurf-recorder-s2_1   | /usr/local/lib/python3.6/dist-packages/twisted/_threads/_team.py:190:doWork
ocs-smurf-recorder-s2_1   | --- <exception caught here> ---
ocs-smurf-recorder-s2_1   | /usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:250:inContext
ocs-smurf-recorder-s2_1   | /usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:266:<lambda>
ocs-smurf-recorder-s2_1   | /usr/local/lib/python3.6/dist-packages/twisted/python/context.py:122:callWithContext
ocs-smurf-recorder-s2_1   | /usr/local/lib/python3.6/dist-packages/twisted/python/context.py:85:callWithContext
ocs-smurf-recorder-s2_1   | smurf_recorder.py:142:start_record
ocs-smurf-recorder-s2_1   | /app/socs/socs/agent/smurf_recorder.py:442:run
ocs-smurf-recorder-s2_1   | /app/socs/socs/agent/smurf_recorder.py:463:read_stream_data
ocs-smurf-recorder-s2_1   | ]
```
Referring to the lines:
```
         ds_factor = (frame['data'].sample_rate/core.G3Units.Hz) \
                     // self.target_rate
         ds_factor = max(int(ds_factor), 1)
```
I'm guessing that the sample_rate is nan somehow, but have not really looked into it. Perhaps this is a reason why the channel monitor should be in a separate agent... or maybe we should wrap the whole thing in a try-catch to keep data-specific errors from crashing the record process.

jlashner:
Closed by https://github.com/simonsobs/socs/pull/130

PR #130, opened 2021-01-05
Handle nan in smurf_recorder read_stream_data

jlashner:
## Description
This is a small fix to handle errors in the `read_stream_data` function without crashing the streamer. It also changes the smurf-recorder so that `read_stream_data` is only run if there are monitored channels.

## Motivation and Context
Solves #126 .

## How Has This Been Tested?
Tested by running on the k2so system, testing with a sequence of commands that was causing the sample_rate to be nan consistently. 

## Types of changes

## Checklist:

### DIFF
--- a/socs/agent/smurf_recorder.py
+++ b/socs/agent/smurf_recorder.py
@@ -3,6 +3,7 @@ from enum import Enum
 
 import time
 import txaio
+import numpy as np
 
 # For logging
 txaio.use_twisted()
@@ -439,7 +440,12 @@ class FrameRecorder:
         """
         self.read_frames()
         self.check_for_frame_gap(10)
-        self.read_stream_data()
+        if len(self.monitored_channels) > 0:
+            try:
+                self.read_stream_data()
+            except Exception as e:
+                self.log.warn("Exception thrown when reading stream data:\n{e}", e=e)
+
         if self.frames:
             self.create_new_file()
             self.write_frames_to_file()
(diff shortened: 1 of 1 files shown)


BrianJKoopman (commented):
This'll stop #126 from crashing the recorder, but do you know what causes `ds_factor` to be `NaN` sometimes? What command sequence reproduces the error?

jlashner:
I was running the following code that produced the error consistently:
```
for c in np.arange(100, 4000, 100):
    print(f"Enabling {c} chans")
    mask = np.arange(0, c)
    S.set_channel_mask(mask)
    S.set_stream_enable(1)
    time.sleep(10)
    S.set_stream_enable(0)
    time.sleep(20)
```
My guess is that changing the channel mask forces a sample buffer flush in the streamer, which might be producing timestreams with a single sample.

[Here's]( https://github.com/CMB-S4/spt3g_software/blob/b6c2c52f33c3eb0b24391a5ccec25cd9528838ab/core/src/G3Timestream.cxx#L233-L236) the function that gets sample rate in spt3g:
```
double G3Timestream::GetSampleRate() const
{
	return double(size() - 1)/double(stop.time - start.time);
}
```
It looks like if there is only one sample in the timestream this will produce a nan.

BrianJKoopman (approved):
Ah, that makes sense, can't give a sample rate of a single sample. Thanks for the explanation!

Issue #78, opened 2020-05-28
Smurf recorder quits on jackhammer soft_reset


jlashner:
Looks like the smurf-recorder at UCSD is shutting down whenever I run jackhammer soft_reset. Strangely it seems like it is handling the disconnection/reconnection well, so I'm not sure exactly why it's crashing... Here's my full output
```
adaq1@adaq1:~/so/ocs-site-configs/ucsd/adaq1$ docker-compose up ocs-smurf-recorder-s2
adaq1_sisock-crossbar_1 is up-to-date
Starting adaq1_ocs-smurf-recorder-s2_1 ... done
Attaching to adaq1_ocs-smurf-recorder-s2_1
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 site_config is setting values of "auto_start" to "True".
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 site_config is setting values of "time_per_file" to "600".
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 site_config is setting values of "data_dir" to "/data/timestreams".
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 site_config is setting values of "port" to "4532".
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 site_config is setting values of "address" to "192.168.0.134".
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 Using OCS version 0.6.0+1.g569b428
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 ocs: starting <class 'ocs.ocs_agent.OCSAgent'> @ observatory.smurf-recorder-s2
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 log_file is apparently None
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 transport connected
ocs-smurf-recorder-s2_1   | 2020-05-28T01:59:59+0000 session joined:
ocs-smurf-recorder-s2_1   | SessionDetails(realm=<test_realm>,
ocs-smurf-recorder-s2_1   |                session=3069769554717731,
ocs-smurf-recorder-s2_1   |                authid=<KPSS-CYPK-WJN6-YWGN-Y33C-M6EG>,
ocs-smurf-recorder-s2_1   |                authrole=<server>,
[...]

PR #134, opened 2021-01-20
Add some quality of life improvements to smurf-rec

### LINKED ISSUE #112: Smurf-recorder setting stream-id in path to None
Sometimes the smurf-recorder is setting the stream-id component in the data path to None even if there is a stream-id present in the frames. I think it does this when it receives the PipelineInfo frame at the start, since that doesn't have the `sostream_id` key. 

For example, on the Penn system the file /data/timestreams/16022/None/1602279295_000.g3 contains the following frames:
```
Frame (PipelineInfo) [
"08-Oct-2020:17:18:10.069906000" (spt3g.core.G3PipelineInfo) => master branch, no local diffs
]
Frame (Observation) [
"frame_num" (spt3g.core.G3Int) => 0
"session_id" (spt3g.core.G3Int) => 1602187012
"sostream_id" (spt3g.core.G3String) => "crate1slot2"
"sostream_version" (spt3g.core.G3Int) => 1
"time" (spt3g.core.G3Time) => 08-Oct-2020:19:56:52.054788000
]
Frame (Wiring) [
"dump" (spt3g.core.G3Int) => 0
"frame_num" (spt3g.core.G3Int) => 170
"session_id" (spt3g.core.G3Int) => 1602187012
"sostream_id" (spt3g.core.G3String) => "crate1slot2"
"sostream_version" (spt3g.core.G3Int) => 1
"status" (spt3g.core.G3String) => "AMCc.SmurfProcessor.ChannelMapper.NumChannels: 63
AMCc.SmurfProcessor.PostDataEmulator.Offset: 0
AMCc.SmurfProcessor.PostDataEmulator.Amplitude: 4294967295
"
"time" (spt3g.core.G3Time) => 08-Oct-2020:19:57:27.892115000
]
```

jlashner:
## Description
Changes consist of:
1. Check if socket is open before trying to create G3Streamer to
   minimize log_fatals
2. Change some info statents into debug, and some debug statements to
   info
3. Discards G3PipelineInfo frames to fix stream-id-less file on startup
   bug.

## Motivation and Context

This fixes some small bugs in the smurf-recorder and makes it easier to monitor during operation. Mainly did this in order to further debug issue #78, but this should also resolve #112.

## How Has This Been Tested?

Tested at UCSD by hammering while the smurf-recorder was running.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agent/smurf_recorder.py
+++ b/socs/agent/smurf_recorder.py
@@ -232,7 +232,7 @@ class FrameRecorder:
         if (check_port(self.host, self.port)):
             try:
                 reader = core.G3Reader(self.address,
-                                       timeout=0)
+                                       timeout=timeout)
                 self.log.debug("G3Reader connection to {addr} established!",
                               addr=self.address)
             except RuntimeError:
--- a/socs/agent/smurf_recorder.py
+++ b/socs/agent/smurf_recorder.py
@@ -4,7 +4,6 @@ from enum import Enum
 import time
 import txaio
 import numpy as np
-import sys
 import socket
 
 # For logging
@@ -23,14 +22,29 @@ class FlowControl(Enum):
     CLEANSE = 3
 
 
-def check_port(host, port, timeout=10):
+def check_port(addr):
+    """
+    This function checks if socket port is currently open on a host. This can
+    be used to check if the smurf-streamer has created it's G3NetworkSender
+    object without attempting to create a G3Reader. This function is
+    non-blocking and should return immediately.
+
+    Parameters
+    ----------
+    addr: str
+        Address describing the port to connect to. For example:
+        ``tcp://localhost:4532``
+    """
+    host, port = addr.split('//')[-1].split(':')
+    port = int(port)
+
     s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-    s.settimeout(timeout)
-    s.setblocking(1)
+    s.setblocking(True)
+    s.settimeout(0.0)
     try:
         s.connect((host, port))
         return True
-    except socket.error as error:
+    except socket.error:
         return False
     finally:
         s.close()
@@ -133,6 +147,8 @@ class FrameRecorder:
         G3Reader object to read the frames from the G3NetworkSender.
     writer : spt3g.core.G3Writer
         G3Writer for writing the frames to disk.
+    data_received : bool
+        Whether data has been received by the current instance of the G3Reader.
     frames : list
         List of frames that have been read from the network. Gets cleared after
         writing to file.
@@ -177,8 +193,6 @@ class FrameRecorder:
         # Parameters
         self.time_per_file = file_duration
         self.address = tcp_addr
-        self.host, self.port = self.address[6:].split(':')
-        self.port = int(self.port)
         self.data_dir = data_dir
         self.stream_id = stream_id
         self.log = txaio.make_logger()
@@ -229,15 +243,14 @@ class FrameRecorder:
         """
         reader = None
 
-        if (check_port(self.host, self.port)):
+        if (check_port(self.address)):
             try:
                 reader = core.G3Reader(self.address,
                                        timeout=timeout)
                 self.log.debug("G3Reader connection to {addr} established!",
-                              addr=self.address)
+                               addr=self.address)
             except RuntimeError:
                 self.log.error("G3Reader could not connect.")
-
 
         # Prevent rapid connection attempts
         if self.last_connection_time is not None:
@@ -302,13 +315,14 @@ class FrameRecorder:
                            if x.type != core.G3FrameType.PipelineInfo]
             if self.frames and not self.data_received:
                 self.data_received = True
-                self.log.info("Frames received from {addr}", addr=self.address)
+                self.log.info("Started receiving frames from {addr}",
+                              addr=self.address)
             return
         else:
             if self.data_received:
                 self.log.info("Could not read frames. Connection " +
-                               "timed out, or G3NetworkSender offline. " +
-                               "Cleaning up...")
+                              "timed out, or G3NetworkSender offline. " +
+                              "Cleaning up...")
             self.close_file()
             self.data_received = False
             self.reader = None
--- a/socs/agent/smurf_recorder.py
+++ b/socs/agent/smurf_recorder.py
@@ -4,6 +4,8 @@ from enum import Enum
 import time
 import txaio
 import numpy as np
+import sys
+import socket
 
 # For logging
 txaio.use_twisted()
@@ -19,6 +21,19 @@ class FlowControl(Enum):
     START = 1
     END = 2
     CLEANSE = 3
+
+
+def check_port(host, port, timeout=10):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+    s.setblocking(1)
+    try:
+        s.connect((host, port))
+        return True
+    except socket.error as error:
+        return False
+    finally:
+        s.close()
 
 
 def _create_dirname(start_time, data_dir, stream_id):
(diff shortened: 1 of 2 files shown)


jlashner:
Looks like tests passed, so whenever you review @BrianJKoopman I'll go ahead and merge.

socs/agent/smurf_recorder.py
```diff
@@ -21,6 +23,19 @@ class FlowControl(Enum):
     CLEANSE = 3
 
 
+def check_port(host, port, timeout=10):
```
BrianJKoopman, line 26: Can we add a short docstring to this?

socs/agent/smurf_recorder.py
```diff
@@ -21,6 +23,19 @@ class FlowControl(Enum):
...
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+    s.setblocking(1)
```
BrianJKoopman, line 29: If I'm reading this right, isn't `s.setblocking(1)` equivalent to `s.settimeout(None)`. It seems like then the line above isn't really doing anything.

socs/agent/smurf_recorder.py
```diff
@@ -162,13 +177,16 @@ def __init__(self, file_duration, tcp_addr, data_dir, stream_id,
         # Parameters
         self.time_per_file = file_duration
         self.address = tcp_addr
+        self.host, self.port = self.address[6:].split(':')
```
BrianJKoopman, line 180: Either add these to the Attributes list in the docstring, or better yet make `check_port()` take the format stored in `self.address`.

socs/agent/smurf_recorder.py
```diff
@@ -162,13 +177,16 @@ def __init__(self, file_duration, tcp_addr, data_dir, stream_id,
...
+        self.port = int(self.port)
         self.data_dir = data_dir
         self.stream_id = stream_id
         self.log = txaio.make_logger()
 
         # Reader/Writer
         self.reader = None
         self.writer = None
+        self.data_received = False
```
BrianJKoopman, line 189: Also should be added to the Attributes list.

socs/agent/smurf_recorder.py
```diff
@@ -210,12 +228,16 @@ def _establish_reader_connection(self, timeout=5):
...
         """
         reader = None
-        try:
-            reader = core.G3Reader(self.address,
-                                   timeout=timeout)
-            self.log.info("G3Reader connection established")
-        except RuntimeError:
-            self.log.error("G3Reader could not connect.")
+
+        if (check_port(self.host, self.port)):
```
BrianJKoopman, line 232: Since `check_port` has `s.setblocking(1)` does this wait here until a connection is made?

The code below here was written to prevent rapid connection attempts when a connection couldn't be made. Do we run into that situation anymore if we wait to connect until `check_port` passes?

socs/agent/smurf_recorder.py
```diff
@@ -275,13 +297,20 @@ def read_frames(self, timeout=5):
 
             # Discard all flow control frames
             self.frames = [x for x in self.frames if 'sostream_flowcontrol' not in x]
-
+            # Discard Pipeline info frame
+            self.frames = [x for x in self.frames
+                           if x.type != core.G3FrameType.PipelineInfo]
+            if self.frames and not self.data_received:
+                self.data_received = True
+                self.log.info("Frames received from {addr}", addr=self.address)
```
BrianJKoopman, line 305: Since we'll only see this at the start of receiving frames, not every time frames are received, can we change the message to reflect that? That's to avoid someone looking at the logs, seeing only one line of "Frames received from {addr}" and no more, thinking that frames aren't still coming in. Maybe something like "Started receiving frames from {addr}".

socs/agent/smurf_recorder.py
```diff
@@ -4,6 +4,8 @@
 import time
 import txaio
 import numpy as np
+import sys
```
BrianJKoopman, line 7: Unused import.

socs/agent/smurf_recorder.py
```diff
@@ -21,6 +23,19 @@ class FlowControl(Enum):
...
 
 
+def check_port(host, port, timeout=10):
+    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    s.settimeout(timeout)
+    s.setblocking(1)
+    try:
+        s.connect((host, port))
+        return True
+    except socket.error as error:
```
BrianJKoopman, line 33: `error` is assigned here but never used. Would it be useful to log that error if it arises?

socs/agent/smurf_recorder.py
```diff
@@ -210,12 +228,16 @@ def _establish_reader_connection(self, timeout=5):
...
+
+        if (check_port(self.host, self.port)):
+            try:
+                reader = core.G3Reader(self.address,
+                                       timeout=timeout)
+                self.log.debug("G3Reader connection to {addr} established!",
+                              addr=self.address)
+            except RuntimeError:
+                self.log.error("G3Reader could not connect.")
+
```
BrianJKoopman, line 240: Remove this extra empty line.

socs/agent/smurf_recorder.py
```diff
@@ -275,13 +297,20 @@ def read_frames(self, timeout=5):
...
+                self.data_received = True
+                self.log.info("Frames received from {addr}", addr=self.address)
             return
         else:
-            self.log.debug("Could not read frames. Connection " +
-                           "timed out, or G3NetworkSender offline. " +
-                           "Cleaning up...")
+            if self.data_received:
+                self.log.info("Could not read frames. Connection " +
+                               "timed out, or G3NetworkSender offline. " +
```
BrianJKoopman, line 310: Lines 310 and 311 are over-indented by 1 space.

socs/agent/smurf_recorder.py
```diff
@@ -210,12 +228,16 @@ def _establish_reader_connection(self, timeout=5):
...
-            self.log.info("G3Reader connection established")
-        except RuntimeError:
-            self.log.error("G3Reader could not connect.")
+
+        if (check_port(self.host, self.port)):
+            try:
+                reader = core.G3Reader(self.address,
+                                       timeout=timeout)
+                self.log.debug("G3Reader connection to {addr} established!",
+                              addr=self.address)
```
BrianJKoopman, line 237: This is under-indented by 1 space.

BrianJKoopman (changes requested):
This is almost ready, needs some docstring updates, and I just want to check my understanding of what `check_port`'s blocking/timeout is doing. Also some very small PEP8 related formatting.

socs/agent/smurf_recorder.py
jlashner: Yeah, looking a bit more into and I am starting to think that I should just be using `s.setblocking(False)`, which would return immediately if the connect attempt fails. Like you say in another comment, the `_establish_reader_connection` function will manually prevent rapid connection attempts.

jlashner: After testing a bit, it seems like when the port is closed it sends `ConnectionRefusedError` immediately regardless of the timeout and blocking. My guess is that timeout and blocking are mostly for reading and writing to the port, and the connection is handled differently by the system. I'll use `blocking=True`, and `timeout=0.`

socs/agent/smurf_recorder.py
jlashner: As described above, I think the connection was being refused immediately even with a non-zero timeout, so this code block is still needed.

socs/agent/smurf_recorder.py
jlashner: I don't think it's especially useful, and I'm not super inclined to add a debug statement since there's no log instance in scope. If you really think I should log something we can, but for now I'll just remove the error assignment.

jlashner:
Alright thanks for the comments! I think I addressed all of them, but let me know if there's anything else I should change before merging.

BrianJKoopman (approved):
The changes look good, thanks Jack! And thanks for clarifying the blocking/timeout bit.

PR #136, opened 2021-02-09
Catch labjack stream error

JakeSpisak:
## Description
If the labjack is disconnected, the data stream will still be active. The previous version of this agent would then throw an error when the agent is restarted, since you can't start a stream if one is already running. The fix here is to check for this error, and if it occurs, stop the stream first. (A cleaner way would be to check if the stream is running, but I couldn't figure out how to do that).

## Motivation and Context
See the description. I noticed this error at UCSD when the ethernet from the labjack was disconnected while the agent was running.

## How Has This Been Tested?
I reproduced the error without the agent changes implemented, then confirmed that the changes stopped the stream and sucessfully started it again.

## Types of changes

## Checklist:

### DIFF
--- a/agents/labjack/labjack_agent.py
+++ b/agents/labjack/labjack_agent.py
@@ -11,6 +11,7 @@ import numpy as np
 ON_RTD = os.environ.get('READTHEDOCS') == 'True'
 if not ON_RTD:
     from labjack import ljm
+    from labjack.ljm.ljm import LJMError
     from ocs import ocs_agent, site_config
     from ocs.ocs_twisted import TimeoutLock
 
@@ -270,8 +271,14 @@ class LabJackAgent:
 
             # Start the data stream. Use the scan rate returned by the stream,
             # which should be the same as the input scan rate.
-            scan_rate = ljm.eStreamStart(self.handle, scans_per_read, num_chs,
-                                         ch_addrs, scan_rate_input)
+            try:
+                scan_rate = ljm.eStreamStart(self.handle, scans_per_read, num_chs,
+                                             ch_addrs, scan_rate_input)
+            except LJMError: #in case the stream is running
+                print("Stopping previous stream")
+                ljm.eStreamStop(self.handle) 
+                scan_rate = ljm.eStreamStart(self.handle, scans_per_read, num_chs,
+                                             ch_addrs, scan_rate_input)
             print(f"\nStream started with a scan rate of {scan_rate} Hz.")
 
             cur_time = time.time()
--- a/agents/labjack/labjack_agent.py
+++ b/agents/labjack/labjack_agent.py
@@ -86,7 +86,6 @@ class LabJackFunctions:
     """
     def __init__(self):
         self.log = txaio.make_logger()
-        pass
 
     def unit_conversion(self, v_array, function_info):
         """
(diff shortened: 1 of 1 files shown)


agents/labjack/labjack_agent.py
```diff
@@ -270,8 +271,14 @@ def start_acq(self, session, params=None):
 
             # Start the data stream. Use the scan rate returned by the stream,
             # which should be the same as the input scan rate.
-            scan_rate = ljm.eStreamStart(self.handle, scans_per_read, num_chs,
-                                         ch_addrs, scan_rate_input)
+            try:
+                scan_rate = ljm.eStreamStart(self.handle, scans_per_read, num_chs,
+                                             ch_addrs, scan_rate_input)
+            except LJMError: #in case the stream is running
```
BrianJKoopman, line 277: There are quite a few different [error codes](https://github.com/labjack/labjack-ljm-python/blob/master/labjack/ljm/errorcodes.py) associated with LJMError. Can we log more info about the type of error raised, i.e.:
```
except LJMError as e:
    print(e)
    print("Stopping and restarting stream.")
```
The `print(e)` will print the associated errorString, i.e. `LJM library error code 1227 LJME_DEVICE_NOT_FOUND`.

BrianJKoopman (commented):
This is good, there are a few agents where similar robustness from network disconnects need to be handled. One inline comment about logging more of a descriptive message from the LJMError object, which should tell us more if some new error that isn't what you're currently seeing ever comes up.

One other thing that I'll mention, that doesn't have to happen now, is that print statements are fine for logging to Docker, however using a log handler provides some more flexibility in terms of providing log levels -- identifying information vs warnings vs errors, and providing the "debug" log level, logs you can turn on temporarily while debugging a problem, then turn back off when running normally.

You can see examples of how to do that in other Agents such as the [bluefors Agent](https://github.com/simonsobs/socs/blob/develop/agents/bluefors/bluefors_log_tracker.py) or the [influxdb publisher agent](https://github.com/simonsobs/ocs/blob/develop/agents/influxdb_publisher/influxdb_publisher.py).

At the top of those Agents you'll see a block like:
```
import txaio

# For logging
txaio.use_twisted()
LOG = txaio.make_logger()
```

Then throughout the logs are written with `LOG.info('message')`, `LOG.debug('message')`, `LOG.warn('message')`, and `LOG.error('messsage')`.

At the top of the `if __name__ == '__main__':` block you'll also need:
```
# Start logging
txaio.start_logging(level=environ.get("LOGLEVEL", "info"))
```
That will allow you to easily change the log level based on an environment variable in your docker-compose file.

I mention all this to say that if you did do this now, you would only need to add the lines I called out here, and change print statements to `LOG.info('message')`. Then you could distinguish the error you're now catching with `LOG.error(e)`. This has the added benefit that they are distinguished as errors when viewing logs in Loki/Grafana.

JakeSpisak:
The recent commit logs the error message and changes the logging format to what was requested. I tested the logs work in both the agent and helper class (I didn't reproduce the error, but that was a very minor change)

BrianJKoopman:
Great, thanks Jake. I added the environment variable configurable log level, so the log level for the Agent can be changed in the future if needed. Also fixed some small PEP08 related things while I was looking. Will merge when checks finish.

BrianJKoopman (approved):

PR #144, opened 2021-02-18
Prologix interface inheritance

BrianJKoopman:
## Description

This makes some changes so that the PSU and Tektronix driver code can both inherit from a common `GPIBInterface`. This GPIB interface is based on the `PrologixInterface` class. 

## Motivation and Context

There was a fair bit of duplicated code among these various classes. This is an attempt to simplify that interface and reduce the amount of duplicated code before another Prologix device gets integrated here. Since I'm not too familiar with the hardware, I'm not sure if the `GPIBInterface` could be eliminated in favor of putting everything in `PrologixInterface`. But maybe there is some distinction to be made here.

I removed some outdated commented out code, some related to the GPIB like functionality in the PrologixInterface, and some related to verbosity of the `tektronixInterface` class. It didn't look like it worked if uncommented. If we do want this verbose functionality it should be updated and readded. I did uncommented similar verbosity code in the `psuInterface` code, but also made the default `verbose=False`, so we shouldn't see it by default.

## How Has This Been Tested?

This hasn't been tested at all. It needs testing on the actual hardware. It's probably sufficient to test the functionality of the Agents, but more thorough testing is welcome too. @JakeSpisak could you test the PSU Agent and @mjrand the Tektronix Agent?

## Types of changes

## Checklist:

### DIFF
--- a/agents/scpi_psu/scpi_psu_agent.py
+++ b/agents/scpi_psu/scpi_psu_agent.py
@@ -2,7 +2,7 @@ import time
 import os
 import socket
 import argparse
-from socs.agent.scpi_psu_driver import psuInterface
+from socs.agent.scpi_psu_driver import PsuInterface
 
 on_rtd = os.environ.get('READTHEDOCS') == 'True'
 if not on_rtd:
@@ -40,7 +40,7 @@ class ScpiPsuAgent:
                 return False, "Could not acquire lock"
 
             try:
-                self.psu = psuInterface(self.ip_address, self.gpib_slot)
+                self.psu = PsuInterface(self.ip_address, self.gpib_slot)
                 self.idn = self.psu.identify()
             except socket.timeout as e:
                 self.log.error("PSU timed out during connect")
@@ -75,8 +75,8 @@ class ScpiPsuAgent:
                     }
 
                     for chan in [1, 2, 3]:
-                        data['data']["Voltage_{}".format(chan)] = self.psu.getVolt(chan)
-                        data['data']["Current_{}".format(chan)] = self.psu.getCurr(chan)
+                        data['data']["Voltage_{}".format(chan)] = self.psu.get_volt(chan)
+                        data['data']["Current_{}".format(chan)] = self.psu.get_curr(chan)
 
                     # self.log.info(str(data))
                     # print(data)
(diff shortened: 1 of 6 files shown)


mhasself:
I haven't looked at this closely but I'm a fan of the consolidation.

My preference (though it's possible this isn't declared clearly anywhere) is to have classes in full CamelCaps, i.e. "class PrologixInterface", "class TektronixInterface".

Also: in the current PR I see "class psuInterface" and also "class GPIBInterface", and this seems to be doing two different things with acronyms in class names.

When treating acronyms in that case, there is significant controversy about how to render acronyms.  I am strongly in the camp that believes that such acronyms should be treated as words and only the first letter capitalized: PsuInterface, GpibInterface, HttpServer etc.  That's controversial but I think I very strong case, uh if you will, can be made for it.  (The case I make is that these are silly: "HTTPToXMLTranslator", "CMBTOD", "FITSWCSExtractor".)

BrianJKoopman:
> My preference (though it's possible this isn't declared clearly anywhere) is to have classes in full CamelCaps, i.e. "class PrologixInterface", "class TektronixInterface".
> 
> Also: in the current PR I see "class psuInterface" and also "class GPIBInterface", and this seems to be doing two different things with acronyms in class names.

Yeah, I was avoiding touching the Agents themselves, just the backend code, though I'm in favor of making this uniform. I'll do that.

> When treating acronyms in that case, there is significant controversy about how to render acronyms. I am strongly in the camp that believes that such acronyms should be treated as words and only the first letter capitalized: PsuInterface, GpibInterface, HttpServer etc. That's controversial but I think I very strong case, uh if you will, can be made for it. (The case I make is that these are silly: "HTTPToXMLTranslator", "CMBTOD", "FITSWCSExtractor".)

Heh, yeah, I'm onboard with that (and clearly indecisive myself, feeling one way before changing it in c401527.)

JakeSpisak:
Ah sorry about the late reply, I'll check this with a PSU tomorrow

JakeSpisak (commented):
I tested the scpi psu changes on a power supply here, all the functionality was fine (setting current, voltage, and monitoring output). Consolidating these changes is great, thanks for doing this. Sorry the review took to long. Unfortunately the tektronix device is currently being used to run buttkicker tests on SAT1, but once that is done we will test it

JakeSpisak:
Okay Michael tested this and it works just fine on the tektronix agent, so I see no reason not to push the changes

JakeSpisak (approved):
Michael tested it on the tektronix agent and it worked fine, so I'm approving this PR

mjrand (approved):

BrianJKoopman:
> Michael tested it on the tektronix agent and it worked fine, so I'm approving this PR

Great, thanks Jake! Rebased on the latest develop, will merge once checks past.

Issue #100, opened 2020-08-04
Pfeiffer Agent Block structure errors


BrianJKoopman:
I don't have time to triage this right now, but this was seen on the UCSD system during their OCS update, and I believe Penn has seen it as well:
```
2020-08-04T16:00:30+0000 Unhandled Error
Traceback (most recent call last):
  File "pfeiffer_tpg366_agent.py", line 177, in <module>
    runner.run(agent, auto_reconnect=True)
  File "/usr/local/lib/python3.6/dist-packages/autobahn/twisted/wamp.py", line 410, in run
    reactor.run()
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
--- <exception caught here> ---
  File "/usr/local/lib/python3.6/dist-packages/twisted/internet/base.py", line 886, in runUntilCurrent
    f(*a, **kw)
  File "/app/ocs/ocs/ocs_feed.py", line 208, in publish_message
    b.append(message)
  File "/app/ocs/ocs/ocs_feed.py", line 34, in append
    raise Exception("Block structure does not match: {}".format(self.name))
builtins.Exception: Block structure does not match: pressures
```

SOCS Version (and Docker tag): v0.0.2-67-g71a8640
OCS Version: 0.6.0+1.g569b428

BrianJKoopman:
This might be proceeded by a timeout, could be the hardware losing its network connection for some reason. I got the agent back online at UCSD after restarting the container, but it very quickly ran into:
```
2020-08-04T20:27:06+0000 acq:0 Crash in thread: [Failure instance: Traceback: <class 'socket.timeout'>: timed out
/usr/lib/python3.6/threading.py:916:_bootstrap_inner                                      
/usr/lib/python3.6/threading.py:864:run
/usr/local/lib/python3.6/dist-packages/twisted/_threads/_threadworker.py:46:work           
/usr/local/lib/python3.6/dist-packages/twisted/_threads/_team.py:190:doWork
--- <exception caught here> ---                                                                 
/usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:250:inContext
/usr/local/lib/python3.6/dist-packages/twisted/python/threadpool.py:266:<lambda>
[...]

PR #145, opened 2021-02-22
Updates to Pfeiffer agent.

tanaybhandarkar:
It now catches when the monitor channels are turned off and prints a statement to the logs. Also sets pressure value to zero if the channel was off. Prior to this, it threw an error when querying pressure values when channels are off

## Description
I added an if statement that first queries the status of the channel. If the status is "on", then the agent goes and queries the pressure reading. If the channel status is off, it returns a pressure value of 0.

## Motivation and Context
The agent used to crash when trying to query a pressure gauge channel that was off. This error is detailed in Issue #100 

## How Has This Been Tested?
I have been testing this on the software development computer at Penn. It has been running on this computer over the weekend, with no crashes, despite turning the pressure gauge channels on and off.

## Types of changes

## Checklist:


agents/pfeiffer_tpg366/pfeiffer_tpg366_agent.py
```diff
@@ -74,10 +93,15 @@ def read_pressure_all(self):
...
-        #gauge_states = np.array(gauge_states, dtype=int)
+        gauge_states = pressure_str.split(',')[::2]
+        gauge_states = np.array(gauge_states, dtype=int)
         pressures = pressure_str.split(',')[1::2]
         pressures = [float(p) for p in pressures]
+        if any(state != 0 for state in gauge_states):
+            index = np.where(gauge_states != 0)
+            for j in range(len(index[0])):
+                channel = np.int(index[0][j])
+                pressures[channel] = 0.
```
BrianJKoopman, line 104: A few things here:
- What's the expected pressure value when a sensor is off that you're setting to `0.` now?
- Looking at the manual for comms with the Pfeiffer device, there are more states than just `0` for OK, and non-zero for "Sensor switched off". State 5 seems to indicate no gauge and outputs `2.0E-2` for pressure. Are other states of interest? And do they return values other than `2.0E-2`? They could provide more information about a sensor that right now would make it appear to just be "off", when it could mean under or over range, error, no gauge, or ID error.
- Rather than doing `for j in range(len(index[0])):` you should use the [enumerate()](https://docs.python.org/3/library/functions.html#enumerate) built-in function. But even then, you don't really need the index here, and could get away with:
```
for j in index[0]:
    pressures[j] = 0.
```
- `np.int` [is deprecated](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated), use `int`, though I believe that `np.where` should give you only integers already, since it's the indices where `gauge_states != 0`, so I excluded the casting in the above example.

agents/pfeiffer_tpg366/pfeiffer_tpg366_agent.py
```diff
@@ -136,6 +159,7 @@ def start_acq(self, session, params=None):
                     'block_name': 'pressures',
                     'data': {}
                 }
+                self.gauge.channel_power()
```
BrianJKoopman, line 162: Currently this will generate a "Something is off!" log message at 2.5 Hz if conditions are met. I would suggest perhaps caching channel states in the `Pfeiffer` object, and logging state changes only when they change, rather than on every iteration of the loop.

agents/pfeiffer_tpg366/pfeiffer_tpg366_agent.py
```diff
@@ -34,6 +34,26 @@ def __init__(self, ip_address, port, timeout=10,
...
+    def channel_power(self):
+        '''
+        Function to check the power status of a given channel
+
+        Args:
+            None
+        Returns:
+            print statement warning about channels being off
+        '''
+        msg = 'SEN\r\n'
```
BrianJKoopman, line 46: Question about this command. In [the manual](https://www.pfeiffer-vacuum.com/filepool/file/measurement/bg5511ben.pdf?referer=1830&detailPdoId=32066&request_locale=en_US) it looks like state 0 means "gauge cannot be switched on/off" 1 is "gauge is off" and 2 is "gauge is on". Should we be checking for state 1 for seeing if gauges are off? Or am I misunderstanding the command docs?

BrianJKoopman (changes requested):
Thanks for the PR Tanay, it'll be great to get that error fixed.

While this currently sets the pressure value for any non-OK state to 0, a valid reading that I expect would solve the block mismatch issue, there are a few questions that I have about the return values from the two commands you're using, and some other minor comments to address.

agents/pfeiffer_tpg366/pfeiffer_tpg366_agent.py
tanaybhandarkar: Will the `txaio` module accomplish this?

BrianJKoopman: Not by itself, it just replaces the print statement. You'd have to still cache the data and check if the state has changed.

agents/pfeiffer_tpg366/pfeiffer_tpg366_agent.py
BrianJKoopman: To answer a few of my questions:
1. When sensors are off they read zero. Here's example output from a 'PRX' command with our DR's TPG366, p1 is off, the rest are on:
```
b'4,+0.0000E+00,0,+7.8800E+00,0,+7.0160E+01,0,+1.7824E+01,0,+2.2000E+01,0,+1.9500E+01\r\n'
```
2. Other states probably are useful, but are currently not checked. Will implement separately so we can consider this resolved for now. The rest of the changes in this comment were implemented.

PR #146, opened 2021-03-02
Remove check_port test

jlashner:
Removes the check-port test from the smurf-recorder

## Description

Removes the check-port test from the smurf-recorder

## Motivation and Context
Recently noticed that the check port option isn't consistently representing whether a G3Streamer is running on that port. I just noticed that it was failing this check, and not connecting to the G3Reader. I believe the difference from when I tested before is that we switched to connecting to the port on the smurf-server directly instead of using a tunnel, which may have changed the behavior of the check_port function.  Best to just take this check out and deal with the log_fatal messages before it causes more problems.

## How Has This Been Tested?
Tested on k2so

## Types of changes

## Checklist:

### DIFF
--- a/socs/agent/smurf_recorder.py
+++ b/socs/agent/smurf_recorder.py
@@ -22,34 +22,6 @@ class FlowControl(Enum):
     CLEANSE = 3
 
 
-def check_port(addr):
-    """
-    This function checks if socket port is currently open on a host. This can
-    be used to check if the smurf-streamer has created it's G3NetworkSender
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
This was initially added just to reduce the amount of "G3Reader could not connect" messages, yeah? I'm alright with it being removed, but does this mean we're back to the noisy logs? Strange that switching from tunneling to using the address directly would affect this.

PR #149, opened 2021-03-23
Created pfeiffer turbo controller agent

JakeSpisak:
Created the pfeiffer turbo controller agent, which controls and monitors the pfeiffer TC400 turbos used to run the Bluefors DRs. Also added documentation.

## Description
The base code is the pfeiffer vaccuum protocal, a set of python routines created by pfeiffer that can send requests and receive information from the turbos. From this a driver and a separate agent.py file were created. 

## Motivation and Context
We've wanted to remotely control and monitor the turbos for a while.

## How Has This Been Tested?
We ran all the tasks and processes: started and stopped both SAT-1 turbos, got error codes, and monitored data. NOTE: The monitoring data procedure potentially has a bug, because we've lost connection twice in about a month. We are not sure if this is a problem with the serial-to-ethernet converter or the agent, but decided to push rather than wait because the tasks work fine and we wanted to start the review process now.

## Types of changes

## Checklist:


JakeSpisak:
For context, this was the error in the most recent crash `
  | 2021-03-05T01:24:37+0000 monitor_turbo:1 Crash in thread: [Failure instance: Traceback: <class 'serial.serialutil.SerialException'>: read failed: [Errno 110] Connection timed out
-- | --
`

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Agent.py
```diff
@@ -0,0 +1,206 @@
...
+
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                self.turbo.ready_turbo()
+                time.sleep(1)
+                self.turbo.turn_turbo_motor_on()
+            else:
+                return False, "Could not acquire lock"
+
+        return True, 'Turned Turbo Motor On.'
```
BrianJKoopman, line 125: I make some suggestions in the driver file near these methods. But it looks to me like if these fail for whatever reason, maybe a current error on the turbo preventing it from being readied and turned on, this would still report that the task succeeded in turning on the turbo motor. At minimum I think we should be checking the returned values from these calls to control the turbo.

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Driver.py
```diff
@@ -0,0 +1,193 @@
...
+            
+        returns:
+            (bool) turbo_response: True for successful, False for failure.
+        """
+        
+        PVP._send_control_command(self.session, self.turbo_address, 10, "111111")
+        
+        addr, rw, param_num, turbo_response = PVP._read_gauge_response(self.session)
+        
+        return turbo_response
```
BrianJKoopman, line 137: I think this, and some of the other methods defined to command the turbo, could benefit from checking the returned values from `_read_gauge_response`, similar to what's done in the `Pfeiffer_Vacuum_Protocol.py` file. What happens if the turbo can't be readied? Should we raise an error? What happens in the scenario the turbo can't be readied, but then we call `turn_turbo_motor_on()`?

What brought me to this suggestion was considering the `turn_turbo_on` Task in the Agent, which boils down to:
```
self.turbo.ready_turbo()
time.sleep(1)
self.turbo.turn_turbo_motor_on()
```

What happens if line 1 fails and returns False? We would move right along to trying to turn on the turbo, and then report it turned on in either case. Raising an error here if the turbo couldn't be readied, or couldn't be turned on (in the case of `turn_turbo_motor_on`) would allow us to check for that error and react accordingly in the Agent.

(Structurally I also wonder if one would ever call `turn_turbo_motor_on` without first calling `ready_turbo`, in which case I would personally probably make `turn_turbo_motor_on` call `ready_turbo` and `time.sleep(1)` internally, rather than requiring one do so externally. In either case, checking we were successful in turning on the turbo is important, when that's what we expect the method to do, either here or in the Agent.)

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Agent.py
```diff
@@ -0,0 +1,206 @@
...
+            if acquired:
+                self.turbo.turn_turbo_motor_off()
+                time.sleep(1)
+                self.turbo.unready_turbo()
+            else:
+                return False, "Could not acquire lock"
+
+        return True, 'Turned Turbo Motor Off.'
+                          
+    def get_turbo_error_code(self, session, params=None):
```
BrianJKoopman, line 140: Having this in a separate Task is going to mean that during normal operations, the `monitor_turbo` Process will be continuously running, an error may then occur, say the turbo overheats. The monitor process will continue to run, reporting temperature and rotation rate, but the particular error code won't be retrieved until someone goes and clicks a button to fetch the error code, or a client has to handle getting an error code, rather than the error being identified right when it happened.

I don't know the interaction with the hardware, but what happens if you check for an error code when there isn't an error? And how fast can you send queries? Would it make sense to check for errors within the monitor process and either continuously publish error codes (assuming there's one for "no error") or publish when an actual error arrises?

BrianJKoopman (changes requested):
This is awesome to see, remote Turbo control is going to be very helpful at the site, and likely at every test institution. I do have some suggestions, mostly made inline here (sorry for the long review.) To summarize those inline comments:
* Whether or not the turbo successfully turned on/off should be checked when we try to do so in the Agent.
* Consider checking error codes within the long running monitor process, rather than having a separate Task that one needs to run to fetch an error code.
* Import the pfeiffer-vacuum-protocol package, rather than making a copy of it here.
* PEP08 and documentation suggestions.

General comments:
* There are also a bunch of "blank line contains whitespace" warnings if you run a linter that would be great to get rid of.
* We need a docs page for this Agent, please make one based on any of the other Agents in `docs/agents/`.

agents/pfeiffer_turbo_controller/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# SOCS Turbo Controller Agent
```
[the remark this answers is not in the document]
BrianJKoopman: I just saw in the PR text that this is for a TC400. My intention with asking this was to suggest we rename this to something like `pfeiffer_tc400`, like we have the `pfeiffer_tpg366` Agent, that follows more of the naming scheme we've been picking up.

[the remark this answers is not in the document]
JakeSpisak: renamed to pfeiffer_tc400

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Driver.py
```diff
@@ -0,0 +1,193 @@
...
+        PVP._send_control_command(self.session, self.turbo_address, 23, "000000")
+        
+        addr, rw, param_num, turbo_response = PVP._read_gauge_response(self.session)
+        
+        return turbo_response
+    
+    
+    def acknowledge_turbo_errors(self):
+        """Acknowledges the turbo errors. This is analagous to clearing the errors.
+            If the errors were fixed, the turbo will turn back on.
```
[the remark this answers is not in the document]
JakeSpisak: It means 'able to be turned on', I clarified

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Agent.py
JakeSpisak: I put in two checks, one for 'ready', one for 'on'

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Driver.py
JakeSpisak: I put return statements if either 'ready' or 'on' fail to return with errors.

agents/pfeiffer_turbo_controller/Pfeiffer_Turbo_Controller_Agent.py
JakeSpisak: I put the turbo error in the main acq process, it shouldn't take too much time

JakeSpisak:
I implemented all changes except importing the pfeiffer_vaccum_protocol module from a requirements file: I did that, then realized it wouldn't work because we need some functions that start with underscores (and I guess were just intended for internal use) because the regular functions don't allow temperature monitoring. And unfortunately the __init__.py file has "from pfeiffer_vaccum_protocol import *", which explicitly excludes underscore functions.

BrianJKoopman:
> I implemented all changes except importing the pfeiffer_vaccum_protocol module from a requirements file: I did that, then realized it wouldn't work because we need some functions that start with underscores (and I guess were just intended for internal use) because the regular functions don't allow temperature monitoring. And unfortunately the **init**.py file has "from pfeiffer_vaccum_protocol import *", which explicitly excludes underscore functions.

Thanks for the changes Jake. They look good. I made some minor formatting fixes in #177, if you don't mind taking a look at that before we merge.

I need to think about the import problem a bit more.

BrianJKoopman (changes requested):
Alright, so I've thought a bit more about the importing of the pfeiffer-vacuum-protocol module, and had some valuable discussion with @mhasself.

There's potentially some licensing issues with just copying that code into this repository that I don't fully understand, so I think we should just work around them.

Matthew pointed out that the private functions can be accessed if you specify the full submodule path. For instance:
```
from pfeiffer_vacuum_protocol.pfeiffer_vacuum_protocol import _send_data_request as pfeiffer_send_data_request
```
This would allow us to install the module from PyPI still, and use the code here with the update that `pvp._send_data_request`'s get replaced (and similarly with other functions.)

However, it's also a bit dangerous to depend on the private functions of a module like this, as they could be changed without notice. I think it's worth discussing either exposing these private functions or adding functions that would get the required information from the Turbo pumps upstream to the pfeiffer-vacuum-protocol module.

JakeSpisak:
I don't fully understand the proposed solution: if directly copying the functions over is bad due to licensing issues, but importing them is also bad, what do you mean by 'expose' or 'add functions'?

BrianJKoopman:
> I don't fully understand the proposed solution: if directly copying the functions over is bad due to licensing issues, but importing them is also bad, what do you mean by 'expose' or 'add functions'?

Sorry, I meant to say that I think we should import the private method with something like the line in my last comment for now. I just wanted to make you aware of the implication that private methods aren't guaranteed not to change.

What I mean by saying we should discuss if those functions could be exposed or to add functions, is that we should get in touch with the author of the `pfeiffer_vacuum_protocol` module and ask those things. Exposing being to remove the leading underscore on the functions (and thus imply that that's an interface that shouldn't change.)

If you want to implement the import above, I'm happy to reach out upstream to see what can be done.

BrianJKoopman (approved):

PR #151, opened 2021-03-26
CHWP Agent Feedback

BrianJKoopman:
Here are some small suggestions to latest chwp agent PR (https://github.com/simonsobs/socs/pull/80). To maybe save a back and fourth on changes I just made updates as I re-reviewed the Agent. A quick summary:
* Found some small typos in the Sphinx docs, fixed those and added the new page to the index so it'll appear in the sidebar.
* Protected the ocs import for the automated documentation build (else that system fails to build the docs and publish to the web)
* Caught two tabs in the agent file, replaced them with spaces
* Added "dumb-init" an init system we started using recently within the Docker containers. Also removed "sisock-" from the default commands, as that's been removed from the OCS system.
* Added the [recommended logging system](https://ocs.readthedocs.io/en/develop/developer/agents.html#logging), and updated most print statements to use it (I didn't touch the print in `pretty_print_irig_info`, just out of concern of messing up the formatting). This'll making distinguishing errors a bit nicer in the [logging system](https://ocs.readthedocs.io/en/develop/user/logging.html) that'll be setup at the site (and that you might have in the lab.)

### DIFF
--- a/agents/chwp/Dockerfile
+++ b/agents/chwp/Dockerfile
@@ -1,5 +1,4 @@
-# comment here...
-# 
+# CHWP Agent
 
 # Use socs base image
 FROM socs:latest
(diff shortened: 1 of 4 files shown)


harukinishino:
These look good to me. Thank you for those changes! Because there are no newer local changes in the original branch, I think there is no problem with merging.

harukinishino (approved):

PR #153, opened 2021-03-30
Add XY Stage Agent for LATRt Stages

kmharrington:
## Description
Adds an agent for interacting with the XY Stage controller we're using with the LATRt. 

## Motivation and Context
We want position readout in `.g3` files. Which means we also need to control the device through OCS.  

## How Has This Been Tested?
It's been tested for the use-cases we need for beam mapping. The lower level controller has the safety features.  

## Types of changes

## Checklist:

### DIFF
--- a/agents/xy_stage/xy_latrt_agent.py
+++ b/agents/xy_stage/xy_latrt_agent.py
@@ -12,7 +12,7 @@ if not ON_RTD:
     ## yes I shouldn't have named that module agent
     from xy_agent.xy_connect import XY_Stage
 
-class LATRtXYAgent:
+class LATRtXYStageAgent:
     """
     Agent for connecting to the LATRt XY Stages
     
@@ -232,12 +232,12 @@ if __name__ == '__main__':
     parser = make_parser()
 
     # Interpret options in the context of site_config.
-    args = site_config.parse_args(agent_class = 'LATRtXYAgent', parser=parser)
+    args = site_config.parse_args(agent_class = 'LATRtXYStageAgent', parser=parser)
     
 
     agent, runner = ocs_agent.init_site_agent(args)
 
-    xy_agent = LATRtXYAgent(agent, args.ip_address, args.port, args.mode, args.sampling_frequency)
+    xy_agent = LATRtXYStageAgent(agent, args.ip_address, args.port, args.mode, args.sampling_frequency)
 
     agent.register_task('init_xy_stage', xy_agent.init_xy_stage_task)
     agent.register_task('move_x_cm', xy_agent.move_x_cm)
--- a/agents/xy_stage/xy_latrt_agent.py
+++ b/agents/xy_stage/xy_latrt_agent.py
@@ -229,16 +229,12 @@ if __name__ == '__main__':
     # Start logging
     txaio.start_logging(level=os.environ.get("LOGLEVEL", "info"))
 
-    # Get the default ocs argument parser.
-    site_parser = site_config.add_arguments()
+    parser = make_parser()
 
-    parser = make_parser(site_parser)
-
-    # Parse comand line.
-    args = parser.parse_args()
-    
     # Interpret options in the context of site_config.
-    site_config.reparse_args(args, 'XY_StageAgent')
+    args = site_config.parse_args(agent_class = 'LATRtXYAgent', parser=parser)
+    
+
     agent, runner = ocs_agent.init_site_agent(args)
 
     xy_agent = LATRtXYAgent(agent, args.ip_address, args.port, args.mode, args.sampling_frequency)
--- a/agents/xy_stage/xy_latrt_agent.py
+++ b/agents/xy_stage/xy_latrt_agent.py
@@ -3,24 +3,24 @@ import argparse
 import time
 import txaio
 
-import numpy as np
-
-## yes I shouldn't have named that module agent
-from xy_agent.xy_connect import XY_Stage
 
 ON_RTD = os.environ.get('READTHEDOCS') == 'True'
 if not ON_RTD:
     from ocs import ocs_agent, site_config
     from ocs.ocs_twisted import TimeoutLock, Pacemaker
 
-class XY_Agent:
+    ## yes I shouldn't have named that module agent
+    from xy_agent.xy_connect import XY_Stage
+
+class LATRt_XY_Agent:
     """
     Agent for connecting to the LATRt XY Stages
-    Args: name
-          ip_addr -- IP address where RPi server is running
-          port    -- Port the RPi Server is listening on
-          mode    -- 'acq': Start data acquisition on initialize
-          samp    -- default sampling frequency in Hz
+    
+    Args: 
+        ip_addr: IP address where RPi server is running
+        port: Port the RPi Server is listening on
+        mode: 'acq': Start data acquisition on initialize
+        samp: default sampling frequency in Hz
     """
 
     def __init__(self, agent, ip_addr, port, mode=None, samp=2):
@@ -51,15 +51,15 @@ class XY_Agent:
         self.agent.register_feed('positions',
                                  record = True,
                                  agg_params = agg_params,
-                                 buffer_time = 1)
+                                 buffer_time = 0)
     
     def init_xy_stage_task(self, session, params=None):
         """init_xy_stage_task(params=None)
         Perform first time setup for communivation with XY stages.
+
         Args:
             params (dict): Parameters dictionary for passing parameters to
                 task.
-        Parameters:
         """
 
         if params is None:
@@ -73,13 +73,11 @@ class XY_Agent:
                 return False, "Could not acquire lock."
             # Run the function you want to run
             self.log.debug("Lock Acquired Connecting to Stages")
-            try:
-                self.xy_stage = XY_Stage(self.ip_addr, self.port)
-                self.xy_stage.init_stages()
-        
-                print("XY Stages Initialized")
-            except ValueError:
-                    pass
+            
+            self.xy_stage = XY_Stage(self.ip_addr, self.port)
+            self.xy_stage.init_stages()
+            print("XY Stages Initialized")
+            
         # This part is for the record and to allow future calls to proceed,
         # so does not require the lock
         self.initialized = True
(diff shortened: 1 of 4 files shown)


agents/xy_stage/xy_ocs_agent.py
```diff
@@ -0,0 +1,243 @@
...
+                self.log.warn("Could not start acq because {} is already running".format(self.lock.job))
+                return False, "Could not acquire lock."
+
+            self.log.info(f"Starting Data Acquisition for XY Stages at {f_sample} Hz")
+            session.set_status('running')
+            self.take_data = True
+            last_release = time.time()
+
+            while self.take_data:
+                if time.time()-last_release > 1.:
```
mhasself, line 172: I believe you need to update `last_release` inside this block...

kmharrington: oh truth. shall fix

kmharrington: Thanks for the tip!

mhasself: Great.  Last thing (before Brian takes over :) ) -- that sleep(random.uniform(...)) in the Pacemaker docs is actually supposed to be simulating your instrument's random sampling time... you don't need it in your agent (you have a real device to provide the variance!).  Maybe @jlashner wants to clarify the docs on that :)

kmharrington: Alright, gotcha. Thanks!

agents/xy_stage/xy_ocs_agent.py
```diff
@@ -0,0 +1,249 @@
...
+                pm.sleep()
+
+                data = {'timestamp':time.time(), 'block_name':'positions','data':{}}
+                pos = self.xy_stage.position
+                self.is_moving = self.xy_stage.moving
+
+                data['data']['x'] = pos[0]
+                data['data']['y'] = pos[1] 
+                self.agent.publish_to_feed('positions',data)
+                self.agent.feeds['positions'].flush_buffer()
```
jlashner, line 191: Instead of flushing the ocs feed every sample, you can set `buffer_time=0` when registering the feed instead of `buffer_time=1`.  That will make it so data is sent as soon as the publish statement is called.

agents/xy_stage/xy_ocs_agent.py
```diff
@@ -0,0 +1,249 @@
...
+                        return False, "could not re-acquire lock"
+                    last_release = time.time()
+                pm.sleep()
+
+                data = {'timestamp':time.time(), 'block_name':'positions','data':{}}
+                pos = self.xy_stage.position
+                self.is_moving = self.xy_stage.moving
+
+                data['data']['x'] = pos[0]
+                data['data']['y'] = pos[1]
```
jlashner, line 189: Having `'is_moving` in the data block might be nice for a grafana dashboard, or might be helpful for analysis later on.

agents/xy_stage/xy_ocs_agent.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        if params is None:
+            params = {}
+
+        
+        f_sample = params.get('sampling_frequency', self.sampling_frequency)
+        pm = Pacemaker(f_sample, quantize=True)
+
+        if not self.initialized:
+            self.init_xy_stage_task(session)
```
jlashner, line 164: I don't think this is explicitly forbidden... I just have never thought of calling an operation from another operation by just passing in the session object. @mhasself is this allowed? It might mess up the "status" flow of the session, which I think has to be ordered: starting < running < stopping < done. Since the `init_xy_stage_task` doesn't set the I don't think this will break, but I'm pretty sure it's not recommended.

jlashner: I think a more official way to do this would be to call the `self.agent`'s  [start](https://github.com/simonsobs/ocs/blob/develop/ocs/ocs_agent.py#L542) and [wait](https://github.com/simonsobs/ocs/blob/ff0c23995c3682b1a66969647c46304963759afb/ocs/ocs_agent.py#L603)  functions with the operation name

jlashner (changes requested):
Looks great! I have a few small comments, but I think the one thing that should probably be changed is how you call the `init_xy_stage_task` from the `acq` process.

agents/xy_stage/xy_ocs_agent.py
```diff
@@ -0,0 +1,249 @@
+import os
+import argparse
+import time
+import txaio
+
+import numpy as np
+
+## yes I shouldn't have named that module agent
+from xy_agent.xy_connect import XY_Stage
```
jlashner, line 9: Also, where is this driver located? Any chance we can put it in the `socs/agents` directory?

jlashner:
One more comment, so seems like k2so and sat are using a slightly different xy stage, so it'll also need it's  own agent. Should we rename this so that it's unique to the latrt instead of a generic xy_stage?

BrianJKoopman (changes requested):
This is look good. One thing that should be added is a docs file in `socs/docs/agents/`. I usually work from one of the other agent's files. Please include an example ocs site config file block for configuring this Agent, and a description for installing any dependencies (if xy_agent doesn't make it into socs.)

If the plan is to build a Docker container for this, we should do that and include an example docker-compose block in the docs as well.

agents/xy_stage/xy_ocs_agent.py
kmharrington: possibly but adding it now is messing up the displays of the other data in grafana?

agents/xy_stage/xy_ocs_agent.py
kmharrington: Something is messing up there but I'll just raise an error if things are initialized.

agents/xy_stage/xy_ocs_agent.py
kmharrington: Drivers are here: https://github.com/kmharrington/xy_stage_control 

I don't want to add the client side of the drivers to socs because I need to be able to use it outside ocs as well. And the would require maintaining two sets of code that do the exact same thing.

kmharrington:
I think the recent commit fix all the comments?

agents/xy_stage/xy_ocs_agent.py
jlashner: It shouldn't effect other displays unless your querying all fields with a wildcard or something

agents/xy_stage/xy_latrt_agent.py
```diff
@@ -0,0 +1,253 @@
...
+
+ON_RTD = os.environ.get('READTHEDOCS') == 'True'
+if not ON_RTD:
+    from ocs import ocs_agent, site_config
+    from ocs.ocs_twisted import TimeoutLock, Pacemaker
+
+    ## yes I shouldn't have named that module agent
+    from xy_agent.xy_connect import XY_Stage
+
+class LATRt_XY_Agent:
```
BrianJKoopman, line 15: Sorry for not noticing this in the first review, but class name should be formatting in CapWords, i.e. `LatrtXyAgent`, without the underscores. This will keep the naming consistent with other Agent class names.

BrianJKoopman (changes requested):
Sorry, one other comment to do with naming consistency among Agents that I missed in the first review but noticed after the renaming of the Agent class.

I also think if we're renaming to avoid future name conflicts, then the `xy_stage/` directory name should be changed while we're at it.

agents/xy_stage/xy_latrt_agent.py
```diff
@@ -0,0 +1,253 @@
...
+    # Get the default ocs argument parser.
+    site_parser = site_config.add_arguments()
+
+    parser = make_parser(site_parser)
+
+    # Parse comand line.
+    args = parser.parse_args()
+    
+    # Interpret options in the context of site_config.
+    site_config.reparse_args(args, 'XY_StageAgent')
```
BrianJKoopman, line 241: `reparse_args` is deprecated. This block should instead read:
```
    parser = make_parser()

    # Interpret options in the context of site_config.
    site_config.parse_args(agent_class='XY_StageAgent', parser=parser)
```

kmharrington: Do you have an example of where this is used? Reparse_args is still in the docs: https://ocs.readthedocs.io/en/latest/developer/agents.html 

Just replacing that block as-is means I don't have the args variable for the rest of that code.

BrianJKoopman: Yeah, better docs are here: https://ocs.readthedocs.io/en/develop/developer/site_config.html#agent-site-related-command-line-parameters

(Working on getting that other page updated: https://github.com/simonsobs/ocs/pull/199)

> Just replacing that block as-is means I don't have the args variable for the rest of that code.

Sorry, I missed an `args = ` before `site_config.parse_args()`.

kmharrington: Found on in the smurf recorder

kmharrington:
Alright, I think that fixes the other comments?

docs/agents/latrt_xy_stage.rst
```diff
@@ -0,0 +1,72 @@
...
+Below are configuration examples for the ocs config file and for running the
+Agent in a docker container.
+
+ocs-config
+``````````
+To configure the LATRt XY Stage Agent we need to add a block to our ocs 
+configuration file. Here is an example configuration block using all of 
+the available arguments::
+
+      {'agent-class': 'LATRt_XY_StageAgent',
```
BrianJKoopman, line 34: Needs underscores removed, and this needs to match the `agent_class` passed to `site_config.parse_args()`, so it should be `'LATRtXYAgent'`.

docs/agents/latrt_xy_stage.rst
```diff
@@ -0,0 +1,72 @@
...
+    xy_agent.move_y_cm.wait()
+    
+    #Get instantaneous position
+    status, message, session = xy_stage.acq.status()
+    print(session['data']['data'])
+    
+Agent API
+---------
+
+.. autoclass:: agents.xy_stage.xy_latrt_agent.LATRt_XY_Agent
```
BrianJKoopman, line 71: This also needs underscores removed in the class name.

BrianJKoopman (changes requested):
Great, thanks for the changes. There are two additional spots that need updating with the class name underscores removed. Then I think we're good to merge.

jlashner (approved):
Looks good to me! Thanks Katie

docs/agents/latrt_xy_stage.rst
kmharrington: Right, sorry to miss the docs! Should be fixed now

BrianJKoopman (approved):
Great, I push one final fix to make the agent class name consistent with what you're using in your ocs-site-configs definitions. These need to match so that ocs can identify which configuration block to use for your Agent. I also renamed the class to keep consistent, even though this isn't technically required it is what that string is meant to reference.

Going to merge once checks pass one last time.

PR #156, opened 2021-04-23
first commit of acu agent to socs

laurensaunders:
## Types of changes

## Checklist:


BrianJKoopman (changes requested):
This is looking good, thanks for cleaning up the file structure! There are still some files I don't think we need here that I've commented about inline.

I also read through the Agent and helper code and have made some small suggestions and requests for documentation.

Some more general comments/questions:
- Are safe defaults/ranges handled by the ACU software at a lower level, or should those be caught here? For instance, a lot of the params for tasks/processes have no default, is `None` handled properly in all cases? Similarly, what about if I say something like `go_3rd_axis(10000)` (by way of the set_boresight task)?
- print statements could be `self.log.info()` statements or equivalent (`.error()`, `.warn()`, `.debug()`, etc.) This'll help distinguish errors when integrated with the logging system.
- It'd be great to run a code linter like flake8 and go through some of the suggestions, particularly the whitespace and indentation suggestions.

Looking forward to getting this into socs! Thanks again for putting the PR together.

BrianJKoopman (changes requested):
Thanks for all the work on the suggested changes! There's a couple of small comments related to the changes. We also still need a documentation page to compile the docstrings you've added here into a page for the Agent. Details are in this comment: https://github.com/simonsobs/socs/pull/156#discussion_r622360846

Other docs pages in [socs/docs/agents/](https://github.com/simonsobs/socs/tree/develop/docs/agents) are a great reference for what sort of information to include and for syntax. I'm also happy to talk about making the docs page anytime.

BrianJKoopman (commented):
Thanks for adding the docs! There were a couple of syntax things stopping them from building that I fixed while trying to test the build. I committed fixes and PR'd them in #179. Please take a look and let me know if you have any questions. If these look alright to you, we should merge that PR, then I'll approve and merge this one.

BrianJKoopman (approved):

PR #160, opened 2021-05-03
Add FTS Agent

kmharrington:
Add an Agent for controlling the FTS.

## Description
The agent can be used to control the two UChicago FTSs. One will be used with the LATRt and the other will be used at K2SO. They will probably also make it to the site at some point.

## How Has This Been Tested?
This has been tested with one of the FTSs for a complete measurement (moving the FTS through a range of positions while taking data). 

## Types of changes

## Checklist:


agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+            self.stage.home()
+
+        return True, "Homing Complete"
+    
+    def move_to(self, session, params):
+        """Move to absolute position relative to stage center (in mm)
+
+        params: {'position':float between -74.8 and 74.8}
+        """
+
```
BrianJKoopman, line 191: You should handle the case that `params` didn't get passed (whether intentionally or not).

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+        params: {'position':float between -74.8 and 74.8}
+        """
+    
+        with self.lock.acquire_timeout(timeout=3, job='move') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start move because lock held by" \
+                                "{self.lock.job}")
+                return False, "Could not get lock"
+            self.stage.move_to( params.get('position') )
```
BrianJKoopman, line 197: Is there a sensible default position? If `params` gets passed but without the `'position'` key, this'll pass `None` to `self.stage.move_to()`, which I don't think handles that case.

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+
+        return True, "Move Complete"
```
BrianJKoopman, line 199: Maybe return the result of `self.stage.move_to()` here? It looks like if I pass an out of bounds position, `self.stag.move_to()` will return `False, "Move out of bounds!"`, but the task here then returns `True, "Move Complete"`. And I think the result would be the stage wouldn't have moved, but the task would report moving.

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+    def home_task(self, session, params=None):
+        """ Home the stage to its negative limit
+        """
+
+        with self.lock.acquire_timeout(timeout=3, job='home') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start home because lock held by" \
+                                "{self.lock.job}")
+                return False, "Could not get lock"
+            self.stage.home()
```
BrianJKoopman, line 182: What's the response read after sending the `HOME X` command? Should it checked here to make sure it was successful?

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+        self.log.debug("Trying to acquire lock")
+        with self.lock.acquire_timeout(timeout=0, job='init') as acquired:
+            # Locking mechanism stops code from proceeding if no lock acquired
+            if not acquired:
+                self.log.warn("Could not start init because {} is already" \
+                                "running".format(self.lock.job))
+                return False, "Could not acquire lock."
+            # Run the function you want to run
+            self.log.debug("Lock Acquired Connecting to Stages")
+            self.stage = FTSStage(self.ip_addr, self.port)
```
BrianJKoopman, line 164: I think this would run into the same scenario that you encountered in #164. A fix was implemented in #168 that you could adapt here to handle this case.

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+                    last_release = time.time()
+                pm.sleep()
+
+                data = {
+                    'timestamp':time.time(),
+                    'block_name':'position',
+                    'data':{}}
+                success, pos = self.stage.get_position()
+                
+                data['data']['pos'] = pos
```
BrianJKoopman, line 243: If `get_position()` isn't successful this is going to put a `None` into the data array, which will cause a `TypeError` when published. We should prevent that from happening. Perhaps check the success state before publishing?

docs/agents/fts_agent.rst
```diff
@@ -0,0 +1,64 @@
+.. highlight:: rst
+
+.. _fts_stage:
+
+=====================
+FTS Agent
+=====================
+
+This agent is used to communicate with the FTS mirror stage for two FTSs with
+Aerotech motion controllers.
```
BrianJKoopman, line 10: I know there are other FTS in the collaboration. Is there a good way to distinguish them? Is it by motion controller? Inevitably we'll probably run into naming choices with the Agents corresponding to this and the other FTS. Is there a better, more specific name we could choose?

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+    
+    def get_position(self):
+        self.send('CMDPOS X\n')
+        time.sleep(0.1)
+        out = self.read()
+        try:
+            M, B = self.TRANSLATE
+            self.pos = (float(out[1:])-B)/M
+            return True, self.pos
+        except:
```
BrianJKoopman, line 73: What's this bare `except` catching?

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+        return True, 'Stage Initialized.'
+
+    def home_task(self, session, params=None):
+        """ Home the stage to its negative limit
+        """
+
+        with self.lock.acquire_timeout(timeout=3, job='home') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start home because lock held by" \
+                                "{self.lock.job}")
```
BrianJKoopman, line 180: The f-string gets broken by that line break. Move the `f` down to the second string.

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+    def move_to(self, session, params):
+        """Move to absolute position relative to stage center (in mm)
+
+        params: {'position':float between -74.8 and 74.8}
+        """
+    
+        with self.lock.acquire_timeout(timeout=3, job='move') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start move because lock held by" \
+                                "{self.lock.job}")
```
BrianJKoopman, line 195: Same f-string problem as above.

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+            raise Exception("Connection to Stages not initialized")
+
+        with self.lock.acquire_timeout(timeout=0, job='acq') as acquired:
+            if not acquired:
+                self.log.warn("Could not start acq because {} is already" \
+                            "running".format(self.lock.job))
+                return False, "Could not acquire lock."
+
+            self.log.info(f"Starting Data Acquisition for FTS Mirror at" \
+                            "{f_sample} Hz")
```
BrianJKoopman, line 223: Same f-string problem as above.

agents/fts_stage/fts_agent.py
```diff
@@ -0,0 +1,304 @@
...
+            session.set_status('running')
+            self.take_data = True
+            last_release = time.time()
+
+            while self.take_data:
+                if time.time()-last_release > 1.:
+                    if not self.lock.release_and_acquire(timeout=20):
+                        self.log.warn(f"Could not re-acquire lock now held by" \
+                                      "{self.lock.job}.")
```
BrianJKoopman, line 232: Same f-string problem as above.

BrianJKoopman (changes requested):
Thanks for the PR, Katie! I think there are a few edge cases we need to preemptively handle before this is ready to merge. I'm also curious to hear your thoughts on naming when considering there will be other FTS Agents for the other FTS that I think is at UCSD.

agents/fts_stage/fts_agent.py
kmharrington: Honestly no idea, that part of the code came from ACT's "agent." Looking at it I'm guessing it's protecting from getting a non-sense response from the controller.

agents/fts_stage/fts_agent.py
BrianJKoopman: Yup, that's it. The following return is still relevant too, and completes the task/process (by saying that it failed.)

docs/agents/fts_agent.rst
kmharrington: That is a good point. We're sending our second FTS to UCSD and it should have the exact same controller as this one. So it should work the same but if there's small differences I'll help debug to get them both working on the same agent. As far as I know these two are the only two that could end up at the site? 

The other FTS I know about is the one being used at K2SO. I don't know if @jseibert575  has plans to make an OCS agent for that one?

agents/fts_stage/fts_agent.py
kmharrington: It returns nothing useful. The only thing is the read call blocks until the home is done. And since the controller ignores everything while it's running I have that read call there mostly just to block.

kmharrington:
Ok! I think I've got everything updated off your review!

BrianJKoopman (approved):
Thanks Katie, looks good! One quick question on the naming, is Aerotech the brand of the stage? Before I merge I'm going to rebase and handle some conflicts that arose, but am also thinking I'll rename the directory this is in from `fts_stage` to `fts_aerotech` or similar. Any better suggestions? (Just wondering if `fts_stage` is still general enough to potentially conflict with another FTS Agent.)

kmharrington:
Yeah Aerotech is the company name for the controller. I left the folder as-is kind of thinking we could have multiple agents in there for different FTSs. But that's totally up to you.

BrianJKoopman:
> Yeah Aerotech is the company name for the controller. I left the folder as-is kind of thinking we could have multiple agents in there for different FTSs. But that's totally up to you.

I'd like to keep them separate, especially if they end up being added to the Docker builds. I moved to `fts_aerotech_stage/` and updated the needed paths in files. Also caught up to the latest develop and handled those few merge conflicts. Will merge when checks finish.

PR #163, opened 2021-05-06
Added set_still_heater() and get_still_heater() to LS372 agent

zhuber21:
## Description

Added the functions set_still_heater() and get_still_heater() to LS372_agent.py so that LS372 clients can successfully set the still heater still power (as opposed to the still heater manual power...a different thing) and can then check that it was set correctly.

set_still_heater() takes one argument (output), which is the percentage of the still heater power and must be between 0 and 100. 
get_still_heater() takes no arguments.

## Motivation and Context

Solves the problem of set_heater_power() being the only way to set the still power, which isn't good because it sets the still heater manual power instead of the still heater still power. Also provides a method (get_still_output) for displaying the current still heater still power using session.data via get_still_output.status().

## How Has This Been Tested?

I tested this change by writing a short client script that set the still heater on and off to a set still heater still power and confirming that the expected changes were made on our Lakeshore 372. Since I added two new functions without changing any old ones (other than the docstrings), I do not expect that this will change any other areas of code.

## Types of changes

## Checklist:

Not sure if it follows the code style, but I based it upon existing methods so I hope so.
The [documentation for the Agent API](https://socs.readthedocs.io/en/latest/agents/lakeshore372.html#agent-api), which I think is what this falls under, is not very updated. It would be great to update that more so that people trying to figure out how to use the agent via a client know which tasks are registered and what their arguments are. I would be happy to help with that if possible.

### DIFF
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -635,7 +635,8 @@ class LS372_Agent:
                                     
     def set_still_output(self, session, params=None):
         """
-        Set the still output on the still heater. This is different than the manual output on the still heater. Use set_heater_output() for that.
+        Set the still output on the still heater. This is different than the manual output
+        on the still heater. Use set_heater_output() for that.
 
         :param params: dict with "output" parameter
         :type params: dict
@@ -671,8 +672,12 @@ class LS372_Agent:
         """
         Gets the current still output on the still heater.
 
-        :param params: dict
-        :type params: dict
+        This task has no useful parameters.
+
+        The still heater output is stored in the session.data
+        object in the format::
+
+          {"still_heater_still_out": 9.628}
 
         """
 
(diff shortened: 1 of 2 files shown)


BrianJKoopman (changes requested):
This is great, always good to see added functionality to Agents.

To add the docstrings from your new tasks to the sphinx page you could add them to the members list in https://github.com/simonsobs/socs/blob/master/docs/agents/lakeshore372.rst. Or better yet, remove the list entirely leaving:
```
.. autoclass:: agents.lakeshore372.LS372_agent.LS372_Agent
    :members:
```
This will put all method docstrings into the page, including for your new tasks. The formatting for a lot of these could use some work, but it'll be good to have the info displayed anyway.

zhuber21:
I've made the changes you requested, @BrianJKoopman, including to the documentation file. Take a look and see if it looks good to you.

PR #168, opened 2021-05-20
Handle LS372 ConnectionError if device powered off at Agent startup

### LINKED ISSUE #164: LS372 Agent Misses Power On
Discovered a fun edge case. If you start the LS372 agent while the box is powered off, it doesn't throw any errors or exit. But, it doesn't connect when you power it on. 

This means that you can't just come back and run `docker-compose up` to restart it, you have to manually going in a stop that container and restart. I think like a better default behavior would be for the agent to just exit.

BrianJKoopman:
## Description

This introduces some error handling for connection errors to the 372 on Agent startup. When establishing the connection in the init task we watch for an OSError, make sure it relates to a ConnectionError and raise that for handling in the Agent.

Within the Agent we catch the ConnectionError and shutdown the reactor, exiting the Agent. 

## Motivation and Context

fixes #164 

## How Has This Been Tested?

This has been tested against a powered off LS372 at Yale.

## Types of changes

## Checklist:

### DIFF
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -166,7 +166,11 @@ class LS372_Agent:
                 except ConnectionError:
                     self.log.error("Could not connect to the LS372. Exiting.")
                     reactor.callFromThread(reactor.stop)
-                    return False, 'Lakeshore initialized failed'
+                    return False, 'Lakeshore initialization failed'
+                except Exception as e:
+                    self.log.error(f"Unhandled exception encountered: {e}")
+                    reactor.callFromThread(reactor.stop)
+                    return False, 'Lakeshore initialization failed'
 
                 print("Initialized Lakeshore module: {!s}".format(self.module))
                 session.add_message("Lakeshore initilized with ID: %s"%self.module.id)
--- a/agents/lakeshore372/LS372_agent.py
+++ b/agents/lakeshore372/LS372_agent.py
@@ -6,6 +6,7 @@ import numpy as np
 import txaio
 import threading
 from contextlib import contextmanager
+from twisted.internet import reactor
 
 from socs.Lakeshore.Lakeshore372 import LS372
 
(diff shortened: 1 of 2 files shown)


agents/lakeshore372/LS372_agent.py
```diff
@@ -160,7 +161,13 @@ def init_lakeshore_task(self, session, params=None):
                 session.add_message("No initialization since faking data")
                 self.thermometers = ["thermA", "thermB"]
             else:
-                self.module = LS372(self.ip)
+                try:
+                    self.module = LS372(self.ip)
+                except ConnectionError:
```
jlashner, line 166: Why are we only crashing the agent on ConnectionError? Are there other safer errors that are ever raised by `_establish_socket_connection` that leave the agent in a usable state?

BrianJKoopman: Not that I know of, but this is what would raise if the network cable were unplugged or the 372 powered off, which are the two scenarios I'm trying to account for. It's better, I think, to be specific with your errors rather than try to catch everything with a general except. That way if other errors exist they can be seen and fixed accordingly.

jlashner: I agree with trying to be specific about errors, but in this case it means that the default behavior is to leave the agent in a state where the module failed to initialize but you can still try to start tasks like `acq`, which I'm not really sure is correct. Shouldn't be a problem if clients are checking if their tasks succeeded / failed, but right now I don't think most people are doing that... I'll leave it to your judgment though, and everything else looks good to me

jlashner (approved):

agents/lakeshore372/LS372_agent.py
BrianJKoopman: Oh, you mean in the case that another error comes up, yeah, I see your point. What do you think of a more general except after this `ConnectionError` handling?
```
except Exception as e:
    self.log.error(f"Unhandled except encountered: {e}")
    reactor.callFromThread(reactor.stop)
    return False, 'Lakeshore initialization failed'
```

jlashner: Yeah something like that looks good to me

BrianJKoopman:
Alright, added that section `except`, just rebased on develop, merging now. Thanks for the comments!

PR #170, opened 2021-05-20
Add network_mode to docker-compose example in SCPI PSU Agent docs

BrianJKoopman:
## Description

@zhuber21 found a missing config option in the SCPI PSU Agent docs. Copying them as is will likely result in connection issues. The Agent should be on the host network to properly see the PSU's IP.

### DIFF
--- a/docs/agents/scpi_psu.rst
+++ b/docs/agents/scpi_psu.rst
@@ -49,6 +49,7 @@ An example docker-compose service configuration is shown here::
   ocs-psuK:
     image: simonsobs/ocs-scpi-psu-agent:latest
     hostname: ocs-docker
+    network_mode: "host"
     volumes:
       - ${OCS_CONFIG_DIR}:/config:ro
     command:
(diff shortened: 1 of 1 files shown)


docs/agents/scpi_psu.rst
```diff
@@ -49,6 +49,7 @@ An example docker-compose service configuration is shown here::
   ocs-psuK:
     image: simonsobs/ocs-scpi-psu-agent:latest
     hostname: ocs-docker
+    network_mode: "host"
```
zhuber21, line 52: I just had 
network_mode: host
and that seemed to work fine

zhuber21 (approved):
I didn't need the quotes around host, but otherwise this matches the fix that we made. I don't know if it matters or not. It looks like we do have the quotations around host for the LS372 agent, so it probably is best to be consistent and keep them.

BrianJKoopman:
Yeah, I noticed the quotes on the 372 docs, so kept consistent. It's also consistent with the [docker-compose reference](https://docs.docker.com/compose/compose-file/compose-file-v3/#network_mode).

PR #171, opened 2021-05-24
bump sodetlib and remove ocs install

jlashner:
Updates sodetlib version to most recent, and removes ocs install since that is now done in the sodetlib docker.

## How Has This Been Tested?

Tested with ocs-pysmurf-controller and emulator on smurf-srv-so3.

### DIFF
--- a/agents/pysmurf_controller/Dockerfile
+++ b/agents/pysmurf_controller/Dockerfile
@@ -1,23 +1,9 @@
-FROM simonsobs/sodetlib:v0.1.0
-
-# Set locale
-ENV LANG C.UTF-8
-
-# Installs Jupyter requirements
-RUN pip3 install jupyter
+FROM simonsobs/sodetlib:v0.2.0-23-g90fd9f2
 
 WORKDIR /app
 
-# Installs OCS requirements
-#COPY requirements.txt .
-#RUN pip3 install -r requirements.txt
-
-# OCS installation
-RUN git clone https://github.com/simonsobs/ocs.git \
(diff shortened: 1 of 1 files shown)


agents/pysmurf_controller/Dockerfile
```diff
@@ -1,23 +1,9 @@
...
-#COPY requirements.txt .
-#RUN pip3 install -r requirements.txt
-
-# OCS installation
-RUN git clone https://github.com/simonsobs/ocs.git \
-    && pip3 install -r ocs/requirements.txt \
-    && pip3 install -e ocs
-
 # SOCS installation
+RUN pip3 install cryptography==3.3.2
```
BrianJKoopman, line 6: I'm wondering about this line. I think it related to fixing the builds when the images were still on Ubuntu 18.04? Do we still need it now? (It's also happening in the sodetlib build, so perhaps not really needed here anyway...)

jlashner: We need to pin cryptography because the pysmurf dockers are still running on 18.04 and I don't really see that changing any time soon. You're right that it's already built into the sodetlib repo, so we can probably remove this line here.

jlashner: Tested and it builds/runs fine without that line, so I think it's redundant

BrianJKoopman: Ah, right, I missed that this was based on 18.04 still. Removing since it's in the sodetlib image still sounds good.

BrianJKoopman (approved):

PR #173, opened 2021-05-27
Add ability to read out custom registers from labjack agent

msilvafe:
## Description

Added a new acq process called `acq_reg` that doesn't use the `ljstream` module but instead uses the `eReadNames` module which allows you to read non-standard registers from the labjack.

## Motivation and Context

The labjack has some configurable built in calibrations for reading out thermocouples see link [here](https://labjack.com/support/software/applications/t-series/kipling/register-matrix/configuring-reading-thermocouple) These registers however cannot be streamed using the `ljstream` module. So we needed to modify the agent to be able to read these out. This was motivated by the need to use a thermocouple instead of the standard warm thermometers because they go to a higher temperature range that can be used with our FTS and beammapper thermal source.

## How Has This Been Tested?

Ran it on adaq1 at UCSD and saw that the feeds were reading into grafana.

## Types of changes

## Checklist:


agents/labjack/labjack_agent.py
```diff
@@ -342,6 +349,104 @@ def stop_acq(self, session, params=None):
...
+            ljm.close(self.handle)
+            self.log.info("Data stream stopped")
+        return True, 'Acquisition exited cleanly.'
+
+    def stop_acq_reg(self, session, params=None):
+        if self.take_data:
+            self.take_data = False
+            return True, 'requested to stop taking data.'
+        else:
+            return False, 'acq is not currently running'
```
BrianJKoopman, line 449: This is exactly duplicating `stop_acq`, since `self.take_data` is used to track when either is running. I wonder if we can just use `stop_acq` in both cases. That would involve removing this, and updating the `register_process` lines on 506 and 507 to use `sensors.stop_acq` instead.

agents/labjack/labjack_agent.py
```diff
@@ -342,6 +349,104 @@ def stop_acq(self, session, params=None):
...
+                          'if you want to sample faster please use'
+                          'standard acquire mode and do not read out '
+                          'any non-standard channels. Setting sampling'
+                          f'rate to maximum = {max_fsamp} Hz')
+            scan_rate_input = max_fsamp
+            scan_rate_dt = 0
+        else:
+            scan_rate_dt = (1/scan_rate_input) - read_dt
+            
+        scans_per_read = max(1, int(scan_rate_input))
```
BrianJKoopman, line 399: Assigned but never used.

agents/labjack/labjack_agent.py
```diff
@@ -342,6 +349,104 @@ def stop_acq(self, session, params=None):
...
+        #and recording the time it takes to read each time. Then
+        #setting the max sample rate to be 50mS greater than the median
+        #of the time it took to read.
+        num_chs = len(self.chs)
+        times = []
+        for i in range(100):
+            times.append(time.time())
+            ljm.eReadNames(self.handle,num_chs,self.chs)[0]
+        read_dt = np.round(np.median(np.diff(times)),2)+0.05
+        max_fsamp = 1/read_dt
```
BrianJKoopman, line 380: This at most is ~2.5 Hz?

agents/labjack/labjack_agent.py
```diff
@@ -206,7 +206,11 @@ def __init__(self, agent, ip_address, active_channels, function_file,
                                  record=True,
                                  agg_params=agg_params_downsampled,
                                  buffer_time=1)
-
+        self.agent.register_feed('registers',
+                                 record = True,
+                                 agg_params = {'frame_length':10*60},
+                                 buffer_time = 0.)
```
BrianJKoopman, line 212: All other feeds in the Agent use a 1 second buffer. Any reason not to here?

BrianJKoopman (changes requested):
Thanks for the PR, Max. A couple of small comments/questions inline below. On top of that, can you add some description to the documentation [here](https://github.com/simonsobs/socs/blob/develop/docs/agents/labjack.rst) about using non-standard registers. Mostly I'm interested in when you might want to use them vs the normal channels, and how a user should adjust their configuration accordingly, i.e. what "active-channels" do they now use. (The why seems covered by the docstring in `start_acq_reg`, so mostly I want how we know what registers to readout and the naming scheme there.)

The new Process isn't being picked up by the autoclass in the docs. In line 131 of t[he docs page](https://github.com/simonsobs/socs/blob/develop/docs/agents/labjack.rst) you could remove the list of members and it will pick up all functions in the Agent with docstrings, or add the new `start_acq_reg` to the list.

```
Agent API
---------

.. autoclass:: agents.labjack.labjack_agent.LabJackAgent
    :members:
```

agents/labjack/labjack_agent.py
```diff
@@ -342,6 +349,104 @@ def stop_acq(self, session, params=None):
...
+                          'standard acquire mode and do not read out '
+                          'any non-standard channels. Setting sampling'
+                          f'rate to maximum = {max_fsamp} Hz')
+            scan_rate_input = max_fsamp
+            scan_rate_dt = 0
+        else:
+            scan_rate_dt = (1/scan_rate_input) - read_dt
+            
+        scans_per_read = max(1, int(scan_rate_input))
+        with self.lock.acquire_timeout(0, job='acq') as acquired:
```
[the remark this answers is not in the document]
msilvafe: Changed this.

agents/labjack/labjack_agent.py
```diff
@@ -378,6 +483,8 @@ def make_parser(parser=None):
     init_params = False
     if args.mode == 'acq':
         init_params = {'auto_acquire': True}
+    if args.mode == 'acq_reg':
```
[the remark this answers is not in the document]
msilvafe: Fixed this.

agents/labjack/labjack_agent.py
msilvafe: Switched to just stop_acq

agents/labjack/labjack_agent.py
msilvafe: Removed

agents/labjack/labjack_agent.py
msilvafe: Made it more clear that it is hard fixed at an upper bound of 2.5 Hz by adding in a line `max_fsamp = min(2.5,1/read_dt)`

agents/labjack/labjack_agent.py
msilvafe: No, changed to 1 second.

BrianJKoopman:
Thanks for the updates, @msilvafe! Can you merge the latest 'develop' branch into this one and resolve the two conflicts that have arisen?

BrianJKoopman (approved):
Thanks for the updates, Max!

Some of the changes to the docstrings broke the autodoc formatting, so I fixed those up. I'll merge shortly, once checks re-run.

PR #174, opened 2021-06-02
Bluefors inode handling

BrianJKoopman:
## Description

Update the Bluefors Agent to handle a change in inode of logfiles.

## Motivation and Context

The Bluefors Agent was originally built to run on Docker for Windows. This was because at the time I ran into issues running on Linux and syncing the logs continuously when it came to keeping track of the active log files. This was dependent on the configuration a bit. The Bluefors software writes to simple .csv files on Windows, and I was using a tool called Syncthing to copy them as they were updated to a Linux box. The Linux box was running the Agent, which would keep files open and continously try to `readline()`. On copy the inode changed, and Python lost track of the file. It worked if you followed the files directly on Windows though, so we ended up with the Docker on Windows solution people currently run.

Docker on Windows has been difficult to install at some institutions. This patch adds monitoring of the file inode to the Agent. This should allow the Agent to be run on Linux, and users can then sync the files regularly from Windows for reading (in any of the various ways that you could, assuming you do so at roughly the rate they're written.) If the inode changes (which it will if you have a setup similar to the above), then the Agent will recognize that and reopen the file to grab the last line. There are some protections in place to make sure we're looking at recent data and not something that's stale.

## How Has This Been Tested?

It's been tested with a Bluefors log simulator that I wrote, both normally (writing and reading directly to the same files) and with an rsync in between -- where I'd write the logs, and rsync them to the location the Agent is monitoring (changing the inode). It works in both situations. I will have Cornell test a development image a bit as well.

## Types of changes

## Checklist:

### DIFF
--- a/agents/bluefors/Dockerfile
+++ b/agents/bluefors/Dockerfile
@@ -14,4 +14,6 @@ COPY . .
 ENTRYPOINT ["dumb-init", "python3", "-u", "bluefors_log_tracker.py"]
 
 # Sensible default instance-id
-CMD ["--instance-id=bluefors"]
+CMD ["--instance-id=bluefors", \
+     "--site-hub=ws://crossbar:8001/ws", \
+     "--site-http=http://crossbar:8001/call"]
--- a/agents/bluefors/bluefors_log_tracker.py
+++ b/agents/bluefors/bluefors_log_tracker.py
@@ -399,7 +399,7 @@ class LogParser:
                 LOG.debug("Data: {d}", d=data)
                 # If the file was reopened due to an inode change we don't know
                 # if the last line is recent enough to be worth publishing. Check
-                if (time.time() - data['timestamp']) < stale_time*60:
+                if (time.time() - data['timestamp']) < int(stale_time)*60:
                     app_session.app.publish_to_feed('bluefors', data)
                 else:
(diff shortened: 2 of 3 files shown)


BrianJKoopman:
In testing with @zhuber21 we found that the modifications I made weren't sufficient for the setup at Cornell. The past batch of commits adds a check on file modification times, and introduces another environment variable to set `MODE`. A combination of setting the mode and file modification times will trigger a reopening of the log file like the inode changes I described initially. The Cornell setup would keep inode static, but the Agent would still lose track of file changes, I think because of the use of sshfs to mount the log directory. Things are working now, and documentation is updated to reflect the recent changes.

zhuber21 (approved):
Looks good to me. We are still seeing some minor bugs with the agent dropping files occasionally when our samba drive resets in the morning, but those can be fixed by restarting the agent. Otherwise the agent has been doing well logging on both our systems, including throughout a cooldown on one of those systems.

BrianJKoopman:
> Looks good to me. We are still seeing some minor bugs with the agent dropping files occasionally when our samba drive resets in the morning, but those can be fixed by restarting the agent. Otherwise the agent has been doing well logging on both our systems, including throughout a cooldown on one of those systems.

Great, thanks Zach. Those last bugs are in #175, will hopefully get to them soon.

PR #176, opened 2021-06-11
Record LS372 Sample Heater

BrianJKoopman:
## Description

Add an option to enable logging the 372's sample heater, typically used to servo a DR. The addition comes in the acq process. Since you might not even be using the sample heater on a device though, we don't want to always query it. Adding `[--sample-heater, True]` to the Agent's argument list will enable.

## Motivation and Context

This value is helpful to understand what your servo is doing.

## How Has This Been Tested?

In use at Yale.

## Types of changes

## Checklist:


BrianJKoopman:
Merging after some discussion on the developers call yesterday.

PR #178, opened 2021-06-23
Magpie agent

jlashner:
This PR adds the Magpie ocs agent.
 
## Description
The magpie agent listens to a smurf data stream, runs some basic processing (lowpass filter and downsample), and repackages the data so lyrebird can understand it. Making this an Agent allows us to control it and set certain parameters like target rate, filter params, etc. without restarting lyrebird. This agent should also be able to do basic demodulation, which has not yet been implemented.  This is named Magpie in the same vein as "Lyrebird" and "Kookaburra" (from polarbear) after birds that like to mimic other animals, just like this agent will be observing and mimicking the smurf datastream.

This interface still needs a lot of work, including:
- docs
- better lyrebird config file generator
- Testing demodulation processing
- Visualization of multiple slots

## Motivation and Context
This agent is required to use lyrebird with for SO data stream.

## How Has This Been Tested?
This has been tested and run at both UCSD and U-Chicago to visualize single slots.

## Types of changes

## Checklist:


jlashner:
Added some docs, so I think this is ready for review since there are people at various institutions who are waiting for some instructions before delving into lyrebird usage

BrianJKoopman (commented):
This is looking good, thanks Jack. Mostly some documentation requests related to documenting the Agent tasks/processes so they show up in Sphinx. Class docstrings could also be added for the two supporting classes.

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,388 @@
...
+        self.scale = 1
+        self.offsets = np.zeros(self.fp.num_dets)
+        self.zero_flag = True
+
+        self.demodulate=True
+        self.demod_filter = FIRFilter.butter_lowpass(1, 200., order=4)
+        self.demod_freq = 4
+
+    def set_scale(self, session, params={}):
+        self.scale = params['scale']
```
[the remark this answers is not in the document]
jlashner: Personally I'd rather the task fail loudly when a required param is missing instead of choosing a default, so I normally just try to access the key and it'll raise KeyError if there's a problem. Do you think I should be handling this differently?

jlashner:
Added lots of docstrings and addressed other comments

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,388 @@
...
+                )
+            else:
+                self.fp = FocalplaneConfig.from_wafer_file(
+                    args.wafer_file, wafer_scale=args.wafer_scale
+                )
+
+        self.sent_cfg = False
+        self.num_dets = self.fp.num_dets
+        self.mask = np.arange(self.num_dets)
+        self.mask = np.arange(4960)
```
mhasself, line 207: This looks like it's supposed to be 4096 ... and like it trashes the line before so not both are needed.

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,388 @@
...
+
+        def det_coords(det):
+            quat =np.array(det['quat']).astype(np.float64)
+            rdir = qa.rotate(quat, zaxis).flatten()
+            ang = np.arctan2(rdir[1], rdir[0])
+            orient = qa.rotate(quat, xaxis).flatten()
+            polang = np.arctan2(orient[1], orient[0])
+            mag = np.arccos(rdir[2]) * 180 / np.pi
+            xpos = mag * np.cos(ang)
+            ypos = mag * np.sin(ang)
```
mhasself, line 157: Consider just xpos, ypos = rdir[:2] * 180/np.pi.  Using arccos to recover small angles (mag) has a lot of numerical error, since you're looking for the delta in cos(delta) ~ 1 - delta^2/2 + ... .

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,388 @@
...
+            self.demodulate = params['state']
+        if 'freq' in params:
+            self.demodulate = params['freq']
+        if ('lowpass_order' in params) or ('lowpass_cutoff'):
+            order = params.get('lowpass_order', 4)
+            cutoff = params.get('lowpass_cutoff', 1)
+            self.demod_filter = FIRFilter.butter_lowpass(cutoff, 200., order=order)
+        return True, f"set demod {params}"
+
+    def process_status(self, frame):
```
mhasself, line 287: Since "process" now conflicts with OCS "Process" language, it's even more useful for the docstrings to point out which things in here are Task and Process functions... and which are just frame handling helpers.

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+def make_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+
+    parser.add_argument('--src', default='tcp://localhost:4532',
+                        help="Address of incoming G3Frames.")
+    parser.add_argument('--dest', type=int, default=8675,
+                        help="Port to server lyrebird frames")
```
mhasself, line 388: typo

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,388 @@
...
+    parser.add_argument('--target-rate', '-t', type=float, default=5)
+    parser.add_argument('--layout', '-l', default='grid', choices=['grid', 'wafer'],
+                        help="Focal plane layout style")
+    parser.add_argument('--xdim', type=int, default=64,
+                        help="Number of pixesl in x-dimension for grid layout")
+    parser.add_argument('--ydim', type=int, default=64,
+                        help="Number of pixesl in y-dimension for grid layout")
```
mhasself, line 395: Is the plural of pixesl pixessl?

docs/agents/magpie.rst
```diff
@@ -0,0 +1,121 @@
...
+The magpie is an incredibly intelligent bird, with a decent ability to mimic
+other bird calls, though not as good as `the superb lyrebird
+<https://www.youtube.com/watch?v=mSB71jNq-yQ>`_. In the context of OCS, the job
+of the Magpie agent is to parse data being passed from the SMuRF Streamer, and
+to translate it into a stream that Lyrebird can interpret and display.
+
+The Magpie agent should be setup and run on the same node that is running
+lyrebird, which must have some sort of graphics card (either integrated or
+dedicated), and must be on the same network as the smurf-server for real-time
+playback.
```
mhasself, line 22: Why must the agent be on the node running lyrebird?  I suspect one can tunnel away this requirement... is there a reason someone should not do that?

I think it would be helpful to have a paragraph describing how you co-operate this with Lyrebird ... launch the agent, then start the main process, then launch lyrebird with default parameter?  Or do you need some particular lyrebird config file?

docs/agents/magpie.rst
```diff
@@ -0,0 +1,121 @@
...
+
+Detector Layouts
+-----------------
+
+Grid Layout
+````````````
+The grid layout is a grid with 4096 elements (maximum number of channels a
+single smurf slot can stream)
+
+.. image:: ../_static/images/lyrebird_grid.jpg
```
mhasself, line 37: Seems like we should be able to get better images here... are the graphics not screen-grabbable for some reason?

mhasself (changes requested):
Cool to see this.  Lots of minor comments inline.  I second Brian's concerns about docstrings for the OCS interface.  I am also keen to see a little more in the documentation that would tell me how to operate this along with Lyrebird.

One feature request (but not a showstopper for now): I think it might be helpful to have a "test_stream" mode, that can push out fake data to Lyrebird in the absence of a true Smurf stream.  It's always helpful to be able to debug connection issues on a leg-by-leg basis.  This could also be used to visualize a layout that one is modifying.

agents/magpie/magpie_agent.py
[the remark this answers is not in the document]
BrianJKoopman: Yeah, I still think it'd be good to handle the `KeyError`. If the proper param isn't provided you could then `return False, 'scale' is a required parameter` or similar to mark the task as failed.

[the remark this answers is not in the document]
mhasself: Would it be helpful if Agent code could raise some special OCS errors to propagate such failures outwards?  That would go along with some functions to help unpack parameters.  E.g. what if you could do:
```
    self.scale = get_required(params, 'scale')
```
and that would raise ocs.ParamMissingError('scale'), which would be caught by the OCSAgent wrapper and returned to the user as (False, "Parameter 'scale' is required.")

I think this approach would be in line with encouraging concise Agent code without too much boilerplate that still can propagate errors adequately.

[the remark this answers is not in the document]
BrianJKoopman: That sounds helpful to me. I've been pointing out the need to handle missing params in recent PRs, typically encouraging what I have here, using `.get()` with a sensible default value. It would definitely help in cases like this where one wants to force a parameter being set.

BrianJKoopman (changes requested):
The new param decorator should be added to the Tasks and Processes here. That'll satisfy the lingering comments I had.

Additionally, the docs should be updated to follow [new guidelines](https://ocs.readthedocs.io/en/develop/developer/agents.html#documentation).

jlashner:
Alright, finaly ready to re-open this PR for review.  I tried to go back and address comments that were made before this update, but I also added and changed a fair bit so I think it's just worth re-reviewing. In particular here are a few things that changed:

 - I'm now calculating the rms of each channel which is sent to lyrebird each frame
 - I separated out the "sending" operation into its own process, with an async queue used to store G3Frames which are to be sent. The `send` process tries to send frames at about the correct cadence so that lyrebird updates smoothly.
 - G3Files are accepted as a data-source, so you can easily stream data from files by specifying the file paths instead of the tcp socket address.
 - Added the requirement of a unique stream-id to avoid conflicts between multiple magpie instances sending data to a single lyrebird client. This honestly could be the same as the `sostream-id` in the G3Stream, but the data-value names are generated before any G3Frames come in so I just added this as an argument.
 - There's a new process for streaming fake data. This can be activated by passing the `--fake-data` argument, which causes the fake-data process to begin at startup instead of the `read` process.
 - More docstrings and docs.

mhasself (approved):
I've not gone exhaustively through this, but it looks to be in a good state.  A few minor comments inline.

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,705 @@
...
+            f = self.out_queue.get(block=True)
+            t = f['timestamp'].time / core.G3Units.s
+            now = time.time()
+            if first_frame_time is None:
+                first_frame_time = t
+                stream_start_time = now
+
+            this_frame_time = stream_start_time + (t - first_frame_time) + self.delay
+            if this_frame_time > now:
+                time.sleep(this_frame_time - now)
```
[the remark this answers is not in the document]
jlashner: Is the `session` object the same in the process start / stop function? I think maybe instead of having like:
```
while self._send_running:
    time.sleep(...)
``` 
we should do have something like:
```
while session.status == 'running':
    ocs_agent.sleep_until_stopped(10000, session)
    ...
```
and then process stop functions should just set `session.status = 'stopping'`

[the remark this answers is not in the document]
mhasself: And more generally, a "default" stop function that moves status to 'stopping' also makes sense.

I think we'd discussed also having status advance to "running" automatically before calling the start function, so people don't have to call it themselves.  I can't find the discussion on that anywhere, but it was the plan IIRC.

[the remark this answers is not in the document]
mhasself: > This seems to come up in a bunch of different contexts (especially for dealing with smurf stuff and streaming data). Is there any way to come up with a generic way to do this? Like a function that sleeps until either a certain duration has passed or the current operation is stopped?

You'd need to "break" manually ... main loops would look like:
```
  while True:
    if not ocs_agent.sleep_while_running(10000, session):
      break
    ...
```
but ya, I think that's useful.  Get it working here and we can migrate it into ocs.

jlashner:
Alright sorry that took so long, but I think I addressed the remaining comments and added some more docs about how lyrebird works internally, specifically about "data values", "equations", and "colormaps", which is what Magpie configures. Should I go ahead and merge this or do either of you want to look it over again?

docs/agents/magpie.rst
```diff
@@ -122,9 +124,164 @@ To get lyrebird running, you must bring software up in the following order:
 
       run_lyrebird --port 8675 8676
 
-   and this will start lyrebird for the two corresponding slots.
+   and this will start lyrebird for the two corresponding slotkjs.
```
mhasself, line 127: Is this polish notation?

The --port arg still looks wrong, above.  Correct me...

docs/agents/magpie.rst
```diff
@@ -0,0 +1,383 @@
...
+         '--delay', 5,
+         '--target-rate', 20,
+
+         '--layout', 'wafer',
+         # Detmap CSV file
+         '--det-map', '/home/jlashner/lyrebird_demo/detmap.csv',
+         '--offset', 0, 0,
+       ]},
+
+Docker
```
BrianJKoopman, line 51: For consistency with other Agent reference pages, please name this section "Docker Compose".

docs/agents/magpie.rst
```diff
@@ -0,0 +1,383 @@
...
+         '--delay', 5,
+         '--target-rate', 20
+         '--layout', 'wafer',
+
+         # Detmap CSV file
+         '--csv-file', '/home/jlashner/lyrebird_demo/detmap.csv',
+         '--offset', 0, 0,
+         '--fake-data',
+       ]},
+
```
BrianJKoopman, line 362: I would consider moving these under the "Configuration File Examples/OCS Site Config" section. (And possibly just describe the differences in the options, though I can understand wanting a whole block to copy.)

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,743 @@
...
+        """
+        if 'session_id' not in frame:
+            return
+        if self.mask_register in frame['status']:
+            status = yaml.safe_load(frame['status'])
+            self.mask = np.array(
+                ast.literal_eval(status[self.mask_register])
+            )
+
+    def process_data(self, frame):
```
BrianJKoopman, line 484: Since this is internal to the Agent, and not a registered Task/Process, I think you should make this private. Same with `process_status`. That'll limit the Agent API to just tasks/processes.

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,743 @@
...
+            target_rate : float
+                Target sample rate for lyrebird (Hz)
+        """
+        self.target_rate = params['target_rate']
+        return True, f'Set target rate to {self.target_rate}'
+
+
+    @ocs_agent.param('delay', type=float)
+    def set_delay(self, session, params):
+        """
```
BrianJKoopman, line 460: Needs a method signature override at the top of the docstring: `set_delay(delay)`

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,743 @@
...
+
+        self.mask = np.arange(MAX_CHANS)
+        self.out_queue = queue.Queue(1000)
+        self.delay = args.delay
+
+        self.avg1, self.avg2 = None, None
+
+
+    @ocs_agent.param('target_rate', type=float)
+    def set_target_rate(self, session, params):
```
BrianJKoopman, line 446: Needs a method signature override at the top of the docstring: `set_target_rate(target_rate)`

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,743 @@
...
+        self._running = True
+        session.set_status('running')
+
+        src_idx = 0 
+        if isinstance(params['src'], str):
+            sources = [params['src']]
+        else:
+            sources = params['src']
+
+        reader = core.G3Reader(sources[src_idx])
```
BrianJKoopman, line 571: What happens if this can't connect to the `tcp://` address?

docs/agents/magpie.rst
jlashner: dziękuję, ya the `--port <port1> <port2>` works for me, and is I think is the correct usage for argparse `nargs='+'`

agents/magpie/magpie_agent.py
jlashner: Currently the agent will just crash. I was planning on eventually making this more robust by adding the logic from the smurf-recorder agent, but wasn't originally planning on doing it in this PR.... I think it does make sense to add it for robustness so I'll try it out.

jlashner:
Alright, went through and addressed remaining comments and changed some of the main process thread to better handle G3Reader timeouts and dropped connections.

docs/agents/magpie.rst
```diff
@@ -0,0 +1,343 @@
...
+Below is an example docker-compose entry for running a magpie corresponding to
+crate-id 1 and slot 2. If crossbar is being run on a different server, you'll
+have to modify the ``site-hub`` and ``site-http`` args accordingly::
+
+    ocs-magpie-crate1slot2:
+        image: simonsobs/ocs-magpie:${SOCS_TAG}
+        hostname: ocs-docker
+        user: ocs:ocs
+        network_mode: host
+        container_name: ocs-magpie
```
BrianJKoopman, line 76: Is there a reason to name the container here? (i.e. something is relying on that name to resolve) If not I think this should be removed to avoid conflicts when running multiple instances for different crates/slots. (Or the name here made more specific, like `ocs-magpie-crate1slot2`, but letting docker-compose do the naming should be fine too.)

docs/agents/magpie.rst
```diff
@@ -0,0 +1,343 @@
...
+         '--det-map', '/home/jlashner/lyrebird_demo/detmap.csv',
+         '--offset', 0, 0,
+       ]},
+
+It is  possible to tell magpie to stream data from existing G3Files instead of
+directly from the smurf-stream. To do this, simply set the ``src`` argument to
+be the filepath of the file you wish to stream. If you want to stream from
+multiple files in series, you can do this by putting multiple source filepaths,
+as is shown below. In this case, once the first file has finished streaming
+magpie will imediately start streaming from the second file::
```
BrianJKoopman, line 56: 'imediately' -> 'immediately'

agents/magpie/magpie_agent.py
```diff
@@ -0,0 +1,770 @@
...
+            target_rate : float
+                Target sample rate for lyrebird (Hz)
+        """
+        self.target_rate = params['target_rate']
+        return True, f'Set target rate to {self.target_rate}'
+
+
+    @ocs_agent.param('delay', type=float)
+    def set_delay(self, session, params):
+        """set_delay(delay=5.0)
```
BrianJKoopman, line 461: Thanks for adding the method signature overrides. However, now they imply the default values you listed. I think either those need to be dropped, or the defaults need to be added to the param decorators.

Some of these look like they're defaults inherent to the Agent setup, like this delay for instance, defaults to 5 on agent startup. But, calling this from a client with `delay.start()` and assuming that the delay is going to default to 5.0 is going to cause an error, as the decorator is indicating delay as a required argument.

This applies to `set_delay`, `set_target_rate`, and `send`.

BrianJKoopman (changes requested):
Thanks for the changes, Jack! Biggest thing here now is the defaults in the method signatures/param decorators.

docs/agents/magpie.rst
jlashner: no good reason other than convenience. I'll remove

jlashner:
Alright, thanks brian! Lemme know if there's anything else.

BrianJKoopman (approved):
Looks good to me!

PR #179, opened 2021-06-30
ACU Agent Addition Patch

BrianJKoopman:
Some final comments/syntax fixes from the most recent round of commits in #156.

- When building the docs, I was running into some warnings about indentation in the `generate_scan()` docstring. Switching to `Args:` instead of `Params:` seemed to help. I also reduced the indentation (which wasn't fixing the problem when using `Params:`, but was the style recommendation from Google, which is the style we're using here.)
- Removed the `agents/acu/docs/` directory, as the file within it is now in `socs/docs/agents/`.
- The `argparse` autodoc section at the top of `acu_agent.rst` needs to be pointed to the function `add_agent_args`, else it can't build the arguments list and throws an error.
- rst links to external pages need to end in an `_`.
- I try to include all arguments for an Agent in the configuration file examples, so I added `--acu_config` with its default value to the example.
- `:members:` specified in the autoclass for the Agent documentation need to refer to the actual method names, not the OCS names associated with the process or task.
- Mocked imports within sphinx's `conf.py` to allow docs builds without soaculib installed.
- Added the ACU agent docs to the main index so it shows in the TOC.
- Merged in the latest develop branch, since a few config files got behind.
- Removed the check for a `'wait'` param in the `go_to()` method. (Related to https://github.com/simonsobs/socs/pull/156#discussion_r658003263)


laurensaunders (approved):
Looks good!

PR #188, opened 2021-08-10
Adds filter and downsample to labjack agent

jlashner:
This adds the ability to filter and downsample labjack data before writing to disk. 

## Description

This adds  a signal processing module with a FIR Filter (used by both this agent and the magpie agent), and a few extra parameters to the labjack agent to enable filtering and downsampling before data is written to disk.

## Motivation and Context

This is mostly to eliminate aliasing in buttkicker accelerometer tests we've been doing on the SAT and LAT. This will allow us to take data at a much higher sample rate, then apply an anti-aliasing filter in software and downsample so that we are writing data at a semi-reasonable cadence to disk. The data rates we're thinking about is raw sampling rates of about 5-10 kHz (for three channels) being filtered then downsampled to ~400 Hz.

## How Has This Been Tested?

Will test this on the SAT this week.

## Types of changes

## Checklist:

### DIFF
--- a/agents/labjack/labjack_agent.py
+++ b/agents/labjack/labjack_agent.py
@@ -8,6 +8,9 @@ import csv
 from scipy.interpolate import interp1d
 import numpy as np
 import txaio
+
+from socs.signal_processing import FIRFilter
+
 txaio.use_twisted()
 
 ON_RTD = os.environ.get('READTHEDOCS') == 'True'
@@ -154,7 +157,8 @@ class LabJackFunctions:
 # LabJack agent class
 class LabJackAgent:
     def __init__(self, agent, ip_address, active_channels, function_file,
-                 sampling_frequency):
+                 sampling_frequency, enable_filter=False, lowpass_order=5,
+                 lowpass_cutoff=200., downsample_factor=1):
         self.active = True
         self.agent = agent
         self.log = agent.log
@@ -163,6 +167,12 @@ class LabJackAgent:
         self.module = None
         self.ljf = LabJackFunctions()
         self.sampling_frequency = sampling_frequency
+
+        # Filtering params
+        self.enable_filter = enable_filter
+        self.lowpass_order = lowpass_order
+        self.lowpass_cutoff = lowpass_cutoff
+        self.downsample_factor = downsample_factor
 
         # Labjack channels to read
         if active_channels == 'T7-all':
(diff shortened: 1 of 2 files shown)


agents/labjack/labjack_agent.py
```diff
@@ -309,9 +347,10 @@ def start_acq(self, session, params=None):
 
                 # The labjack outputs at exactly the scan rate but doesn't
                 # generate timestamps. So create them here.
-                timestamps = [cur_time+i/scan_rate for i in range(scans_per_read)]
-                cur_time += scans_per_read/scan_rate
-                data['timestamps'] = timestamps
+                t1 = t0 + read_period
+                timestamps = np.linspace(t0, t1, scans_per_read)
+                data['timestamps'] = timestamps[downsample_mask]
+                t0 = t1
```
jlashner, line 353: I'm kind of skeptical of this method of generating timestamps, since there's nothing to protect against timestamp drifts if the sample rate isn't exactly what we expect. I didn't want to change the logic here in this PR, but maybe we should think about whether this is the correct way to calculate timestamps

mhasself: Agreed this is not good practice... unless the Labjack has access to a realtime clock, then it will drift relative to the agent's computer.  I think it should be fixed for real (unfortunately that's not trivial), but perhaps for now you could at least test for drift and write an error message.

jlashner:
Tested this on SAT and works well now. Here's a comparison of the accelerometer data with and without the aliasing filter, sampled at 8 kHz and downsampled to 400 Hz:
waterfall_unfiltered
waterfall_filtered
Most of the bouncing lines are aliasing artifacts, and adding the anti-aliasing filter both removes most of them and reduces the noise floor.

agents/labjack/labjack_agent.py
```diff
@@ -259,8 +283,32 @@ def start_acq(self, session, params=None):
...
+        downsample_factor = params.get('downsample_factor',
+                                       self.downsample_factor)
+        save_unfiltered = params.get('save_unfiltered', False)
+        chs = params.get('channels', self.chs)
+        if not set(chs).issubset(set(self.chs)):
+            raise ValueError(
+                f"Invalid channels specified! Must be in {self.chs}"
+            )
+        self.log.info(f"Starting taking data for channels {chs}")
+        block_name = '_'.join(sorted(chs))
```
jlashner, line 300: Question about this: I was having trouble implementing the ability to set the channels being read out through the stream command because the block structure wouldn't match if you stopped the stream, and restarted the stream with a different channel set without restarting the agent. This is the solution I came up with (to make sure the block-name is unique for each combination of channels you could chose. But I just wanted to make sure this wouldn't break anything in the data-loading. I.e.  a field "AIN0V" happens to appear in two different blocks in a single frame, would the hk data loader know how to handle that? Should I change this so every channel is published to a unique block?

mhasself: I think that might violate something in the specification (like that a field must not appear in more than one block name for an active provider).  The intended solution for cases like this is that you resign the active provider and start a new provider.  The natural way to do that is to restart the agent.

Alternatively, you could publish to a different feed name (rather than same feed but different block name)... that would trigger a new provider instance (IIRC) **but** then the downstream systems wouldn't naturally group the fields together.

In summary... I don't think this is supported naturally.  A solution might be to add a trigger in the aggregator so it could somehow be told to re-register a provider, so you can change the field contents in a block.

jlashner: Ok I think I'll probably just remove this arg. Was useful for testing but not sure how useful it'll be in reality

BrianJKoopman (commented):
I don't have a ton to add to this beyond the discussion already ongoing about timestamps and block names. Some small documentation comments for now.

jlashner:
Nope, I think its safe to close this.

PR #190, opened 2021-08-16
Control input for ls 372

jlashner:
This PR adds the ability to monitor the control input of the LS372 through OCS. This is what is currently being run on adaq1.

## Description
Adds support for reading the control channel from the LS372 along with some extra command line args and tasks to enable / disable it. Also switches the lakeshore to use site_config.parse_args instead of reparse_args.

## Motivation and Context
We need to PID off of control in k2so, so this agent has been in use for a while in a dev image. We'd like to switch it to an official image.

## How Has This Been Tested?

It is currently running on k2so.

## Types of changes

## Checklist:


jseibert575 (approved):

agents/lakeshore372/LS372_agent.py
```diff
@@ -314,17 +326,36 @@ def start_acq(self, session, params=None):
...
+                    if self.control_chan_enabled:
+                        temp = self.module.get_temp(unit='kelvin', chan=0)
+                        res = self.module.get_temp(unit='ohms', chan=0)
+                        cur_time = time.time()
+                        data = {
+                            'timestamp': time.time(),
+                            'block_name': 'control_chan',
+                            'data': {
+                                'control_T': temp,
+                                'control_R': res
```
BrianJKoopman, line 348: I debated on whether I'd prefer this name or "Input A", which the Lakeshore defaults to calling this channel (via `INNAME? A`.) We aren't using the channel name feature of the 372 at all, except for grabbing the default channel names, but if you wanted to grab names like done above it'd be:
```
channel_str = self.module.channels[0].name.replace(' ', '_')
```

agents/lakeshore372/LS372_agent.py
```diff
@@ -314,17 +326,36 @@ def start_acq(self, session, params=None):
...
+                            }
+                        }
+                        session.app.publish_to_feed('temperatures', data)
+                        self.log.debug("{data}", data=session.data)
+                        # Updates session data w/ control field
+                        session.data['fields'].update({
+                            'control': {
+                                'T': temp, 'R': res, 'timestamp': cur_time
```
BrianJKoopman, line 356: Please add an example line to the `session.data` example in the docstring with "control" as a key?

agents/lakeshore372/LS372_agent.py
```diff
@@ -753,10 +784,12 @@ def make_parser(parser=None):
                               second if it is set longer than a channel's dwell\
                               time. This ensures at least one second of data\
                               collection at the end of a scan.")
-    pgroup.add_argument('--auto-acquire', type=bool, default=True,
+    pgroup.add_argument('--auto-acquire', action='store_true',
```
BrianJKoopman, line 787: This is definitely needed, but we should also update the docs page ocs-site-config example block: `['--auto-acquire', True]` -> `['--auto-acquire'],`

agents/lakeshore372/LS372_agent.py
```diff
@@ -753,10 +784,12 @@ def make_parser(parser=None):
...
                         help='Automatically start data acquisition on startup')
     pgroup.add_argument('--sample-heater', type=bool, default=False,
                         help='Record sample heater output during acquisition.')
+    pgroup.add_argument('--enable-control-chan', action='store_true',
+                        help='Enable reading of the control input each acq cycle')
```
BrianJKoopman, line 792: This argument should also be added to the ocs-site-config example block in the docs.

BrianJKoopman (changes requested):
A good addition, didn't know anyone was actually using this input. Mostly have documentation change requests, and a small debate over naming of the field.

agents/lakeshore372/LS372_agent.py
jlashner: If it's up for debate, I personally think the names 'control_T', and 'control_R' are ok, and I'm not super keen to change it if we don't have to, only because we've been using these for a while. But I can switch if you think that would be best.

agents/lakeshore372/LS372_agent.py
jlashner: Do you know if anybody is currently using this arg? If so this will probably break their setup.

agents/lakeshore372/LS372_agent.py
jlashner: Are you sure? I don't think labs will typically using that argument, so I'm not sure if it should go into the example block.

jlashner: It should still show up in the argparse autodoc though.

agents/lakeshore372/LS372_agent.py
BrianJKoopman: Probably everyone. The way it's written right now, does passing `False` even work? Or does site_config handle the type conversion of that bool properly? Unmodified argparse will resolve that to `True` anyway.

If site_config handles it, then I'm tempted to say keep the current implementation so that we don't break configs.

agents/lakeshore372/LS372_agent.py
BrianJKoopman: > Are you sure? I don't think labs will typically using that argument, so I'm not sure if it should go into the example block.

I've been trying to put every available option in the example block with whatever the default value is. Options like this are, of course, not mandatory, but if someone wants to use it I want there to be a good example.

PR #221, opened 2021-12-09
Lakeshore425 agent. This is for one of the sensors in CHWP system.

ykyohei:
Developped new agent for sensor in CHWP system
## Description
New agent for Lakeshore Model 425 gaussmeter

## Motivation and Context
For monitoring CHWP system

## How Has This Been Tested?
Tested at UTokyo and TSAT.

## Types of changes

## Checklist:


agents/lakeshore425/LS425agent.py
```diff
@@ -0,0 +1,192 @@
...
+            return True, 'requested to stop taking data.'
+        else:
+            return False, 'acq is not currently running'
+    
+    def operational_status(self, session, params=None):
+        self.log.info(self.dev.OPST())
+        return True, 'operational status is checked'
+    
+    def zero_calibration(self, session, params=None):
+        self.log.info(self.dev.ZeroCalibration())
```
BrianJKoopman, line 135: `self.dev.ZeroCalibration()` -> `self.dev.set_zero()`

Also, does this `ZeroCalibration()` method return anything? It doesn't look like it. If that's the case, we don't need the `self.log.info()` wrapper here.

agents/lakeshore425/LS425agent.py
```diff
@@ -0,0 +1,192 @@
...
+        return True, 'Zero calibration is done'
+    
+    def anycommand(self, session, params=None):
+        #send serial command to Lakeshore 425
+        command = params['command']
+        print('Input command: ' + command)
+        print('Results: ', self.dev.anycommand(command))
+        return True, 'anycommand is finished cleanly'
```
BrianJKoopman, line 143: This should be updated for the new `command`/`query` methods in the Lakeshore425 class. You could mirror the split in distinguishing between "command" and "query" by adding a "query" task, or you could identify queries based on the presence of a '?', which is something we do in other Lakeshore driver code.

BrianJKoopman (changes requested):
Thanks for the PR, Kyohei! The Agent is looking good.

I have some thoughts on the driver code to keep things uniform with our other Lakeshore driver code/the Lakeshore Python library. I wrote those in an inline comment.

Other general comments for the Agent:
- On all Tasks/Processes you should make use of the params decorator. This will ensure values are passed from a Client correctly. For details see: https://ocs.readthedocs.io/en/develop/developer/agents.html#operation-parameters
- This Agent needs a documentation page located under [socs/docs/agents/](https://github.com/simonsobs/socs/tree/develop/docs/agents). See this page for details on how to write this documentation: https://ocs.readthedocs.io/en/develop/developer/agents.html#documentation
- For the Docker image builds the Agent should be added to [socs/docker-compose.yml](https://github.com/simonsobs/socs/blob/develop/docker-compose.yml). There are good examples in there already for other Agents, we should call this Agent Docker image `ocs-lakeshore425-agent`.
- There are some whitespace issues, here's some output from flake8 (I'm mostly concerned with trailing whitespace and the blank lines with whitespace):
```
LS425agent.py:10:26: W291 trailing whitespace
LS425agent.py:17:1: E302 expected 2 blank lines, found 1
LS425agent.py:104:1: W293 blank line contains whitespace
LS425agent.py:129:1: W293 blank line contains whitespace
LS425agent.py:133:1: W293 blank line contains whitespace
LS425agent.py:137:1: W293 blank line contains whitespace
LS425agent.py:145:1: E302 expected 2 blank lines, found 1
LS425agent.py:158:1: E302 expected 2 blank lines, found 1
LS425agent.py:164:1: W293 blank line contains whitespace
LS425agent.py:169:1: W293 blank line contains whitespace
LS425agent.py:191:1: E305 expected 2 blank lines after class or function definition, found 1
```

ykyohei:
I tried to use the decolator but I get the following error, what is wrong? version of ocs is old?
```
ls425_1                  |     @ocs_agent.param('command', type=str)
ls425_1                  | AttributeError: module 'ocs.ocs_agent' has no attribute 'param'
```

ykyohei:
I think all the required changes are done except for params decolator.
Read the Docs build is failed. Could you tell me how I can fix this?

BrianJKoopman:
> I tried to use the decolator but I get the following error, what is wrong? version of ocs is old?
> 
> ```
> ls425_1                  |     @ocs_agent.param('command', type=str)
> ls425_1                  | AttributeError: module 'ocs.ocs_agent' has no attribute 'param'
> ```

It could be, what version are you on? You can check with `python3 -c 'import ocs; print(ocs.__version__)'`. Ideally you'd update to at least v0.9.0 if you are behind.

BrianJKoopman (commented):
Thanks for all the changes! I've made a couple of small inline comments.

> I think all the required changes are done except for params decolator. Read the Docs build is failed. Could you tell me how I can fix this?

Right, the params issue likely is an ocs version problem like you pointed out. See my other comment responding to that.

The docs break because they import the Agent for the argparse section, which in turn tries to import the Lakeshore425 module, but it's unable to do so (the Agent is relying on the module being right next to it, but in the context of building the docs this isn't sufficient.) There are a couple ways to solve this, but I think what we should do is move that file into the socs library next to all the other Lakeshore driver code we've written.

To do that move that file to `/socs/socs/Lakeshore/` and then update the import line in the Agent to: `from socs.Lakeshore import Lakeshore425 as ls`

Note that if you're working with the Docker container locally you'll need to rebuild the socs image before building the Lakeshore425 Agent container in order for the code in the new location to be accessible.

This will fix this docs, as they install the socs package before building.

One other thing on the docs is that you should add a line in the [index file](https://github.com/simonsobs/socs/blob/develop/docs/index.rst) for the new Agent.

ykyohei:
I think all the requested changes are done.
Read the Docs is still failed.

BrianJKoopman:
> I think all the requested changes are done. Read the Docs is still failed.

I'm not sure why that was failing. I re-ran the build and it passed.

BrianJKoopman (commented):
Thanks for the last round of changes! Just to wrap things up with some last small comments I went ahead and opened #225 with some final changes. If those look alright to you we can merge those in here, and then merge into `develop`.

BrianJKoopman (approved):

PR #228, opened 2022-01-12
Device Emulator pytest Fixture

BrianJKoopman:
## Description

This PR introduces a fixture for emulating a hardware device for integration testing. It currently only works for a device communicating over serial.

The idea is to create a fixture, which is passed to each test. This provides an object which allows the Agent to connect to it via the Agent's typical connection mechanism (e.g. a serial or socket connection). This object listens for commands from the Agent in a new thread. Within each test you start by defining the commands and expected responses for your test. When the Agent sends a defined command during testing the Responder object simply sends back the defined response for that command to the Agent.

A simple example:
```python
# define the fixture, including any responses needed for Agent startup
responder = create_responder_fixture({'*IDN?': 'LSCI,MODEL425,4250022,1.0'})

@pytest.mark.integtest
def test_ls425_operational_status(wait_for_crossbar, responder, run_agent, client):
    # define the responses needed for this test
    responses = {'OPST?': '001'}
    responder.define_responses(responses)

    # run task that sends the 'OPST?' command
    resp = client.operational_status()
    assert resp.status == ocs.OK
    assert resp.session['op_code'] == OpCode.SUCCEEDED.value
```

The thing still missing from this is a docs page in Sphinx, but I'm hoping to get any feedback on this before I write that.

## Motivation and Context

This is similar to the work done on the Lakeshore 372 integration testing, but I think is more portable to emulating other devices for testing a wider range of Agents. In the Lakeshore 372 case we have an extensive hardware simulator program that tracks the internal state of the simulated 372 and generates random data. This would require a lot of work to do for each piece of hardware we have an Agent for.

This fixture takes an approach more like the unit testing done for the Lakeshore 372, where we simply send back a pre-defined response.

The context for development was for testing the new Lakeshore 425 Agent. You can see it in use on the [koopman/lakeshore425-integration-testing branch](https://github.com/simonsobs/socs/blob/koopman/lakeshore425-integration-tests/tests/integration/test_ls425_agent_integration.py).

## How Has This Been Tested?

It has been used to write integration tests for the Lakeshore 425 Agent. Here are some useful links:
- [Successful GH Actions run of integration tests](https://github.com/simonsobs/socs/runs/4792042642?check_suite_focus=true)
- [Code coverage output for those tests](https://coveralls.io/github/simonsobs/socs?branch=koopman/lakeshore425-integration-tests)
- [Working branch that uses the new fixture](https://github.com/simonsobs/socs/tree/koopman/lakeshore425-integration-tests)

## Types of changes

## Checklist:


BrianJKoopman:
I've renamed the `Responder` to `DeviceEmulator`, which I feel is a bit more descriptive. I've also moved it to be within the socs library, rather than just in the testing suite. I added a `relay_type` option to the function used to create the pytest fixture, for future use when the TCP relay is implemented. Lastly, I wrote some docs.

I'm going to self-merge this one.

PR #235, opened 2022-01-26
Scan helpers generator

laurensaunders:
## Description

Update the linear scan generator to improve functionality and testing ability.

## Motivation and Context

- Several values that were hard-coded in the previous version of this function are now optional arguments.
- Problems that previously printed a line and then returned now raise errors.
- The generator now handles values for scan endpoints where the time step size does not divide evenly into the difference between the endpoints. This fix prevents stall-outs and ensures that the correct flags are passed.

## How Has This Been Tested?

Generator was tested in a Jupyter notebook environment. Once the correct values were consistently given by the function, the Agent was updated and the generator process tested with the ACU in emulator mode.

## Types of changes

## Checklist:

### DIFF
--- a/agents/acu/scan_helpers.py
+++ b/agents/acu/scan_helpers.py
@@ -184,7 +184,7 @@ def generate_linear_turnaround(az_endpoint1, az_endpoint2, az_speed, acc, el_end
     el = el_endpoint1
     if step_time < 0.05:
         raise ValueError('Time step size too small, must be at least 0.05 seconds')
-    daz = round(step_time*az_speed, 4)
+    daz = step_time * az_speed
     el_vel = el_speed
     az_flag = 0
     if az < az_endpoint2:
@@ -195,13 +195,6 @@ def generate_linear_turnaround(az_endpoint1, az_endpoint2, az_speed, acc, el_end
         az_vel = -1*az_speed
     else:
         raise ValueError('Need two different motion endpoints')
-    az_remainder = (az_max - az_min) % daz
-    if az_remainder == 0.0:
-        leftover_az = 0.0
-        leftover_time = 0.0
-    else:
-        leftover_az = round(az_remainder, 4)
-        leftover_time = az_remainder / az_speed
     i = 0
     while i < stop_iter:
         i += 1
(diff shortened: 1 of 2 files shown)


laurensaunders:
I'm not sure whether the merge into postfat_updates is the correct one, mostly chosen because this is a branch off of postfat_updates.

laurensaunders:
@mhasself I don't think I can request a re-review because there are no requests for changes. I updated this to no longer rely on rounding

mhasself (changes requested):
New algorithm looks good!

I have some picky interface requests.  Also formatting, but I would understand if you prefer to leave all of that for Brian to specify.

agents/acu/scan_helpers.py
```diff
@@ -195,6 +200,15 @@ def generate_linear_turnaround(az_endpoint1, az_endpoint2, az_speed, acc, el_end
...
         raise ValueError('Need two different motion endpoints')
+    if num_batches == None:
+        stop_iter = float('inf')
+    else:
+        stop_iter = num_batches
+        leftover_in_leg = (az_max - az_min) % daz
+        if leftover_in_leg == 0:
+            batch_size = (az_max - az_min) / daz
+        else:
+            batch_size = (az_max - az_min) / daz + 1
```
mhasself, line 247: This computation leaves batch_size as a float, and then range(batch_size) fails later.

If you are looking for a "round up" function, it is np.ceil (ceiling):
```
batch_size = int(np.ceil((az_max - az_min) / daz))
```

If I fix that, and pass num_batches=2, I get this trajectory:
image

Note how there is a single "dangling" point for the next scan.  Also, if you zoom in at t=0, you'll find there's no point there.

I think you should probably move your ```point_block[0].append(t + t0) ...``` lines from the end of each iteration to the start of each iteration.  That will fix the dangling point, and make sure you're actually starting from the starting point.

mhasself (changes requested):
Sorry, two more little things!

agents/acu/scan_helpers.py
laurensaunders: This suggestion doesn't quite work (some cases had early dangling points instead of late ones). I think the new push fixes it though

agents/acu/scan_helpers.py
```diff
@@ -240,16 +240,19 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
...
         i += 1
         point_block = [[],[],[],[],[],[],[]]
         for j in range(batch_size):
+            point_block[0].append(t + t0)
+            point_block[1].append(az)
+            point_block[2].append(el)
+            point_block[3].append(az_vel)
+            point_block[4].append(el_vel)
+            point_block[5].append(az_flag)
+            point_block[6].append(el_flag)
```
mhasself, line 255: Thanks -- but el_flag is not initialized outside the loop, so this causes an error.

Please test your code after you make changes!

agents/acu/scan_helpers.py
mhasself: That's odd, I really thought ceil was right.

In the latest code, I'm getting dangling points in the case that daz divides (az_max - az_min) exactly:

image

```
sh.generate_constant_velocity_scan(
        10, 1., 1., 1.2,
        30, 30, 0,
        step_time=0.05,
        start_time=0.,
        num_batches=2,
        ptstack_fmt=False)
```

laurensaunders: I think I figured out the issue. np.ceil looked incorrect because of the +1, but it is working with np.ceil and removing the +1

mhasself (approved):
Thanks!

PR #236, opened 2022-01-27
Acu commanded position None handling

laurensaunders:
## Description

Previously, the None values from the ACU status stream were converted to zeros to ensure the data type never changed for the aggregator, and because Influx would not accept float('nan') values.

This update changes handling so the None values are recorded as float('nan') in g3 files. The commanded position values are then sent to their own feeds, which only publish to Influx, and are only published with values are not 'nan'.

## Motivation and Context

Recording the values in g3 files as nan gives no false impressions that the commanded position is actually 0. Splitting the influx publishing into 3 separate feeds for these values allows them to only be published when they have non-nan values.

## How Has This Been Tested?

This update has been tested using the ACU in emulator mode at UCologne.
## Types of changes

## Checklist:

### DIFF
--- a/agents/acu/acu_agent.py
+++ b/agents/acu/acu_agent.py
@@ -204,10 +204,6 @@ class ACUAgent:
                                   record=True,
                                   agg_params=fullstatus_agg_params,
                                   buffer_time=1)
-        self.agent.register_feed('acu_status_commands',
-                                  record=True,
-                                  agg_params=fullstatus_agg_params,
-                                  buffer_time=1)
         self.agent.register_feed('acu_status_general_errs',
                                   record=True,
                                   agg_params=fullstatus_agg_params,
@@ -221,6 +217,18 @@ class ACUAgent:
                                   agg_params=fullstatus_agg_params,
                                   buffer_time=1)
         self.agent.register_feed('acu_status_influx',
+                                 record=True,
+                                 agg_params=influx_agg_params,
+                                 buffer_time=1)
+        self.agent.register_feed('acu_commands_az_influx',
+                                 record=True,
+                                 agg_params=influx_agg_params,
+                                 buffer_time=1)
+        self.agent.register_feed('acu_commands_el_influx',
+                                 record=True,
+                                 agg_params=influx_agg_params,
+                                 buffer_time=1)
+        self.agent.register_feed('acu_commands_bs_influx',
                                  record=True,
                                  agg_params=influx_agg_params,
                                  buffer_time=1)
@@ -397,7 +405,7 @@ class ACUAgent:
                         elif type(value) == int or type(value) == float:
                             self.data['status'][category][self.monitor_fields[category][key]] = value
                         elif value == None:
-                            self.data['status'][category][self.monitor_fields[category][key]] = float(0.0)#'None'
+                            self.data['status'][category][self.monitor_fields[category][key]] = float('nan')#'None'
                         else:
                             self.data['status'][category][self.monitor_fields[category][key]] = str(value)
             self.data['status']['summary']['ctime'] =\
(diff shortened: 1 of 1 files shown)


agents/acu/acu_agent.py
```diff
@@ -407,25 +419,47 @@ def start_monitor(self, session, params=None):
...
-                        if self.data['status'][category][statkey] in ['None', 'True', 'False']:
-                            influx_status[statkey + '_influx'] = tfn_key[self.data['status'][category][statkey]]
-                        else:
-                            influx_status[statkey + '_influx'] = mode_key[self.data['status'][category][statkey]]
-                    elif type(self.data['status'][category][statkey]) == int:
-                        if statkey in ['Year', 'Free_upload_positions']:
-                            influx_status[statkey + '_influx'] = float(self.data['status'][category][statkey])
+                        elif type(self.data['status'][category][statkey]) == str:
+                            if self.data['status'][category][statkey] in ['None', 'True', 'False']:
+                                influx_status[statkey + '_influx'] = tfn_key[self.data['status'][category][statkey]]
```
mhasself, line 434: This will turn 'None' into a float, but 'True' and 'False' into ints.  So I think this should be two separate "if" blocks.

agents/acu/acu_agent.py
```diff
@@ -407,25 +419,47 @@ def start_monitor(self, session, params=None):
...
+                            else:
+                                influx_status[statkey + '_influx'] = mode_key[self.data['status'][category][statkey]]
+                        elif type(self.data['status'][category][statkey]) == int:
+                            if statkey in ['Year', 'Free_upload_positions']:
```
mhasself, line 438: The hard-coding of schema in the Agent isn't great.  For now we can leave it but I think we'll want to move to a system where we have a config file that lists each field (you already have that, in soaculib.status_keys) and also describes any non-standard processing to do on that field.

mhasself (changes requested):
The logic for deciding how to condition status data has nested levels... consider flattening that (removing the nesting when it's not needed).  For example, you only convert 'Year' and 'Free_upload_positions' to float if the data are int.  But we know those fields will always present as int, and furthermore we want to crash if they aren't int (or more accurately, if they would not cleanly convert to float).  So it would make sense to do the processing of 'Year' and 'Free_upload_positions' without first checking the type of the data.   Similarly, you don't have to know that val is string to test (val in ['True', 'False']).  So that doesn't need to be inside the string test.

Around line 533 you still have a try/except for publishing to influx.  Does this still need to be protected?  If so, improve the error message.
```
            try:
                self.agent.publish_to_feed('acu_status_influx', acustatus_influx, from_reactor=True)
            #    print(acustatus_influx)
            except:
#                print(acustatus_influx)
                print('failed')
```

agents/acu/acu_agent.py
```diff
@@ -224,6 +224,22 @@ def __init__(self, agent, acu_config='guess'):
...
+ #                                agg_params=influx_agg_params,
+ #                                buffer_time=1)
+ #       self.agent.register_feed('acu_commands_el_influx',
+ #                                record=True,
+ #                                agg_params=influx_agg_params,
+ #                                buffer_time=1)
+ #       self.agent.register_feed('acu_commands_bs_influx',
+ #                                record=True,
+ #                                agg_params=influx_agg_params,
+ #                                buffer_time=1)
```
mhasself, line 238: Remove old code.

agents/acu/acu_agent.py
```diff
@@ -407,25 +423,43 @@ def start_monitor(self, session, params=None):
...
-                            influx_status[statkey + '_influx'] = float(self.data['status'][category][statkey])
-                        else:
-                            influx_status[statkey + '_influx'] = int(self.data['status'][category][statkey])
-                    else:
-                        print(statkey)
-
+                if category != 'commands':
+                    for statkey, statval in self.data['status'][category].items():
+                        if type(statval) == float:
+                            influx_status[statkey + '_influx'] = statval#self.data['status'][category][statkey]
```
mhasself, line 429: remove comment (and those below)

agents/acu/acu_agent.py
```diff
@@ -407,25 +423,43 @@ def start_monitor(self, session, params=None):
...
+                                influx_status[statkey + '_influx'] = tfn_key[statval]#self.data['status'][category][statkey]]
+                            else:
+                                influx_status[statkey + '_influx'] = mode_key[statval]#self.data['status'][category][statkey]]
+                        elif type(statval) == int:
+                            if statkey in ['Year', 'Free_upload_positions']:
+                                influx_status[statkey + '_influx'] = float(statval)#self.data['status'][category][statkey])
+                            else:
+                                influx_status[statkey + '_influx'] = int(statval)#self.data['status'][category][statkey])
+                       # else:
+                       #     print(statkey)
```
mhasself, line 443: Remove commented block

agents/acu/acu_agent.py
```diff
@@ -407,25 +423,43 @@ def start_monitor(self, session, params=None):
...
+                elif category == 'commands':
+                    if str(self.data['status']['commands']['Azimuth_commanded_position']) != 'nan':
+                        acucommand_az = {'timestamp': self.data['status']['summary']['ctime'],
+                                         'block_name': 'ACU_commanded_positions',
```
mhasself, line 447: Thanks for making this one feed.  Technically, you should give each field set a different block_name.  They are just for  internal bookkeeping so you can just call them 'az', 'el', 'boresight'.

Since this is an a feed for influx instead of the aggregator, using the same value ("ACU_commanded_positions") for all three sets of fields might not cause trouble... but it's good to be consistent.

mhasself (changes requested):
Thanks -- one last minor fix and we should be good.

mhasself (approved):
Thanks!

PR #239, opened 2022-02-02
Create PWV Web API

BrianJKoopman:
## Description

This PR creates a small Flask app for serving the latest PWV value via a web API.

I've temporarily imported the required read function, which I moved out of the Agent to this new `pwv_web` module, so the Agent should work as it did before. The Agent should be updated to read data from this web application, rather than straight from file, then this temporary import can be removed.

Details for how this will be deployed still need to be worked out, but I think it's likely to be with gunicorn in a systemd service or similar, possibly behind an nginx proxy. However these details aren't needed at the moment for further Agent development.

## Motivation and Context

The PWV data is collected and kept on a separate computer from the OCS Agents. This web server will allow OCS, as well as other interested parties, to get the latest PWV reading from the radiometer via a simple GET request.

## How Has This Been Tested?

I've run the app locally with a sample PWV data file.


sanahabhimani (commented):
No comments! Tested successfully in my local setup.

PR #240, opened 2022-02-03
Build and Deploy socs to PyPI

BrianJKoopman:
## Description

Following https://github.com/simonsobs/ocs/pull/243, this PR sets up socs to build and deploy the package to PyPI.

## Motivation and Context

Adding another package to the list of SO DAQ python modules now published to PyPI. Hopefully makes installation easier for end users.

## How Has This Been Tested?

I've built and installed the wheel locally, and published a release candidate to PyPI already: https://pypi.org/project/socs/

The workflow hasn't been fully tested yet, but it is a copy of the working OCS workflow changes.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,15 +21,9 @@ jobs:
 
     # (start steps from pytest workflow)
     # Install
-    - name: clone ocs
-      uses: actions/checkout@v2
-      with:
-        repository: 'simonsobs/ocs'
-
-    - name: Install ocs
+    - name: Install so3g
       run: |
-        pip3 install -r requirements.txt
-        pip3 install .
+        pip3 install so3g
 
     # Fetch all history for all tags and branches
     - name: clone socs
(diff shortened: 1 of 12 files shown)


BrianJKoopman:
I'm unsure why that coverage is dropping, but I can reproduce it locally, so I'll try to figure that out before moving forward with this. Converting to draft for now.

BrianJKoopman:
Alright, I figured out that the coverage drop was due to [4699f31](https://github.com/simonsobs/socs/pull/240/commits/4699f31b98cc9f82c120140c4feeae19592fe283). The coverage determination is relying on the path for the module when `-e` is used. Dropping the use of containers for running the tests and instead installing all dependencies directly in the workflow environment and still using `-e` there works (and is something we should do now that so3g is easily pip installable anyway.)

I'm still confused why the LS425 auto acquire test seems to no longer report its coverage, but it's only 3 lines, and I'll likely dig into that when expanding test coverage in a later PR.

BrianJKoopman:
Alright, tested the full workflow with a tag in f5b81917099ac525b7c9cf0980db9eb8bc711440. Will self merge this one by end of day.

PR #245, opened 2022-02-07
Expand base test coverage for agents

BrianJKoopman:
## Description

This PR sets up a bunch of skeleton testing files for the Agents which simply import the Agents. This is to get an accurate assessment of the total coverage across all the Agents and the socs library.

My focus is on Agents we're going to field, so I'm explicitly excluding some code from coverage that I don't care to test at the moment, like the Lakeshore370 and Lakeshore336 Agents, which both are only used in labs. PRs for tests on these are certainly welcome, I'm just not going to prioritize it.

Some other Agents that are left out at the moment that are difficult to test:
- ACU Agent - This needs the soaculib library, which isn't public at the moment.
- Labjack Agent - This one we could probably do, but we'll need to setup and install the driver module in the CI environment. I'll leave it out for now, but we should add that.
- LATRtXYStage Agent - This has some external library dependence, and I'm not sure if this is going to see use in the field or if it's lab use only. Maybe @kmharrington can comment?

Misc updates:
- Update the 372 integration tests to use new ocs testing fixtures
- Add cancel previous run actions to commonly run workflows

## Motivation and Context

I wanted to establish a base level of coverage, which is currently hidden since Agents won't be included in the coverage report unless imported.

## How Has This Been Tested?

Tested locally and in CI pipeline on commit.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -9,6 +9,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Cancel Previous Runs
+      uses: styfle/cancel-workflow-action@0.9.1
+      with:
+        access_token: ${{ github.token }}
+
     - name: Install system packages
       run: |
         sudo apt-get install -y socat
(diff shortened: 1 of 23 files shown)


kmharrington:
While it is possible the LATRt motor controller makes it to the field it will not be a permanently installed device there. So I think it's fine to leave it out for these tests.

PR #248, opened 2022-02-15
Fix npy extension issue

jlashner:
This should fix a small bug in the pysmurf-monitor where numpy files that were registered without the `.npy` extension were not being added to the database correctly. This is because numpy will automatically add that extension if it doesn't exist, meaning the registered file path is not the same as the real file path.

The monitor used to catch this and fix it, but looks like I missed that when I was restructuring the agent. This adds it back.

## Description

Catches for npy files without the .npy extension.

## Motivation and Context

We need to archive tunefiles

## How Has This Been Tested?

Not tested yet, but can test later this week on k2so (or maybe @kmharrington can test before then?)

## Types of changes

## Checklist:


BrianJKoopman (approved):
Looks good to me!

PR #258, opened 2022-02-23
pip install socs directly from socs repo in pysmurf-controller image

BrianJKoopman:
## Description

This effectively replaces installing dependencies via requirements.txt with those specified in setup.py. This avoids installing so3g, which we don't need here, and which doesn't support python3.6, which is what the sodetlib image is on.

## Motivation and Context

#257 added so3g to the requirements file, which [broke](https://github.com/simonsobs/socs/runs/5305731607?check_suite_focus=true) the pysmurf-controller image build.

```
Collecting so3g (from -r socs/requirements/testing.txt (line 6))
  Could not find a version that satisfies the requirement so3g (from -r socs/requirements/testing.txt (line 6)) (from versions: )
No matching distribution found for so3g (from -r socs/requirements/testing.txt (line 6))
The command '/bin/sh -c git clone https://github.com/simonsobs/socs.git     && pip3 install -r socs/requirements.txt     && pip3 install -e socs' returned a non-zero code: 1
Service 'ocs-pysmurf-agent' failed to build : Build failed
```

## How Has This Been Tested?

I'm able to build the image locally now, however it would be good to test it runs alright on a system.

## Types of changes

## Checklist:

### DIFF
--- a/agents/pysmurf_controller/Dockerfile
+++ b/agents/pysmurf_controller/Dockerfile
@@ -3,9 +3,7 @@ FROM simonsobs/sodetlib:v0.3.0-101-gccbc921
 WORKDIR /app
 
 # SOCS installation
-RUN git clone https://github.com/simonsobs/socs.git \
-    && pip3 install -r socs/requirements.txt \
-    && pip3 install -e socs
+RUN python3 -m pip install -e git+https://github.com/simonsobs/socs.git#egg=socs
 
 ENV OCS_CONFIG_DIR /config
 
--- a/agents/pysmurf_controller/Dockerfile
+++ b/agents/pysmurf_controller/Dockerfile
@@ -3,7 +3,7 @@ FROM simonsobs/sodetlib:v0.3.0-101-gccbc921
 WORKDIR /app
 
 # SOCS installation
-RUN python3 -m pip install -e git+https://github.com/simonsobs/socs.git#egg=socs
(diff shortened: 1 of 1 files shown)


jlashner (approved):
Just tested building this at UCSD and it seems to run. A perhaps unintended consequence is that it changes the location of the `socs` repo inside of the docker image from `/app/socs` to `/app/src/socs`. However the pysmurf-controller directory is still copied to `/app/socs/agents/pysmurf_controller`, so mounting in a dev copy of `socs` to `/app/socs` still works (which is how I like to develop). 

Since this still works fine I think we can probably just merge, but technically I think it would be better if the socs directory path was consistent with other socs dockers.

BrianJKoopman:
> Just tested building this at UCSD and it seems to run. A perhaps unintended consequence is that it changes the location of the `socs` repo inside of the docker image from `/app/socs` to `/app/src/socs`. However the pysmurf-controller directory is still copied to `/app/socs/agents/pysmurf_controller`, so mounting in a dev copy of `socs` to `/app/socs` still works (which is how I like to develop).
> 
> Since this still works fine I think we can probably just merge, but technically I think it would be better if the socs directory path was consistent with other socs dockers.

Great, thanks for testing! I made a small change so that pip clones to `/app/socs/`, so things should be consistent now. Will merge when checks finish.

PR #270, opened 2022-03-24
Separate shared code from several PRs

BrianJKoopman:
## Description

This pulls out `moxaSerial.py` and `pmx.py` from #157, adds testing, and makes some small modifications to the way encoding/decoding the messages works.

## Motivation and Context

#269, #183, and #157 all use `moxaSerial.py`, and #183 and #157 both use `pmx.py`. These submitted Agents should all use the same source for this.

To update your code you will need to update imports, and usage of `Serial_TCPServer.write()`. The new imports will be:
```
from socs.agent.moxaSerial import Serial_TCPServer
from socs.agent.pmx import PMX, Command
```
(Note that `Command` here used to be separated into `command.py`.)

Also, wherever you use `Serial_TCPServer.write()`, you likely were previously encoding every string you passed, either with `str.encode()` or by submitting a bytes string, i.e. `b'MSG\r'`. You no longer need to do this, and can just use a string. The encoding now happens within `Serial_TCPServer.write()`.

Similarly, decoding happens when `Serial_TCPServer` reads a response from a given device, so you no longer need to decode responses yourself either.

## How Has This Been Tested?

I wrote some tests for the methods in these two modules that I noticed among all the PRs were used. I think the responses I made up are reasonable, but please take a look and suggest more realistic/correct responses if you know them. These are the dictionaries in the `test_moxaSerial.py` and `test_pmx.py` files, i.e.:
```
{'OUTP OFF': '',
 'OUTP?': 'off'}
```

## Types of changes

## Checklist:

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,25 @@
+API
+===
+
+This page contains the auto-generated documentation for the socs package.
+
+socs.agent
+----------
+
+The ``agent/`` directory contains code that supports one or more socs Agents.
+
+socs.agent.moxaSerial
+`````````````````````
+
+.. automodule:: socs.agent.moxaSerial
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
+socs.agent.pmx
+``````````````
+
+.. automodule:: socs.agent.pmx
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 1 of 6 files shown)


BrianJKoopman:
I cleaned up the formatting of these two files, and moved a bunch of documentation that was in `moxaSerial.py` from being block comments at the top of the file to docstrings for each relevant class and method. Going to go ahead and merge this to develop for use in the related PRs.

PR #276, opened 2022-03-28
Add functionality for LS372 Agent to Initialize Device/Channel Settings Using Config File

sanahabhimani:
## Description
Wrote a new method to input a formatted .yaml config file to upload channel and device settings (autoscan, enable/disable channel, dwell, pause, calibration curve number) to a 372 upon agent initialization. This method is only used during 'init' mode of the 372 agent, as it's only meant to be used on agent startup. 

## Motivation and Context
Keeping up with changes made across channels for each 372 for each telescope is not well-organized. Introducing a 372 config file + input method for the agent allows the user to initialize a desired set of controlled settings for the device across each telescope, i.e., how may channels to enable upon startup, the calibration curve associated with each channel, the dwell, etc. 

## How Has This Been Tested?

Tested on docker containerized Lakeshore 372 simulator, as well as a real 372 in the lab at Yale. 

## Types of changes

## Checklist:


agents/lakeshore372/LS372_agent.py
```diff
@@ -977,6 +985,79 @@ def get_still_output(self, session, params=None):
 
         return True, "Current still output is {}".format(still_output)
 
+    def input_configfile(self, configfile, filepath='/ls372configs/'):
```
BrianJKoopman, line 988: Since this isn't a Agent Task or Process, and is only called internally within the Agent, we should mark it as "private" by naming it with a leading underscore, i.e. `_input_configfile()`. This'll keep it off the automatically generated list of tasks/processes in the Agent docs.

agents/lakeshore372/LS372_agent.py
```diff
@@ -977,6 +985,79 @@ def get_still_output(self, session, params=None):
...
+        """ Upload 372 configuration file to initialize channel/device
+        settings upon Agent startup.
+
+        Args:
+            configfile (str): name of .yaml config file
+        """
+        # include /ls372configs/ dir where configfiles are in docker container
+        ls372config = os.path.join(filepath, configfile)
```
BrianJKoopman, line 996: To use `$OCS_CONFIG_DIR` here you'll need to access the environment variable via `os.environ.get("OCS_CONFIG_DIR", "/config/")`. (`OCS_CONFIG_DIR` is set in the containers to `/config` by default, but this default to that in the rare case it isn't set.)

agents/lakeshore372/LS372_agent.py
```diff
@@ -977,6 +985,79 @@ def get_still_output(self, session, params=None):
...
+        with open(ls372config) as f:
+            config = yaml.safe_load(f)
+
+        ls = self.module
+        ls_id = self.module.id
+        ls_info = ls_id.split(',')
+        ls_serial = ls_info[2]
```
BrianJKoopman, line 1003: Could be a one-liner: `ls_serial = self.module.id.split(',')[2]`

agents/lakeshore372/LS372_agent.py
```diff
@@ -977,6 +985,79 @@ def get_still_output(self, session, params=None):
...
+
+        # TODO: assert 372 instance id == serial num?
+
+        device_config = config[ls_serial]['device_settings']
+        ls_chann_settings = config[ls_serial]['channel']
+
+        # enable/disable autoscan
+        if device_config['autoscan'] == 'on':
+            ls.enable_autoscan()
+            print('Autoscan enabled')
```
BrianJKoopman, line 1013: It'd be better to use `self.log.info()` for all print statements here.

agents/lakeshore372/LS372_agent.py
```diff
@@ -977,6 +985,79 @@ def get_still_output(self, session, params=None):
...
+        elif device_config['autoscan'] == 'off':
+            ls.disable_autoscan()
+            print('Autoscan disabled')
+
+        for i in ls_chann_settings:
+            lschann = Channel(self.module, i)
```
BrianJKoopman, line 1019: The `LS372` object already has objects for each `Channel` initialized, so I don't think you should reinitialize new `Channel` objects here, but instead use those already in the `ls` instance here.

agents/lakeshore372/ls372_configfiles.yaml
```diff
@@ -0,0 +1,24 @@
+LSASIM:
```
BrianJKoopman, line 1: So I think rather than including a file among the Agent code we should provide an example in the documentation, specifically on the page [docs/agents/lakeshore372.rst](https://github.com/simonsobs/socs/blob/develop/docs/agents/lakeshore372.rst).

Since labs use docker for this installation, they don't necessarily have a copy of the repo even.

agents/lakeshore372/LS372_agent.py
```diff
@@ -990,23 +1071,25 @@ def make_parser(parser=None):
...
+            choices=['idle', 'init', 'acq'],
+            help="Starting action for the Agent.")
     pgroup.add_argument('--sample-heater', type=bool, default=False,
-                        help='Record sample heater output during acquisition.')
+            help='Record sample heater output during acquisition.')
     pgroup.add_argument('--enable-control-chan', action='store_true',
-                        help='Enable reading of the control input each acq cycle')
+            help='Enable reading of the control input each acq cycle')
+    #config file
+    pgroup.add_argument('--configfile', type=str, help='Yaml file for initializing 372 settings')
```
BrianJKoopman, line 1092: Happy to see the help string, this'll automatically be added to the arguments list on the docs page. But I think we should also add an example to the ocs-site-configs block on that docs page.

BrianJKoopman (changes requested):
Adding this functionality is going to be great! I do have some changes I'd like made before merging, see below. Let me know if you have any questions about the comments or suggestions.

agents/lakeshore372/LS372_agent.py
BrianJKoopman: An update to this after the call (and to summarize @mhasself's point) -- this could be made into a Task so that we can run it separately to re-establish the device state. In which case, we don't want to mark this as private.

BrianJKoopman (approved):

PR #281, opened 2022-04-19
Implement hwp rotation agent review suggestions, remove need for telnet, add tests

BrianJKoopman:
## Description

This work mostly wraps up the HWP Rotation Agent, implementing suggestions from #157, and making some other changes. Summary of some of the bigger changes:
- [Remove agent copy of moxaSerial and pmx modules](https://github.com/simonsobs/socs/commit/07d45b3e9982c0c11f5c2e9b5ae2a4f8f3ce0d47)
- [Add iv_acq data to session.data](https://github.com/simonsobs/socs/commit/3f050640887500eaf0e5939d0bbd00df0cbf85ce)
- [Replace telnet with TCP connection PID controller](https://github.com/simonsobs/socs/commit/634ec8506a136aec6e50e21e2fd7b27346f1f8bf)
- [Fix process TimeoutLock usage](https://github.com/simonsobs/socs/commit/f5596b87123dfa779ea8d2ecc8ba5baaa3d5c1f8) (enabling the removal of the switching lock mechanism)
- [Add flag for running iv_acq process at startup](https://github.com/simonsobs/socs/commit/66db0d5ce713ca81f3c247a9d8e81f988f00285c)
- [rotation_agent: Rename pid parameters for set_pid task](https://github.com/simonsobs/socs/commit/a7ec1726ecc6630f2fdbeddff854cc025bdf8da3) (requires updates to any existing client scripts setting the PID parameters)
- [rotation_agent: Remove switching lock mechanism](https://github.com/simonsobs/socs/commit/fde8b5af5cfe240c112a147d984866e9f9204c6f)

## Motivation and Context

Mostly this was wrapping up #157. One of the largest changes was f5596b87123dfa779ea8d2ecc8ba5baaa3d5c1f8, replacing the use of telnet via bash scripts with sending commands via python's `socket` module.

### `set_scale`
In implementing these changes, and the tests associated with them, I found that the default parameters that were set in the Agent for the `pid_controller`'s `set_scale` method for the "offset" parameter don't work:

```
    >>> PID.get_scale_hex(0, 2)
    Traceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/home/koopman/git/socs/agents/hwp_rotation/src/pid_controller.py", line 49, in get_scale_hex
        expo = int(str(num*10**40 + 0.01).split('+')[1])-40
    IndexError: list index out of range
```

I have changed the default from 0 to 0.1 in order to get something that works, but I don't know that this is a sensible default. If 0 is a valid input, this bug should be fixed and the default reverted.

### Splitting Agent
One thing that I realized in working on this agent was that there is a group of tasks that only deal with the PID controller, and a group that only deal with the Kikusui PSU. Probably this Agent should be split into two different Agents, one for each piece of hardware. Is there any compelling reason to not do this, perhaps related to the way the hardware works? i.e. the PID controller can't do anything if the PSU isn't on.

## How Has This Been Tested?

I've tested components of this (mostly the TCP communication with the PID controller) on the hardware, but not yet the full Agent. The accompanying tests do test all tasks/processes within the Agent with emulated hardware connections.

## Types of changes

## Checklist:


BrianJKoopman:
@bbixler500, this largely wraps up any concerns I had with the rotation agent. There's a few lingering questions I have (see description above). If you'd like to take a look at the changes, let me know. Else, I'm happy to merge this into #157, then move onto to merging into the `develop` branch.

This should be tested with the hardware whenever there's a convenient time for that. I've mocked up the hardware communication for testing, so I don't anticipate running into any major problems.

BrianJKoopman:
We just tested this with the hardware in the lab at UCSD. Found some bugs, fixed, going to merge.

PR #283, opened 2022-04-25
Smurf file emulator realistic g3

jlashner:
This modifies the smurf file emulator so that the output g3 files follow the frame structure in the real g3 files.

## Description

This makes it so that files output by the smurf-file-emulator are identical to those output by the smurf-streamer. 

Each streaming session will start with a Observation frame and a Wiring frame with a status dump, followed by a continuous stream of Scan frames. No "status update" frames are sent, but the status content in the scan frame is copied directly from a real status frame, so this should be sufficient to read in the observation with G3tSmurf.

## Motivation and Context

This is required for end-to-end testing with the bookbinder.

## How Has This Been Tested?

I've run coverage tests on all the functions and manually checked the output g3 files and verified that they have the correct frame structure.

## Types of changes

## Checklist:


BrianJKoopman (approved):
This looks great, thanks Jack. And thanks for the tests, made it simple to test out myself.

Just a small typo and stray unused import from some of the modifications to fix.

PR #284, opened 2022-04-26
Scpi psu monitor update

mjrand:
Added ability to monitor subset of channels

## Description
I added a line to parse the parameters of the start_monitor function for a list of specific channels to monitor, instead of the function monitoring all channels only. By default, if no channels are specified, the function works in the same way it did before the change.

## Motivation and Context
In some circumstances, trying to monitor all channels will crash the agent because the psu won't respond to messages requesting the voltage/current on that channel. This causes the socket to timeout on the socket.read() call in the PrologixInterface class. This occurred when one channel on the psu was set to "in series" mode. By allowing the user to specify which channels to monitor, we can avoid this problem by just not monitoring the problem channel. 

## How Has This Been Tested?
This was tested using a local image in the UCSD lab. I tested monitoring a single channel.

## Types of changes

## Checklist:

### DIFF
--- a/agents/scpi_psu/scpi_psu_agent.py
+++ b/agents/scpi_psu/scpi_psu_agent.py
@@ -57,18 +57,17 @@ class ScpiPsuAgent:
             Args:
                 wait (float, optional):
                     time to wait between measurements [seconds].
-                    
-                channels (integer list, optional):
-                    channels to monitor [monitors all channels by default]
+                channels (list[int], optional):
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
This looks good to me.

I'm going to clean up some of the extra whitespace added (and some missing), and rebase, as this is based on a quite old commit, though there are no conflicts.)

PR #287, opened 2022-04-28
Configure pre-commit

BrianJKoopman:
## Description

This PR configures [pre-commit](https://pre-commit.com/) to run a few different hooks. The base hooks, provided by pre-commit (with descriptions from https://pre-commit.com/hooks.html):

- check-ast - simply checks whether the files parse as valid python.
- check-yaml - checks yaml files for parseable syntax.
- end-of-file-fixer - ensures that a file is either empty, or ends with one newline.
- trailing-whitespace - trims trailing whitespace.

Additional hooks (from external plugins):
- autopep8 - runs [autopep8](https://github.com/hhatto/autopep8), which automatically formats Python code to conform to [PEP8](https://peps.python.org/pep-0008/).
- flake8 - runs [flake8](https://github.com/PyCQA/flake8), catching any linting that autopep8 doesn't auto-format.

As is in this PR, this is not required for developers who do not wish to use it. However I do think we should encourage its use.

## Motivation and Context

The base hooks seem sensible, ensuring python and yaml are valid, enforcing proper ends to files, and trimming whitespace (which often sneaks in).

As for the additional hooks, I'm often running flake8, but sometimes forget to. This ensures I won't forget. autopep8 I found recently, and I have found it to be very time saving when working with code that has a lot of the smaller PEP8 inconsistencies, i.e. missing whitespace, lines too long, etc. It's a quick way to make the code more readable, and to clear up the output of flake8 for the more important things that it catches.

The configuration of flake8/autopep8 are both done in the `.flake8` configuration. And this is very minimally configured at the moment. I expect, as I use this to update this configuration to ignore simpler flake8 codes that we don't care to enforce.

Eventually, and if usage proves...useful, I would like to move to using [pre-commit.ci](https://pre-commit.ci/), which would automatically run `pre-commit` on modified files in each PR. If changes are needed, i.e. it trims some whitespace in your commits, it'll make them and commit them automatically.

I'm not doing this yet, as it forces the usage on everyone, and at the moment it looks like that also means it'll run on all files in the repository and try to fix them. We have a lot of clean up to do before we can do that though.

For my own reference later, [this post](https://stackoverflow.com/a/67611150) describes the choices to run pre-commit automatically, stating that a github action for running is no longer encouraged for new usage. However, [this issue](https://github.com/pre-commit/action/issues/7) describes how to have pre-commit only run on changed files, not all files, when modifications are needed. I haven't found that functionality in pre-commit.ci yet, so if that is used, I think we have to clean things up first.

I'd be curious to hear thoughts from @mhasself and @jlashner on this. I know we discussed auto-formatters before and decided against it. However, autopep8 has again been pretty useful to me recently, and isn't as strict as say using [black](https://github.com/psf/black). 

## How Has This Been Tested?

I've run `pre-commit` locally on the repo.

## Types of changes

## Checklist:

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,10 @@
+[flake8]
+max_line_length = 80
+ignore = E501
+in-place = true
+recursive = true
+aggressive = 2
+exclude =
+    agents/meinberg_m1000/mibs/MBG-SNMP-LTNG-MIB.py
+    agents/meinberg_m1000/mibs/SNMPv2-MIB.py
+    agents/meinberg_m1000/mibs/MBG-SNMP-ROOT-MIB.py
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -7,6 +7,8 @@ on:
       - 'docs/**'
       - '*.rst'
       - '*.md'
+      - '.flake8'
+      - '.pre-commit-config.yaml'
 
 jobs:
   test:
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -7,6 +7,8 @@ on:
       - 'docs/**'
       - '*.rst'
       - '*.md'
+      - '.flake8'
+      - '.pre-commit-config.yaml'
   workflow_call:
 
 jobs:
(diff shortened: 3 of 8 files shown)


BrianJKoopman:
Going to merge this after having discussed it briefly on our call Wednesday.

PR #293, opened 2022-05-06
DLM agent

### LINKED ISSUE #60: DLM Power Supply Agent
Create agent for control and readout of the DLM Power Supply on the LAT.

tanaybhandarkar:
First PR for the DLM agent. This agent is used to communicate with the Sorenson DLM power supply, used for powering the LATR's 4k heaters and magnetic field tests. 

## Description
Completed the data acquisition functions, added a function to set current values, and adding tasks for clients (tasks will control voltage and current outputs, and overvoltage protections values).

## Motivation and Context
This is the first PR for the DLM agent. It will be used for DAQ purposes at LATR labs and in deployment.

Closes #60.

## How Has This Been Tested?
This has been tested with the Sorenson DLM hardware at Penn. It can publish voltage and current readings to a feed, and I have written a client that can set voltages (not in this PR). 

## Types of changes

## Checklist:

Regarding tasks and process failing to get locks: If I use the set_voltage task (or any task), I need to stop the acq process otherwise the task does not get the lock. I think there was a discussion about this a while back, but I cannot find the thread, what is the correct way to do this?

### DIFF
--- a/agents/sorenson_dlm/Dockerfile
+++ b/agents/sorenson_dlm/Dockerfile
@@ -7,11 +7,11 @@ FROM socs:latest
 
 # Set the working directory to registry directory
 
-WORKDIR /app/socs/agents/sorenson-dlm/
+WORKDIR /app/socs/agents/sorenson_dlm/
 
 # Copy this agent into the app/socs/agents directory
 
-COPY . /app/socs/agents/sorenson-dlm/
+COPY . /app/socs/agents/sorenson_dlm/
 
 # Run agent on container startup
 
--- a/agents/sorenson_dlm/Dockerfile
+++ b/agents/sorenson_dlm/Dockerfile
@@ -1,6 +1,6 @@
-# SOCS PTC Agent
+# SOCS DLM Agent
 
-# SOCS Agent container for interacting with PTC over ethernet
+# SOCS Agent container for interacting with DLM over ethernet
 # Use socs base image
 
 FROM socs:latest
--- a/agents/sorenson_dlm/Dockerfile
+++ b/agents/sorenson_dlm/Dockerfile
@@ -1,0 +1,21 @@
+# SOCS PTC Agent
+
+# SOCS Agent container for interacting with PTC over ethernet
+# Use socs base image
+
+FROM socs:latest
+
+# Set the working directory to registry directory
+
+WORKDIR /app/socs/agents/sorenson-dlm/
+
+# Copy this agent into the app/socs/agents directory
+
+COPY . /app/socs/agents/sorenson-dlm/
+
+# Run agent on container startup
+
+ENTRYPOINT ["python3", "-u", "dlm__agent.py"]
+
+CMD ["--site-hub=ws://sisock-crossbar:8001/ws", \
+     "--site-http=http://sisock-crossbar:8001/call"]
--- a/agents/sorenson_dlm/Dockerfile
+++ b/agents/sorenson_dlm/Dockerfile
@@ -15,7 +15,7 @@ COPY . /app/socs/agents/sorenson_dlm/
 
 # Run agent on container startup
 
-ENTRYPOINT ["python3", "-u", "dlm_agent.py"]
+ENTRYPOINT ["dumb-init","python3", "-u", "dlm_agent.py"]
 
-CMD ["--site-hub=ws://sisock-crossbar:8001/ws", \
-     "--site-http=http://sisock-crossbar:8001/call"]
+CMD ["--site-hub=ws://crossbar:8001/ws", \
+     "--site-http=http://crossbar:8001/call"]
--- a/agents/sorenson_dlm/Dockerfile
+++ b/agents/sorenson_dlm/Dockerfile
@@ -15,7 +15,7 @@ COPY . /app/socs/agents/sorenson-dlm/
 
 # Run agent on container startup
 
-ENTRYPOINT ["python3", "-u", "dlm__agent.py"]
+ENTRYPOINT ["python3", "-u", "dlm_agent.py"]
 
 CMD ["--site-hub=ws://sisock-crossbar:8001/ws", \
      "--site-http=http://sisock-crossbar:8001/call"]
--- a/agents/sorenson_dlm/dlm_agent.py
+++ b/agents/sorenson_dlm/dlm_agent.py
@@ -2,20 +2,17 @@
 # Tanay Bhandarkar, Jack Orlowski-Scherer
 import time
 import socket
-import numpy as np
-from contextlib import contextmanager
 from ocs import site_config, ocs_agent
-from ocs.ocs_twisted import TimeoutLock
+from ocs.ocs_twisted import TimeoutLock, Pacemaker
 
 
 class DLM:
-    def __init__(self, ip_address='10.10.10.21', port=9221, timeout=10):
+    def __init__(self, ip_address, port=9221, timeout=10):
         self.ip_address = ip_address
         self.port = port
         self.comm = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.comm.connect((self.ip_address, self.port))
         self.comm.settimeout(timeout)
-        self.volt_prot = None
 
     def send_msg(self, cmd):
         """
@@ -86,12 +83,15 @@ class DLM:
 
 
 class DLMAgent:
-    '''
-    TO DO:
-    set_current_and_voltage (simulatneously set current and voltage,
-    not sure if needed)
+    """ Agent to connect to a Sorenson DLM power supply via ethernet.
+    
+    Args:
+        ip_address: str
+            IP address of the DLM
+        port: int
+            Port number for DLM; default is 9221
 
-    '''
+    """
 
     def __init__(self, agent, ip_address, port, f_sample=2.5):
         self.active = True
@@ -100,41 +100,60 @@ class DLMAgent:
         self.lock = TimeoutLock()
         self.f_sample = f_sample
         self.take_data = False
-        self.over_volt = 0
-        self.dlm = DLM(ip_address, int(port))
+        self.over_volt = 0.
+
+        try:
+            self.dlm = DLM(ip_address, int(port))
+        except socket.timeout as e:
+            self.log.error("DLM power supply has timed out"
+                    + f"during connect with error {e}")
+            return False, "Timeout"
+
         agg_params = {'frame length': 60, }
         self.agent.register_feed('voltages',
                                  record=True,
                                  agg_params=agg_params,
                                  buffer_time=1)
 
-    def start_acq(self, session, params=None):
-        '''
+    def acq(self, session, params=None):
+        """acq("wait=1, test_mode=False)
         Get voltage and current values from the sorenson, publishes them to
         the feed
 
         Args:
-            sampling_frequency: defaults to 2.5Hz
-        '''
+            sampling_frequency: float
+                defaults to 2.5Hz
+        """
         if params is None:
             params = {}
-
+        
+        
         f_sample = params.get('sampling_frequency')
         if f_sample is None:
             f_sample = self.f_sample
-
-        sleep_time = 1. / f_sample - 0.01
+        if f_sample % 1 == 0:
+            pm = Pacemaker(f_sample, True)
+        else:
+            pm = Pacemaker(f_sample)
+        wait_time = 1 / f_sample
         with self.lock.acquire_timeout(timeout=0, job='init') as acquired:
             # Locking mechanism stops code from proceeding if no lock acquired
+            pm.sleep()
             if not acquired:
                 self.log.warn("Could not start init because {} is already running".format(self.lock.job))
                 return False, "Could not acquire lock."
 
             session.set_status('running')
-
+            last_release = time.time()
             self.take_data = True
 
             while self.take_data:
+                # About every second, release and acquire the lock
+                if time.time() - last_release > 1.:
+                    last_release = time.time()
+                    if not self.lock.release_and_acquire(timeout=10):
+                        print(f"Could not re-acquire lock now held by {self.lock.job}.")
+                        return False
                 data = {
                     'timestamp': time.time(),
                     'block_name': 'voltages',
@@ -142,31 +161,34 @@ class DLMAgent:
                 }
                 voltage_reading = self.dlm.read_voltage()
                 current_reading = self.dlm.read_current()
-                print('Voltage: {}'.format(voltage_reading))
-                print('Current: {}'.format(current_reading))
-                data['data']["voltage"] = np.float(voltage_reading)
-                data['data']["current"] = np.float(current_reading)
+                self.log.debug('Voltage: {}'.format(voltage_reading))
+                self.log.debug('Current: {}'.format(current_reading))
+
+                data['data']["voltage"] = float(voltage_reading)
+                data['data']["current"] = float(current_reading)
 
                 self.agent.publish_to_feed('voltages', data)
-                time.sleep(sleep_time)
+                time.sleep(wait_time)
 
             self.agent.feeds['voltages'].flush_buffer()
         return True, 'Acquistion exited cleanly'
-
+    
+    @ocs_agent.param('voltage', default=0., type=float, check=lambda V: 0 <= V <= 300)
     def set_voltage(self, session, params=None):
-        """
+        """set_voltage(voltage=None)
+
         Sets voltage of power supply:
         Args:
             voltage (float): Voltage to set.
 
-        Examples:
+        Examples::
             Example of a client, setting the current to 1V:
 
                 client.set_voltage(voltage = 1.)
 
         """
 
-        with self.lock.acquire_timeout(timeout=0, job='init') as acquired:
+        with self.lock.acquire_timeout(timeout=3, job='init') as acquired:
             if acquired:
                 if self.over_volt == 0:
                     return False, 'Over voltage protection not set'
@@ -180,41 +202,44 @@ class DLMAgent:
         return True, 'Set voltage to {}'.format(params['voltage'])
 
     def set_over_volt(self, session, params=None):
-        """
-        Sets over voltage protection of power supply:
+        """set_over_volt(over_volt=None)
+
+        Sets over voltage protection of power supply.
+
         Args:
             over_volt (int): Over voltage protection to set
 
-        Examples:
+        Examples::
             Example of a client, setting the overvoltage protection to 10V:
 
                 client.set_over_volt(over_volt = 10.)
 
         """
 
-        with self.lock.acquire_timeout(timeout=0, job='init') as acquired:
+        with self.lock.acquire_timeout(timeout=3, job='init') as acquired:
             if acquired:
-                print("ACQUIRED", params)
                 self.dlm.set_overv_prot(params['over_volt'])
                 self.over_volt = float(params['over_volt'])
             else:
                 return False, 'Could not acquire lock'
         return True, 'Set over voltage protection to {}'.format(params['over_volt'])
-
+    
+    @ocs_agent.param('current', default=0., type=float, check=lambda I: 0 <= I <= 2)
     def set_current(self, session, params=None):
-        """
+        """set_current(current=None)
+
         Sets current of power supply:
         Args:
             current (float): Current to set.
 
-        Examples:
+        Examples::
             Example of a client, setting the current to 1A:
 
                 client.set_current(current = 1.)
 
         """
 
-        with self.lock.acquire_timeout(timeout=0, job='init') as acquired:
+        with self.lock.acquire_timeout(timeout=3, job='init') as acquired:
             if acquired:
                 if self.over_volt == 0:
                     return False, 'Over voltage protection not set'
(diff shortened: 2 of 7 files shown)


BrianJKoopman (changes requested):
Thanks for the PR @tanaybhandarkar, a few things before I look at this in more detail. This is based on a commit that's a little over two years old. Can you rebase onto the latest develop branch? Since this only adds new files that should go smoothly:
```
$ git checkout develop
$ git pull origin develop
$ git checkout dlm-agent
$ git rebase develop
$ git push --force-with-lease origin dlm-agent
```

The Agent also needs an [Agent Reference Page](https://ocs.readthedocs.io/en/develop/developer/agents.html#agent-reference-pages) in the documentation. See also the syntax for documenting individual tasks and processes here: https://ocs.readthedocs.io/en/develop/developer/agents.html#task-and-process-documentation

> Regarding tasks and process failing to get locks: If I use the set_voltage task (or any task), I need to stop the acq process otherwise the task does not get the lock. I think there was a discussion about this a while back, but I cannot find the thread, what is the correct way to do this?

You need to add the release and reacquire lock to the acq process, and increase the timeouts in the locks on the individual tasks. This'll allow the acq process to give up the lock each loop, at which point a task can grab it and run it's command. This should only be done if the tasks are short. Details here: https://ocs.readthedocs.io/en/develop/developer/agents.html#release-and-reacquire

agents/sorenson_dlm/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# SOCS DLM Agent
```
BrianJKoopman, line 1: For this docker image to automatically build once we merge this also needs to be added to the main `docker-compose.yaml` file. Add a block like:

```
  ocs-dlm-agent:
    image: "ocs-dlm-agent"
    build: ./agents/sorenson_dlm/
    depends_on:
      - "socs"
```

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+        Examples:
+            Example of a client, setting the overvoltage protection to 10V:
+
+                client.set_over_volt(over_volt = 10.)
+
+        """
+
+        with self.lock.acquire_timeout(timeout=0, job='init') as acquired:
+            if acquired:
+                print("ACQUIRED", params)
```
BrianJKoopman, line 197: This looks like it was for debugging and should be removed now.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+                self.dlm.set_overv_prot(params['over_volt'])
```
BrianJKoopman, line 198: So there are several cases accounted for in `DLM.set_overv_prot` that result in the over voltage protection not actually being set, however they don't pass anything reflecting that in their returned values, and so the Agent can't determine that this is the case and will always return `True, 'Set over voltage protection to {}'.format(params['over_volt'])`. So it's always going to appear successful to the agent (and thus a client).

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+            # Locking mechanism stops code from proceeding if no lock acquired
+            if not acquired:
+                self.log.warn("Could not start init because {} is already running".format(self.lock.job))
+                return False, "Could not acquire lock."
+
+            session.set_status('running')
+
+            self.take_data = True
+
+            while self.take_data:
```
BrianJKoopman, line 137: To add the release and reacquiring of the lock here we'd modify this to:
```
last_release = time.time()
self.take_data = True
while self.take_data:
    # About every second, release and acquire the lock
    if time.time() - last_release > 1.:
        last_release = time.time()
        if not self.lock.release_and_acquire(timeout=10):
            print(f"Could not re-acquire lock now held by {self.lock.job}.")
            return False
```

The timeouts on the locks in the other tasks should also be increased to something like 3 seconds. This allows them to wait for this process to drop the lock so they can run, then the process will resume.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+            sampling_frequency: defaults to 2.5Hz
+        '''
+        if params is None:
+            params = {}
+
+        f_sample = params.get('sampling_frequency')
+        if f_sample is None:
+            f_sample = self.f_sample
+
+        sleep_time = 1./f_sample - 0.01
```
BrianJKoopman, line 126: There's a tool in ocs to help accomplish this called the [Pacemaker](https://ocs.readthedocs.io/en/develop/api.html?highlight=pacemaker#ocs.ocs_twisted.Pacemaker) that you might consider using.

You can see it in use in a couple agents, including the weather monitor: https://github.com/simonsobs/socs/blob/c1d9c1c77d329fc44f106c14d9fb77139781ff95/agents/vantagePro2_agent/vantage_pro2_agent.py#L123

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+            while self.take_data:
+                data = {
+                    'timestamp': time.time(),
+                    'block_name': 'voltages',
+                    'data': {}
+                }
+                voltage_reading = self.dlm.read_voltage()
+                current_reading = self.dlm.read_current()
+                print('Voltage: {}'.format(voltage_reading))
+                print('Current: {}'.format(current_reading))
```
BrianJKoopman, line 146: Useful for debugging, but we don't want to always print this. If you'd like to keep it around, please change the print's to `self.log.debug()` instead.

BrianJKoopman (changes requested):
Here's a more detailed review. There were some relevant changes that it'd be good to get into here, can you run that rebase again?

Let me know if there's anything I can help with.

agents/sorenson_dlm/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+WORKDIR /app/socs/agents/sorenson-dlm/
+
+# Copy this agent into the app/socs/agents directory
+
+COPY . /app/socs/agents/sorenson-dlm/
+
+# Run agent on container startup
+
+ENTRYPOINT ["python3", "-u", "dlm_agent.py"]
```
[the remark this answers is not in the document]
tanaybhandarkar: I added `"dumb-init"` and got an error. Running `sudo -E docker-compose up -d` returns the following error, I think that is related to this conversation. 

```
Recreating ra_ocs-dlm_1               ... error
Creating ra_ocs-smurf-crate-monitor_1 ... done
Starting ra_ocs-pfeiffer_1            ... done

ERROR: for ra_ocs-dlm_1  Cannot start service ocs-dlm: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "dumb-init": executable file not found in $PATH: unknown

ERROR: for ocs-dlm  Cannot start service ocs-dlm: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "dumb-init": executable file not found in $PATH: unknown
ERROR: Encountered errors while bringing up the project.
```

agents/sorenson_dlm/Dockerfile
tanaybhandarkar: Can you clarify where this should be added? I checked other agents' `Dockerfile` and do not see this block in there.

BrianJKoopman: Yup, sorry, over here: https://github.com/simonsobs/socs/blob/develop/docker-compose.yml

agents/sorenson_dlm/Dockerfile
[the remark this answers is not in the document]
BrianJKoopman: My guess is that the socs base image you built the dlm-agent container on is old and doesn't have it installed. Run:
```
$ cd socs/
$ docker build -t socs .
$ docker-compose build
```
then try again.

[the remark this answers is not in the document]
tanaybhandarkar: Oh yes, that was it! Thanks!

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
+# Script to log and control the Sorenson DLM power supply, for heaters
+# Tanay Bhandarkar, Jack Orlowski-Scherer
+import time
+import socket
+import numpy as np
+from contextlib import contextmanager
```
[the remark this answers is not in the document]
tanaybhandarkar: I removed `import contextlib` and replaced the `np.float` on line 146 to just `float` so as to eliminate the `numpy` dependency. But we still need `time` and `socket`.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+                    return False, 'Voltage greater then over voltage protection'
+                else:
+                    self.dlm.send_msg('SOUR:VOLT {}'.format(params['voltage']))
+            else:
+                return False, "Could not acquire lock"
+
+        return True, 'Set voltage to {}'.format(params['voltage'])
+
+    def set_over_volt(self, session, params=None):
+        """
```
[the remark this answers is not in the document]
tanaybhandarkar: Done for most tasks, can I assume `start_acq` and `stop_acq` don't need this?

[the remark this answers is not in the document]
BrianJKoopman: `start_acq` does, `stop_acq` doesn't. A subtle thing with this method signature override is the function names need to match, and we want to present in the documentation the function name registered for use by a client. So actually `start_acq` should be renamed to `acq`. Also, since we don't want `stop_acq` to really show up at all in the docs, `stop_acq` should be renamed to `_stop_acq` as well.

This is done in many other agents at this point, so there are some examples, see for instance: https://github.com/simonsobs/socs/blob/4ce174e0c5dd512db992f0818d3722cdb9fb646e/agents/pfeiffer_tc400/pfeiffer_tc400_agent.py#L263

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,259 @@
...
+                data['data']["voltage"] = np.float(voltage_reading)
+                data['data']["current"] = np.float(current_reading)
+
+                self.agent.publish_to_feed('voltages', data)
+                time.sleep(sleep_time)
+
+            self.agent.feeds['voltages'].flush_buffer()
+        return True, 'Acquistion exited cleanly'
+
+    def set_voltage(self, session, params=None):
```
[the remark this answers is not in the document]
tanaybhandarkar: Implemented for this function and `set_current`, please let me know if I did so correctly

agents/sorenson_dlm/dlm_agent.py
tanaybhandarkar: Added, please review the usage

agents/sorenson_dlm/dlm_agent.py
tanaybhandarkar: I think I implemented this correctly, but please review

agents/sorenson_dlm/dlm_agent.py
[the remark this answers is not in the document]
BrianJKoopman: There's some small formatting issues here, fixing in a commit that I'll push shortly.

agents/sorenson_dlm/dlm_agent.py
BrianJKoopman: Adding the pacemaker looks alright. I'd probably put the `pm.sleep()` down where you have `time.sleep(wait_time)`, i.e. sleep after you've taken a measurement, rather than before. And then you can remove your `time.sleep(wait_time)`, that's what the pacemaker is doing. As is you're waiting twice as long as you need to for the desired sampling frequency.

agents/sorenson_dlm/dlm_agent.py
BrianJKoopman: Yes, returning False when `set_overv_prot` fails, and reacting accordingly in the agent sounds good to me.

BrianJKoopman (changes requested):
A few other comments on the changes.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,305 @@
...
+        self.f_sample = f_sample
+        self.take_data = False
+        self.over_volt = 0.
+
+        try:
+            self.dlm = DLM(ip_address, int(port))
+        except socket.timeout as e:
+            self.log.error("DLM power supply has timed out"
+                           + f"during connect with error {e}")
+            return False, "Timeout"
```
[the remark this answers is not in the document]
tanaybhandarkar: I'm a bit unsure about how to correctly implement the `init_dlm` task. I just pushed some code where I essentially copy/pasted the LS372 code and trimmed what I did not need for the DLM. Two questions I have:

1. How do I treat the code you highlighted above? Just cut it out totally?
2. What do I need for `init_params`?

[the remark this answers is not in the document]
BrianJKoopman: 1. You could just remove the `return` on line 111.
2. `init_params` (like used in the 372 Agent) is a dictionary that contains parameters you want to pass to the task at startup. So that depends on what parameters you put on your `init_dlm` task.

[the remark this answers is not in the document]
tanaybhandarkar: Ok, I just pushed code to implement these changes- hopefully I did this correctly. I might need to trim some arguments in the `init_params` dictionary, let me know if I should.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -255,6 +255,49 @@ def set_current(self, session, params=None):
...
+        Parameters:
+            auto_acquire (bool, optional): Default is False. Starts data
+                acquisition after initialization if True.
+            acq_params (dict, optional): Params to pass to acq process if
+                auto_acquire is True.
+            force (bool, optional): Force initialization, even if already
+                initialized. Defaults to False.
+        """
+        if params is None:
+            params = {}
```
BrianJKoopman, line 273: This is no longer needed, as you can assume the `params` dict exists since you're using the param decorator.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -255,6 +255,49 @@ def set_current(self, session, params=None):
...
+
+        if self.initialized and not params.get('force', False):
```
BrianJKoopman, line 275: Using `params.get()` is no longer needed either. The default is set in the params decorator, this is just redundant. And you can safely assume `params['force']` exists and just use that directly.

So this line should read: `if self.initialized and not params['force']:`

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -255,6 +255,49 @@ def set_current(self, session, params=None):
...
+                as acquired2:
+            if not acquired1:
+                self.log.warn(f"Could not start init because "
+                              f"{self._lock.job} is already running")
+                return False, "Could not acquire lock"
+            if not acquired2:
+                self.log.warn(f"Could not start init because "
+                              f"{self._acq_proc_lock.job} is already running")
+                return False, "Could not acquire lock"
+        session.set_status('running')
```
BrianJKoopman, line 290: A couple things here. First, this doesn't actually initialize the DLM object, that's still being done in `__init__()`. That should be moved here.

This looks like it's based off of the Lakeshore 372, which is perhaps not the best example, it's a bit old. Check out the example here: https://ocs.readthedocs.io/en/develop/developer/agent_references/timeoutlock.html#release-and-reacquire

That example doesn't have this strange double lock context.

Also, I'm noticing that you have two locks, `self.lock` and `self._lock`. Get rid of one of these and use a single lock everywhere. Otherwise you might "have" one of the locks, then check the other one, meaning you're potentially communicating simultaneously with the DLM, which is what the lock is trying to prevent in the first place.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -319,14 +318,24 @@ def make_parser(parser=None):
                         + "converter ip address")
     pgroup.add_argument('--port', type=int, help="Serial-to-ethernet "
                         + "converter port")
-
+    pgroup.add_argument('--mode', default='acq',
+                        choices=['idel', 'acq', 'acq_reg'])
```
BrianJKoopman, line 322: What's `acq_reg`? It's not used anywhere.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -319,14 +318,24 @@ def make_parser(parser=None):
...
                                   parser=parser)
+    # Automatically acquire data if request (default)
+    init_params = False
+    if args.mode == 'init':
+        init_params = {'auto_acquire': False,
+                       'acq_params': {},
+                       'configfile': args.configfile}
+    elif args.mode == 'acq':
+        init_params = {'auto_acquire': True,
+                       'acq_params': {}}
```
BrianJKoopman, line 338: This looks good, but then this `init_params` dict needs to be passed to the `init_dlm` task when it is registered on line 364. It should be passed to the `startup` argument.

BrianJKoopman (changes requested):
Thanks for the updates @tanaybhandarkar, I've made a few comments on the recent changes.

Can I also get you to look at #318 before anymore work is done? That adds tests for this Agent, and fixes a few things. I'd like to get this merged to your branch, you finish working on updates, then I'll take another look. Sound good?

agents/sorenson_dlm/dlm_agent.py
tanaybhandarkar: Several questions:

> First, this doesn't actually initialize the DLM object, that's still being done in __init__(). That should be moved here.

What should be moved here? The guts of the `__init__()` function? If this is the case, where do I supply arguments like `ip_address` and `f_sample`, in `init_dlm()`? (I.E. `init_dlm(self, session, params=None, ip_address, port, f_sample=2.5)`)

> This looks like it's based off of the Lakeshore 372, which is perhaps not the best example, it's a bit old. Check out the example here: https://ocs.readthedocs.io/en/develop/developer/agent_references/timeoutlock.html#release-and-reacquire

The double lock context being the `acquired1` and `acquired2` locks? Should i just be inserting the code block from the example that start at `with self.lock.acquire_timeout(timeout=0, job='acq') as acquired:`

BrianJKoopman: > What should be moved here? The guts of the `__init__()` function? If this is the case, where do I supply arguments like `ip_address` and `f_sample`, in `init_dlm()`? (I.E. `init_dlm(self, session, params=None, ip_address, port, f_sample=2.5)`)

The DLM objection initialization is this block of the `__init__()` function: https://github.com/simonsobs/socs/blob/2b56f4bdbe2421680f6bdfec529412e1a64635a4/agents/sorenson_dlm/dlm_agent.py#L106-L111

> The double lock context being the `acquired1` and `acquired2` locks? Should i just be inserting the code block from the example that start at `with self.lock.acquire_timeout(timeout=0, job='acq') as acquired:`

This task is a "short action", so you need this block:
```
        with self.acquire_timeout(timeout=3, job='short_action') as acquired:
            if not acquired:
                print(f"Lock could not be acquired because it is held by {self.lock.job}")
                return False

            # Your code that performs the initialization here
```

tanaybhandarkar: So where do I pass the  `ip_address` and `port` arguments for the `init_dlm` function? Should that go in `init_params`?

BrianJKoopman: Keep them as arguments to the class, and store them as class attributes. Then use them within `init_dlm` via `self.ip_address`. The Lakeshore 372 Agent does this for example within its `init` task: https://github.com/simonsobs/socs/blob/53758a1f5866fcb03b0525adadd01537fb5eaf2c/agents/lakeshore372/LS372_agent.py#L195

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,349 @@
...
+    if parser is None:
+        parser = argparse.ArgumentParser()
+
+    pgroup = parser.add_argument_group('Agent Options')
+    pgroup.add_argument('--ip-address', type=str, help="Serial-to-ethernet "
+                        + "converter ip address")
+    pgroup.add_argument('--port', type=int, help="Serial-to-ethernet "
+                        + "converter port")
+    pgroup.add_argument('--mode', default='acq',
+                        choices=['idel', 'acq'])
```
BrianJKoopman, line 323: `idel` -> `idle`

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,349 @@
...
+if __name__ == '__main__':
+    parser = make_parser()
+    args = site_config.parse_args(agent_class='DLMAgent',
+                                  parser=parser)
+    # Automatically acquire data if request (default)
+    init_params = False
+    if args.mode == 'init':
+        init_params = {'auto_acquire': False,
+                       'acq_params': {},
+                       'configfile': args.configfile}
```
BrianJKoopman, line 336: `init` isn't a valid mode, per the modes listed under `choices` in the `--mode` argument of the parser. You can add that as a mode, but need to make sure the parameters here in this dictionary, which will be passed to `init_dlm` are valid as well.

agents/sorenson_dlm/dlm_agent.py
```diff
@@ -0,0 +1,349 @@
...
+        self.ip_address = ip_address
+        self.port = port
+        self.log = agent.log
+        self.lock = TimeoutLock()
+        self.f_sample = f_sample
+        self.take_data = False
+        self.over_volt = 0.
+        self.initialized = False
+        self._acq_proc_lock = TimeoutLock()
+        self._lock = TimeoutLock()
```
BrianJKoopman, line 111: There are three different `TimeoutLock`'s here, you only need one. Remove these and use `self.lock` everywhere.

BrianJKoopman (changes requested):
A few things still. Also, the tests aren't passing. I haven't dug too deep into those (one issue at least looks like it relates to the locks), but we need to get them working before merging.

Here's an example for how to run the tests:
```
cd socs/tests/
python3 -m pytest integration/test_dlm_agent_integration.py -s -k 'test_dlm_set_over_volt'
```
This runs a specific test, and prints output to the terminal. You can run all the tests for the DLM Agent by dropping the `-k 'test_dlm_set_over_volt`, but it's often helpful to run one test at a time when trying to fix things, just for speed of iterating.

agents/sorenson_dlm/dlm_agent.py
tanaybhandarkar: Removing this block, I don't think it's needed

BrianJKoopman:
Closing this as not implementing after a discussion with @tanaybhandarkar. It sounds like the hardware is at site, but unused. This can be revisited in the future if needed.

PR #306, opened 2022-06-10
Added iBootbar agent with MIB and added set function to snmp.py

### LINKED ISSUE #63: ibootbar Agent
Create ibootbar Agent for control of power distribution.

davidvng:
## Description

Added a set function to snmp.py to handle any set commands used in agent scripts.
Added new agent directory with Dockerfile, agent python script, and MIB (in sub-directory mibs).
The iBootbar agent contains 1 process and 4 tasks:
- acq (process) - gets outlet status and sends to session.data and feed
- get_names (task) - returns list of outlet names and updates them in session.data and feed
- set_outlet (task) - sets a particular outlet to on/off
- cycle_outlet (task) - cycles a particular outlet for a given amount of time
- reboot (task) - reboots the entire system

## Motivation and Context

Initial iBootbar agent PR.

Closes #63.

## How Has This Been Tested?

Tested with iBootbar equipment in the lab. Ran each individual task and verified the changes done on the system. Tested the main process by verifying data is seen on OCS web and Grafana.

## Types of changes

## Checklist:


agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+                # Convert given state parameter to integer
+                if params['state'] == 'on':
+                    state = 1
+                else:
+                    state = 0
+
+                # Send set command to given outlet
+                outlet = [('IBOOTPDU-MIB', 'outletControl', params['outlet'])]
+                setcmd = yield self.ibootbar.snmp.set(outlet, self.ibootbar.version, state)
+                print(setcmd)
```
BrianJKoopman, line 421: A debug print? Or do we want to log this info. Either way, use `self.log.info()` or `self.log.debug()` if it's sticking around.

(Similarly in `cycle_outlet` after running an `snmp.set()`.)

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+
+                return True, 'Set outlet {} to {}'.\
+                    format(params['outlet'], params['state'])
+            else:
+                return False, "Could not acquire lock"
+
+    @ocs_agent.param('outlet', choices=[0, 1, 2, 3, 4, 5, 6, 7])
+    @ocs_agent.param('cycle_time', default=10, type=int)
+    @inlineCallbacks
+    def cycle_outlet(self, session, params=None):
```
BrianJKoopman, line 435: One problem here is that since acq is only querying power states every minute, we likely won't see a cycle happen in the output in `session.data`.

The simplest solution would probably be to just query much more often, like once a second. Otherwise you could force more rapid updates during the cycle somehow.

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+                print(setcmd2)
+
+                return True, 'Cycling outlet {} for {} seconds'.\
+                    format(params['outlet'], params['cycle_time'])
+            else:
+                return False, "Could not acquire lock"
+
+    @ocs_agent.param('_')
+    @inlineCallbacks
+    def reboot(self, session, params=None):
```
BrianJKoopman, line 458: What happens to the SNMP communication going on in `acq` during a reboot? Does it lose communication, or gracefully reconnect when the ibootbar is back online? Related, what state does the acq process show for all outlets during that time? (It should show they're all off.)

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+
+# For logging
+txaio.use_twisted()
+
+on_rtd = os.environ.get('READTHEDOCS') == 'True'
+if not on_rtd:
+    from ocs import ocs_agent, site_config
+
+
+class ibootbarSNMP:
```
BrianJKoopman, line 25: I'm glad the `MeinbergSNMP` class was useful here (though the whole timing structure I built for that to query in interval groups is perhaps not really needed here.) However, there's enough code re-use here that it'd be worth pulling a lot of this out to a class that `ibootbarSNMP` and `MeinbergSNMP` can use as a base class, for instance.

BrianJKoopman (changes requested):
Thanks for the PR @davidvng! It's looking good, I have some suggestions for the structure, some questions, and documentation requests inline below.

One thing we're still missing is an [Agent Reference Page](https://ocs.readthedocs.io/en/develop/developer/agents.html#agent-reference-pages). That'll go in `socs/docs/agents/` and should be added to the index `socs/docs/index.rst` under the "Agent Reference" section. You can then try building the docs locally, which I usually do to check formatting.

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+                result = yield self.ibootbar.snmp.get(get_list, self.ibootbar.version)
+                for i in range(len(result)):
+                    self.ibootbar.names.append(result[i][1].prettyPrint())
+
+                return True, "The outlet names are: {}".\
+                    format(self.ibootbar.names)
+            else:
+                return False, "Could not acquire lock"
+
+    @ocs_agent.param('outlet', choices=[0, 1, 2, 3, 4, 5, 6, 7])
```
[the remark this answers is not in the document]
davidvng: The outlets on the actual device are labeled 1-8. The MIB table structure is indexed 0-7, which also shows in session.data. Should I re-index to allow users/client to choose 1-8 instead?

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+        return True, "Stopping Recording"
+
+    @ocs_agent.param('_')
+    @inlineCallbacks
+    def get_names(self, session, params=None):
+        """
+        **Task** - Gets list of outlet names and updates the name list for the next get request
+        """
+        with self.lock.acquire_timeout(3, job='get_names') as acquired:
+            if acquired:
```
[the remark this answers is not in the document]
davidvng: Just being cautious, but I have tested the other commands (that use SET) without the lock and they cause a failure in the task even though it physically goes through with the set command (and also causes the OCS connection to cut for a second).

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,520 @@
...
+                # Clear the default names defined in self.ibootbar
+                self.ibootbar.names = []
+
+                # Send get command and append results to self.ibootbar.names
+                result = yield self.ibootbar.snmp.get(get_list, self.ibootbar.version)
+                for i in range(len(result)):
+                    self.ibootbar.names.append(result[i][1].prettyPrint())
+
+                return True, "The outlet names are: {}".\
+                    format(self.ibootbar.names)
```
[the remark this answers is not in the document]
davidvng: This task sends the names to self.names, which is used in ibootbarSNMP functions and they are then sent to session.data. They are given to their respective session.data objects as attributes. Querying the name OIDs in the acq process would mean they appear as separate session.data objects from the status ones, but if that actually makes it easier for the panel development then I can do that.

agents/ibootbar/ibootbar.py
```diff
@@ -0,0 +1,411 @@
...
+                {"fields":
+                    {ibootbar:
+                        {outletStatus_0: {"status": 1, "name": Outlet-1, "description": "on"},
+                         outletStatus_1: {"status": 0, "name": Outlet-2, "description": "off"},
+                         ...
+                         ibootbar_connection: {'last_attempt': 1656085022.680916, 'connected': True},
+                         timestamp: {1656085022.680916}
+                        }
+                    }
+                }
```
BrianJKoopman, line 219: Based on `update_cache()` I think the structure here might be flatter than this. It looks like the "field" and "ibootbar" levels are not there, i.e. it should read:
```
>>> response.session['data']
{outletStatus_0: {"status": 1, "name": Outlet-1, "description": "on"},
 outletStatus_1: {"status": 0, "name": Outlet-2, "description": "off"},
 ...
 ibootbar_connection: {'last_attempt': 1656085022.680916, 'connected': True},
 timestamp: {1656085022.680916}}
```

You can confirm this structure by running the Agent and using a client to run:
```
>>> response = client.acq.status()
>>> response.session['data']
```

agents/ibootbar/ibootbar.py
```diff
@@ -341,29 +200,67 @@ def __init__(self, agent, address, port=161, version=2):
...
+            for item in name_result:
+                names.append(item[1].prettyPrint())
+            
+            # Do not publish if ibootbar connection has dropped
+            try:
+                message = _build_message(get_result, names, read_time)
+                self.log.debug("{msg}", msg=message)
+                session.app.publish_to_feed('ibootbar', message)
+            except ConnectionError as e:
+                self.log.error(f'{e}')
```
BrianJKoopman, line 259: I think I'm missing something here. Where does the ConnectionError come up? I would have thought in the `publish_to_feed` call, but that would be if the crossbar connection has dropped, not if the ibootbar connection dropped, as indicated by the comment. If the ibootbar disconnects, what comes back from the above `snmp.get`'s?

`session.data` is then still updated, even if we hit this, which seems inconsistent.

agents/ibootbar/ibootbar.py
```diff
@@ -461,15 +350,17 @@ def reboot(self, session, params=None):
...
-            else:
+            if not acquired:
                 return False, "Could not acquire lock"
+            # Send SNMP SET command to set rebootSystem to True
+            reboot = [('IBOOTPDU-MIB', 'rebootSystem', 0)]
+            setcmd = yield self.snmp.set(reboot, self.version, 1)
+            self.log.info('{}'.format(setcmd))
+
+        # Force SNMP GET status commands
+        self.lastGet = self.lastGet - 60
```
BrianJKoopman, line 362: Related to my question in the first review: https://github.com/simonsobs/socs/pull/306#discussion_r905217711

Does this result in all outlets turning "off" in the acq process during the reboot?

BrianJKoopman (commented):
This is looking good! I have a couple questions still, and some clarification is needed on the session.data format.

agents/ibootbar/ibootbar.py
davidvng: Yep, you're right, moving `update_cache()` inside the `try` block. If the ibootbar disconnects, `snmp.get` returns `None` which then raises a `ConnectionError` in `update_cache()`

BrianJKoopman: Ah, yeah, I see where this is coming from when looking at the M1000 agent. The first thing it does there is `update_cache()`, which would hit `ConnectionError` raise like you said: https://github.com/simonsobs/socs/blob/d12fea99596ac553628d551dc005e504dbcb3ee6/agents/meinberg_m1000/meinberg_m1000_agent.py#L298-L311

agents/ibootbar/ibootbar.py
BrianJKoopman: Renaming the task sounds good. Users might interpret "reboot" as "cycle all outlets". I'm not really sure we need a "cycle all outlets" task at the moment, so let's just stick to renaming this. If we find we need that we can add it later.

BrianJKoopman (approved):

PR #320, opened 2022-06-29
Adds MultiSlotController

jlashner:
Client object useful for commanding multiple pysmurf controllers

## Description
Class to contain OCS client objects for all pysmurf-controller agnets on a network, with methods to easily dispatch start/stop/status commands.

## Motivation and Context
This is required for operating multiple pysmurf controllers at once without sorunlib (for instance during this LATR cooldown).

## How Has This Been Tested?
Tested at USCD with the full smurf-crate in loopback mode.

## Types of changes

## Checklist:

### DIFF
--- a/socs/smurf/MultiSlotController.py
+++ b/socs/smurf/MultiSlotController.py
@@ -1,0 +1,116 @@
+from ocs.ocs_client import OCSClient
+
+class MultiSlotController:
+    """
+    Simple client controller useful for managing all pysmurf-controller
+    agents that exist on a network. This can dispatch operations and check
+    operation status for any or all pysmurf-controllers on a network.
+
+    Args
+    -----
+    client_args: list
+        List of command line arguments to pass onto the OCSClient objects.
+
+    Attributes
+    ---------------
+    smurfs : dict
+        Dictionary containing OCSClient objects for each pysmurf-controller
+        on the network. This assumes pysmurf controller agents have
+        instance-id's of the form ``pysmurf-controller-<id>`` where id
+        can be the uxm-id or something like ``crate1slot2``. The keys of this
+        dict will be ``<id>``. This is populated from the registry agent on
+        init or using the funciton ``get_active_smurfs``.
+    """
+    def __init__(self, client_args=[]):
+        self.client_args=client_args
+        self.reg = OCSClient('registry', args=client_args)
+        self.smurfs = {}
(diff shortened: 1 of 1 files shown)


socs/smurf/MultiSlotController.py
```diff
@@ -0,0 +1,118 @@
...
+    Attributes
+    ---------------
+    smurfs : dict
+        Dictionary containing OCSClient objects for each pysmurf-controller
+        on the network. This assumes pysmurf controller agents have
+        instance-id's of the form ``pysmurf-controller-<id>`` where id
+        can be the uxm-id or something like ``crate1slot2``. The keys of this
+        dict will be ``<id>``. This is populated from the registry agent on
+        init or using the funciton ``get_active_smurfs``.
+    """
```
BrianJKoopman, line 24: Since this likely will see wider usage at test institutions, can you add an Examples section to the end here with some simple examples.

socs/smurf/MultiSlotController.py
```diff
@@ -0,0 +1,118 @@
...
+            List of pysmurf-controller id's. This defaults to running on all
+            controllers.
+        **kwargs
+            Any additional keyword arguments are passed onto the start
+            operation as params
+        """
+        if ids is None:
+            ids = list(self.smurfs.keys())
+        elif isinstance(ids, str):
+            ids = [ids]
```
BrianJKoopman, line 69: Is this meant to catch times a user submits just a single id?

I could imagine something like `.start('uxm_setup', ids='pysmurf-controller-crate1slot1, pysmurf-controller-crate1slot2')` being submitted, which won't work here.

BrianJKoopman (commented):
Thanks for the PR, Jack, this looks useful.

Couple of inline comments below, and one other question -- do we need a `wait` method?

socs/smurf/MultiSlotController.py
jlashner: Eh it's less about catching mistakes and more that if I'm passing a single ID I find it more convenient to pass directly instead of in a list, though I admit it doesn't really matter. I'm just used to using `np.atleast_1d` in a lot of my sodetlib functions which handles this kind of input. I can remove it if you want though.

jlashner:
Thanks for the quick review! I added a wait function that behaves like you'd expect and also an examples section.

PR #321, opened 2022-06-29
Stop and clear overrides scan in ACU Agent

laurensaunders:
## Description
stop_and_clear successfully stopped motion with the ACU but could not stop _run_specified_scan from uploading points. We add a check on the number of points to determine whether stop_and_clear has cleared all uploaded points and, if so, end the Task.

## Motivation and Context
Implemented to allow the end of _run_specified_scan if a scan needs to be aborted

## Types of changes

## Checklist:

### DIFF
--- a/agents/acu/acu_agent.py
+++ b/agents/acu/acu_agent.py
@@ -910,7 +910,7 @@ class ACUAgent:
             self.log.info('Uploaded a group')
         self.log.info('No more lines to upload')
         free_positions = self.data['status']['summary']['Free_upload_positions']
-        while free_positions < FULL_STACK:
+        while free_positions < FULL_STACK - 1:
             yield dsleep(0.1)
             modes = (self.data['status']['summary']['Azimuth_mode'],
                      self.data['status']['summary']['Elevation_mode'])
@@ -924,21 +924,18 @@ class ACUAgent:
         self.log.info('No more points in the queue')
         current_az = self.data['broadcast']['Corrected_Azimuth']
         current_el = self.data['broadcast']['Corrected_Elevation']
-        az_attempt_number = 0
-        while az_attempt_number < 10:
-            while round(current_az - end_az, 1) != 0.:
-                self.log.info('Waiting to settle at azimuth position')
-                yield dsleep(0.1)
-                az_attempt_number += 1
-                current_az = self.data['broadcast']['Corrected_Azimuth']
+        while round(current_az - end_az, 1) != 0.:
+            print('still rounding az')
+            self.log.info('Waiting to settle at azimuth position')
+            yield dsleep(0.1)
+            current_az = self.data['broadcast']['Corrected_Azimuth']
         if not azonly:
-            el_attempt_number = 0
-            while el_attempt_number < 10:
-                while round(current_el - end_el, 1) != 0.:
-                    self.log.info('Waiting to settle at elevation position')
-                    yield dsleep(0.1)
-                    el_attempt_number += 1
-                    current_el = self.data['broadcast']['Corrected_Elevation']
+            while round(current_el - end_el, 1) != 0.:
+                print('still rounding el')
+                self.log.info('Waiting to settle at elevation position')
+                yield dsleep(0.1)
+                current_el = self.data['broadcast']['Corrected_Elevation']
+        print('rounding completed')
         yield dsleep(self.sleeptime)
         yield self.acu_control.stop()
         self.data['uploads']['Start_Azimuth'] = 0.0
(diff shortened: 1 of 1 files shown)


agents/acu/acu_agent.py
```diff
@@ -892,10 +892,14 @@ def _run_specified_scan(self, session, times, azs, els, vas, ves, azflags, elfla
...
             text = ''.join(upload_lines)
+            yield dsleep(0.05)
             free_positions = self.data['status']['summary']['Free_upload_positions']
             while free_positions < 9899:
                 free_positions = self.data['status']['summary']['Free_upload_positions']
-                yield dsleep(0.1)
+                yield dsleep(0.05)
+                if free_positions > 9997:
+                    self.log.warn('Scan aborted!')
+                    return False
```
mhasself, line 902: In the next loop down, the trap is on the axis mode (modes != ProgramTrack), though it is interpreted as a fault.  It would perhaps be better to use that mode check here, too.  If you are trying to discriminate between an unexpected fault (where free positions is not reset) and an abort (where free_positions -> 0), then that's great but you should probably do both things in both places.

I think the free_upload_positions check is not obviously reliable, because of the race condition between "stop_and_clear", which should set it to 10000, and this loop which is trying to upload points (which would usually put it below 9899).  It's not obvious that you couldn't get the sequence:
- _run_specified_scan: uploads some points ...
- _run_specified_scan: while free_pos < 9899, watch for abort condition...
- _run_specified_scan: prepare for next iteration
- _abort! stop_and_clear sets free_pos = 10000 and mode='stop'
- _run_specified_scan: uploads 120 points, free_pos -> 9700 or something
- _run_specified_scan: while free_pos < 9899, watch for abort condition <--- never exits, because points aren't being consumed.

Maybe that can't actually happen?  But it's not obvious, and we prefer obvious things.  And when something is true but not obvious it requires some comments!

So I think mode will be better.  Mode probably has its own weirdnesses, but at least we're not actively setting it in these loops.

mhasself: Also:

There are two loops, below, with dsleeps in them, that could in principle hang if the telescope suddenly stops moving.  Those need a similar check / checks.  (A local function to do this repeated task is probably a good idea.)

While you're in there I am tempted to request that all these stack size integers (10000, 9999, 9997 ... why 9997? 9899) get turned into constants / variables, or at least be explained... but I recognize they are a pre-existing condition.

agents/acu/acu_agent.py
```diff
@@ -918,15 +924,21 @@ def _run_specified_scan(self, session, times, azs, els, vas, ves, azflags, elfla
...
+                az_attempt_number += 1
+                current_az = self.data['broadcast']['Corrected_Azimuth']
+        if not azonly:
+            el_attempt_number = 0
+            while el_attempt_number < 10:
+                while round(current_el - end_el, 1) != 0.:
+                    self.log.info('Waiting to settle at elevation position')
+                    yield dsleep(0.1)
+                    el_attempt_number += 1
+                    current_el = self.data['broadcast']['Corrected_Elevation']
```
mhasself, line 941: What is the idea behind these "attempt" loops?  They look like they might have some pathological behavior (specifically they will exit automatically after 1s even if the telescope is still not in position; but if the telescope _does_ reach target before 1s has elapsed then they will block forever).

agents/acu/acu_agent.py
```diff
@@ -924,21 +924,18 @@ def _run_specified_scan(self, session, times, azs, els, vas, ves, azflags, elfla
...
-                    self.log.info('Waiting to settle at elevation position')
-                    yield dsleep(0.1)
-                    el_attempt_number += 1
-                    current_el = self.data['broadcast']['Corrected_Elevation']
+            while round(current_el - end_el, 1) != 0.:
+                print('still rounding el')
+                self.log.info('Waiting to settle at elevation position')
+                yield dsleep(0.1)
+                current_el = self.data['broadcast']['Corrected_Elevation']
+        print('rounding completed')
```
mhasself, line 938: Ok, so now these loops will run indefinitely, unless the "current" positions reach the "end" positions.

I think there is still a potential for this to hang; imagine the situation where someone sees the telescope reach the end of the program, and then while it is still settling (i.e. it is stuck in these loops), they run stop-and-clear.  Then the telescope is stopped, near-to-but-not-quite-at the final positions.

So I think you either need to check for a mode switch (did someone send it to Stop?) or to simply bail out after some max total wait time (which I think was what you were thinking in the previous version).

You could do something like:
```
    give_up_time = time.time() + 5
    while round(current_az - end_az, 1) != 0. and time.time() < give_up_time:
      ...
```

I see that, below,  you do run .stop() before exiting...

BrianJKoopman:
These changes will be addressed in an upcoming PR from the [acu_aug_lat_updates](https://github.com/simonsobs/socs/tree/acu_aug_lat_updates) branch.

PR #323, opened 2022-06-30
New parameter for scan start position in generator scans

laurensaunders:
Include new string parameter az_start to indicate the beginning point
of a generator scan. Defaults to starting scans from the midpoint between
the two az endpoint inputs.

## Types of changes

## Checklist:

### DIFF
--- a/agents/acu/acu_agent.py
+++ b/agents/acu/acu_agent.py
@@ -948,7 +948,9 @@ class ACUAgent:
     def generate_scan(self, session, params):
         """generate_scan(az_endpoint1=None, az_endpoint2=None, az_speed=None, \
                          acc=None, el_endpoint1=None, el_endpoint2=None, \
-                         el_speed=None)
+                         el_speed=None, num_batches=None, start_time=None, \
+                         wait_to_start=None, step_time=None, batch_size=None, \
+                         az_start=None, ptstack_fmt=None)
 
         **Process** - Scan generator, currently only works for
         constant-velocity az scans with fixed elevation.
@@ -963,7 +965,26 @@ class ACUAgent:
                 currently both el endpoints should be equal
             el_speed (float): speed of motion for a scan with changing
                 elevation. For dev, currently set to 0.0
-
+            num_batches (int or None): number of times to traverse a direction
+                of the scan. Default value is None, interpreted as infinite
+            start_time (float or None): a ctime at which to start the scan.
+                Default is None, interpreted as now
+            wait_to_start (float): number of seconds to wait before starting a
+                scan. Default is 3 seconds
+            step_time (float): time between points on the constant-velocity
+                parts of the motion. Default is 0.1 s. Minimum 0,05 s
+            batch_size (int): number of values to produce in each iteration.
+            Default is 500. Batch size is reset to the length of one leg of the
+            motion if num_batches is not None.
+        az_start (str): part of the scan to start at. Options are:
+            'az_endpoint1', 'az_endpoint2', 'mid_inc' (start in the middle of
+            the scan and start with increasing azimuth), 'mid_dec' (start in
+            the middle of the scan and start with decreasing azimuth).
+        ptstack_fmt (bool): determine whether values are produced with the
+            necessary format to upload to the ACU. If False, this function will
+            produce lists of time, azimuth, elevation, azimuth velocity,
+            elevation velocity, azimuth flags, and elevation flags. Default is
+            True.
         """
         ok, msg = self._try_set_job('control')
         if not ok:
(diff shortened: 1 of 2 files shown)


mhasself (approved):
Looks good, thanks!

BrianJKoopman:
I've updated sorunlib to start at the endpoints, like the default behavior before this patch. I think we should expose this argument in the Agent's parameters, since right now the Agent just uses the defaults in `sh.generate_constant_velocity_scan()`:
https://github.com/simonsobs/socs/blob/f856eaf4de6417f1827045c37c1af0166b05579d/agents/acu/acu_agent.py#L981-L982

agents/acu/acu_agent.py
```diff
@@ -963,7 +965,22 @@ def generate_scan(self, session, params):
...
+                scan. Default is 3 seconds
+            step_time (float): time between points on the constant-velocity
+                parts of the motion. Default is 0.1 s. Minimum 0,05 s
+            batch_size (int): number of values to produce in each iteration.
+            Default is 500. Batch size is reset to the length of one leg of the
+            motion if num_batches is not None.
+        az_start (str): part of the scan to start at. Options are:
+            'az_endpoint1', 'az_endpoint2', 'mid_inc' (start in the middle of
+            the scan and start with increasing azimuth), 'mid_dec' (start in
+            the middle of the scan and start with decreasing azimuth).
```
mhasself, line 983: Indentation is off here.

agents/acu/acu_agent.py
```diff
@@ -963,7 +965,22 @@ def generate_scan(self, session, params):
...
-
+            num_batches (int or None): sets the number of batches for the
+                generator to create. Default value is None (interpreted as infinite
+                batches).
+            start_time (float or None): a ctime at which to start the scan.
+                Default is None, interpreted as now
+            wait_to_start (float): number of seconds to wait before starting a
+                scan. Default is 3 seconds
+            step_time (float): time between points on the constant-velocity
+                parts of the motion. Default is 0.1 s. Minimum 0,05 s
```
mhasself, line 976: "0.05 s."

agents/acu/acu_agent.py
```diff
@@ -974,12 +991,22 @@ def generate_scan(self, session, params):
...
+        az_start = params.get('az_start', 'mid_inc')
 
         yield self.acu_control.stop()
         g = sh.generate_constant_velocity_scan(az_endpoint1, az_endpoint2,
-                                               az_speed, acc, el_endpoint1, el_endpoint2, el_speed)
+                                               az_speed, acc, el_endpoint1,
+                                               el_endpoint2, el_speed,
+                                               num_batches, start_time,
+                                               wait_to_start, step_time,
+                                               batch_size, az_start)
```
mhasself, line 1009: It is best practice, in python, to pass a function's keyword arguments as keyword arguments rather than by position.  I.e. `num_batches=num_batches`, etc.  This may seem verbose, but it saves trouble later if the interface needs to be altered.

That ^^^, above, is my request for change.

This stuff below vvv is an alternative solution for you to consider.

In cases where you're just passing through a bunch of things from another function, you can bundle up the keyword args in a dict, and pass them in using **.  For example:
```
  # Populate params dict.
  scan_args = {
    'num_batches': params.get('num_batches', None),
    'start_time': params.get('start_time', None),
    ...
  }

  ...
  g = sh.generate_constant_velocity_scan(..., **scan_params)
```

Note that when you do it this way, you can fairly easy recreate the case where you _don't specify_ a parameter at all ... if you want the default values to be chosen by sh.generate_constant_velocity_scan, you can do something like:
```
    # Put only the non-None args into scan_args
    scan_params = {k: params.get(k) for k in ['num_batches', 'start_time', ...]
                   if params.get(k) is not None}
    ...

    g = sh.generate_constant_velocity_scan(..., **scan_params)
```

agents/acu/acu_agent.py
```diff
@@ -948,7 +948,9 @@ def _run_specified_scan(self, session, times, azs, els, vas, ves, azflags, elfla
     def generate_scan(self, session, params):
         """generate_scan(az_endpoint1=None, az_endpoint2=None, az_speed=None, \
                          acc=None, el_endpoint1=None, el_endpoint2=None, \
-                         el_speed=None)
+                         el_speed=None, num_batches=None, start_time=None, \
+                         wait_to_start=None, step_time=None, batch_size=None, \
+                         az_start=None, ptstack_fmt=None)
```
mhasself, line 953: Remove ptstack_fmt.

agents/acu/acu_agent.py
```diff
@@ -948,7 +948,9 @@ def _run_specified_scan(self, session, times, azs, els, vas, ves, azflags, elfla
     def generate_scan(self, session, params):
```
mhasself, line 948: Is there a reason to not use the @ param decorator on this process?

mhasself (changes requested):
Some minor requests ... and a more serious question about the param decorator.

PR #337, opened 2022-08-03
Suprsync update copy attempts

jlashner:
Fixes a bug in the suprsync agent where failed_copy_attempts wasn't being updated.

## Description
Changes 'copy_files' function so that if md5sums don't match at the end of the function, failed_copy_attempts will be incremented. Before, `failed_copy_attempts` was built into the database, but I think after the latest refactoring it just never gets updated. 

This also fixes an issue with getting files from the db where files weren't being filtered by failed_copy_attempts

Also changes `max-copy-attempts` default from None to 5 so it doesn't attempt to copy files forever by default.

Also adds better handling of files where it can't find the local path on the host (different issue, but found during the same debugging session)

## Motivation and Context
 to fix the system at Penn, where local file overwrites made it was impossible to copy over files correctly since the local md5sums have been changed.

## How Has This Been Tested?
Currently running successfully on the penn system

## Types of changes

## Checklist:

### DIFF
--- a/agents/suprsync/suprsync.py
+++ b/agents/suprsync/suprsync.py
@@ -120,7 +120,7 @@ def make_parser(parser=None):
                         help="Time (sec) after which this agent will delete "
                              "local copies of successfully transfered files. "
                              "If None, will not delete files.")
-    pgroup.add_argument('--max-copy-attempts', type=int,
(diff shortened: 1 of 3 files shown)


BrianJKoopman (approved):
Looks good to me!

PR #345, opened 2022-08-19
New control options for suprsync

mhasself:
Modifications to suprsync for use in LAT testing, where we had a low bandwidth connection (at least some of the time).

This is a feature request in the form of a PR ... @jlashner let me know if you'd rather this be done differently.

## Description

New options:
- `--compression`: enables the -z switch to rsync
- `--bwlimit`: pass through to --bwlimit (bandwidth limit) switch of rsync

New method in SupRsyncFilesManager to just return a list of registered files... this helps auto-sync scripts to not re-add files that are already entered... maybe there's already some other way to do this?  I didn't notice the bin/suprsync script when I worked out a solution for our tests.

## How Has This Been Tested?

This version is running at LAT testing now.

## Types of changes

## Checklist:

### DIFF
--- a/agents/suprsync/suprsync.py
+++ b/agents/suprsync/suprsync.py
@@ -92,6 +92,9 @@ class SupRsync:
             except subprocess.TimeoutExpired as e:
                 self.log.error("Timeout when copying files! {e}", e=e)
                 op['error'] = 'timed out'
+            except subprocess.CalledProcessError as e:
+                self.log.error("rsync returned non-zero exit code! {e}", e=e)
+                op['error'] = 'nonzero exit'
             op['stop_time'] = time.time()
             session.data['last_copy'] = op
 
--- a/agents/suprsync/suprsync.py
+++ b/agents/suprsync/suprsync.py
@@ -64,11 +64,41 @@ class SupRsync:
         self.compression = args.compression
         self.bwlimit = args.bwlimit
 
+        # Feed for counting transfer errors, loop iterations.
+        self.agent.register_feed('transfer_stats',
+                                 record=True,
+                                 agg_params={
+                                     'exclude_aggregator': True,
+                                 })
+
+
     def run(self, session, params=None):
         """run()
 
         **Process** - Main run process for the SupRsync agent. Continuosly
         checks the suprsync db checking for files that need to be handled.
+
+        Notes:
+
+            The session data object contains info about recent
+            transfers::
+
+                >>> response.session['data']
+                {
+                  "activity": "idle",
+                  "timestamp": 1661284493.5398622,
+                  "last_copy": {
+                    "start_time": 1661284493.5333128,
+                    "files": [],
+                    "stop_time": 1661284493.5398622
+                  },
+                  "counters": {
+                    "iterations": 1,
+                    "copies": 0,
+                    "errors_timeout": 0,
+                    "errors_nonzero": 0
+                  },
+                }
         """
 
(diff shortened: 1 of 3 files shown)


mhasself:
Two things I forgot to mention:
- adds agent to the plugin registration script
- adds some stuff to session.data, with ocs-web in mind

jlashner (approved):
Thanks Matthew! This looks great and I can't see any issues with it, so if you've tested it I'm happy to merge.

Since you added the `get_known_files` function I feel like we should talk about what the behavior should be when double-registering files. I was debating adding something like this or making the `local_path` a unique field in the database, but it actually seems like many times the double-registering is not the same exact file being registered twice, but a file being written, registered, and then a new file with the same path overwriting the first, due to details for how some pysmurf paths are generated. This is far from ideal, but I think since this is the case, preventing double-registering might actually be wrong, since that would prevent us from syncing either version of the file, since the first version has been overwritten, and the second version has a different local-md5sum.

I kind of like how things are currently handled, where if a double-register occurs due to a file being overwritten, the first version will just fail out due to checksum failure, and the one that's on disk will be synced correctly.

I think if `get_known_files` is useful for scripts that add files to the database then it's a good addition, I just think that it might not be the right thing to use for the pysmurf-monitor.

mhasself:
Thanks -- one more little update, to not crash the process on rsync error.  (I went to check on my test instance and found it had indeed crashed with such an error ...)

It does strike me as a bit odd that the database can have more than one row for a single filename.  As you say, as long as one of those rows has the correct file checksum, then everything should work out.  But from a forensics / error trapping standpoint this is a bit odd because you have a bunch of misleading entries in the db -- all the previous, invalidated entries for that filename.  Some of those entries might report that the file was copied successfully... which is weird, because that file (with that old checksum) won't exist on the remote.  Other entries might indicate failure, if the checksum was changed before the file could be copied.  Either way, those are odd artifacts to keep around -- the first kind gives a false sense that file X exists with checksum Y; the second might make you worry about a failed transfer... it will take more logic to pull out a list of truly, concerningly failed transfers.

Minor point if you've read this far ... I think it would be good to track file sizes in the db... that's a breaking change, I imagine, but maybe we can watch for a good time to add it in?  The usefulness is that it allows one to estimate file transfer times (based on bwlimit perhaps) and then use it to guide the batch size / timeout of the transfer.

agents/suprsync/suprsync.py
```diff
@@ -74,21 +76,33 @@ def run(self, session, params=None):
...
-                                   num_files=self.files_per_batch)
+                session.data['activity'] = 'copying'
+                op['files'] = handler.copy_files(max_copy_attempts=self.max_copy_attempts,
+                                                 num_files=self.files_per_batch)
             except subprocess.TimeoutExpired as e:
                 self.log.error("Timeout when copying files! {e}", e=e)
+                op['error'] = 'timed out'
+            except subprocess.CalledProcessError as e:
+                self.log.error("rsync returned non-zero exit code! {e}", e=e)
+                op['error'] = 'nonzero exit'
```
jlashner, line 97: I'm fine with this, though I do think that we should also add a grafana feed that we can use to alert when rsync starts failing. Before we could use the OpCode for the `run` operation to set alerts when things broke, but I think this removes the ability to do that.

jlashner (approved):

mhasself:
Ok, one last thing then ... publishes error counters to a feed.  Not sure if this is suitable for trapping or whatever, but it's graphable.

PR #357, opened 2022-10-13
Chopper MC2000B Agent

### LINKED ISSUE #249: Update labjack Agent to handle chopper data
The beam mapper chopper requires an update to the labjack Agent to be able to record chopper frequency as a function of time.

sanahabhimani:
## Description
A new agent for communicating with the chopper controller MC2000B device from Thorlabs. This agent requires a windows setup as it uses a .dll library to command the chopper controller. As a result, not implemented in a docker setup (no Dockerfile). Is an in-lab agent; will not be implemented at site. 

## Motivation and Context
Is necessary for beam map tests at UCSD during P10R1. Fixes #249 (though doesn't include a labjack setup).

## How Has This Been Tested?
Tested at Yale with an MC2000B Chopper device on a Windows 10 pro setup with supporting software from Thorlabs installed.

## Types of changes

## Checklist:


agents/chopper_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+            self.log.warn("Chopper not connected")
+
+        # Start data acquisition if requested
+        if params['auto_acquire']:
+            resp = self.agent.start('acq', params={})
+            self.log.info(f'Response from acq.start(): {resp[1]}')
+
+        return True, "Chopper controller agent initialized"
+
+    @ocs_agent.param('freq', type=int)
```
BrianJKoopman, line 131: Does the set frequency of the chopper have to be an `int`? Or can it be a `float`?

agents/chopper_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+    @ocs_agent.param('bladetype', type=str, default='MC1F2')
+    def set_bladetype(self, session, params):
+        """set_bladetype(bladetype=None)
+
+        **Task** - Set the bladetype of the chopper. Selecting a bladetype
+        influences the range of frequencies permitted for the chopper.
+
+        Parameters
+        __________
+        bladetype : str
```
BrianJKoopman, line 163: Is there documentation from Thorlabs that we can link to that describes valid blade types? Or is this something that's obvious to users? If not, we should list valid blade types.

agents/chopper_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+    @ocs_agent.param('reference', type=str, default='internalinner')
+    def set_blade_reference(self, session, params):
+        """set_blade_reference(reference=None)
+
+        **Task** - Set the reference mode for the blade. This is the point on
+        the chopper blades for the controller to measure and set frequency.
+
+        Parameters
+        __________
+        reference : str
```
BrianJKoopman, line 217: Similar to blade type, are there Thorlabs docs that describe valid reference modes? If not, we should list here.

agents/chopper_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+                                      f"currently held by {self.lock.job}.")
+                        continue
+
+                freq_in = [0]
+                MC2000BGetFrequency(self.hdl, freq_in)
+                input_freq = freq_in[0]
+
+                freq_out = [0]
+                MC2000BGetReferenceOutFrequency(self.hdl, freq_out)
+                output_freq = freq_out[0]
```
BrianJKoopman, line 274: Are the `freq_in`/`freq_out` arrays getting modified by `MC2000BGetFrequency()`/`MC2000GetReferenceOutFrequency()` functions?

agents/chopper_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+                              'timestamp': time.time(),
+                              'data': {'input_freqs': input_freq}
+                              }
+                output_data = {'block_name': 'output_freqs',
+                               'timestamp': time.time(),
+                               'data': {'output_freqs': output_freq}
+                               }
+
+                self.agent.publish_to_feed('input_freqs', input_data)
+                self.agent.publish_to_feed('output_freqs', output_data)
```
BrianJKoopman, line 287: We don't actually need two feeds here. Publishing to a single feed with different blocks/field names should be sufficient.

My suggestion would be to drop one of the feeds, and rename the remaining one something like 'chopper_freqs'.

agents/chopper_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+    args = site_config.parse_args(agent_class='ControllerAgent', parser=parser)
+
+    init_params = False
+    if args.mode == 'init':
+        init_params = {'auto_acquire': False}
+    elif args.mode == 'acq':
+        init_params = {'auto_acquire': True}
+
+    agent, runner = ocs_agent.init_site_agent(args)
+    controller_agent = ControllerAgent(agent, args.comport)
```
BrianJKoopman, line 330: `args.comport` will become `args.com_port` with the above suggestion.

docs/agents/chopper_mc2000b.rst
```diff
@@ -0,0 +1,56 @@
...
+========================
+
+The MC2000B Chopper Agent is an OCS agent which helps monitor input and output
+frequencies of the chopper, and sends commands to set the frequency of the chopper,
+as well as other features such as the bladetype and reference modes of the device.
+
+Dependencies
+------------
+
+Python Packages:
```
BrianJKoopman, line 16: I'd say remove the ":" and make this a subsection header with an underline of backticks.

docs/agents/chopper_mc2000b.rst
```diff
@@ -0,0 +1,56 @@
...
+
+========================
+Chopper MC2000B Agent
+========================
+
+The MC2000B Chopper Agent is an OCS agent which helps monitor input and output
+frequencies of the chopper, and sends commands to set the frequency of the chopper,
+as well as other features such as the bladetype and reference modes of the device.
+
+Dependencies
```
BrianJKoopman, line 13: Mention here that this needs to run on Windows. Any insight into setting that up is also useful, especially since this doesn't have a Docker image. This is our only Windows Agent, so details like setting the `$OCS_CONFIG_DIR` appropriately would be useful.

docs/agents/chopper_mc2000b.rst
```diff
@@ -0,0 +1,56 @@
...
+Configuration File Examples
+---------------------------
+
+Below are configuration examples for the ocs config file.
+
+OCS Site Config
+````````````````
+
+To configure the Holography FPGA Agent we need to add a FPGAAgent block to our
+ocs configuration file. Here is an example configuration block using all of the
```
BrianJKoopman, line 35: Replace mention of the holography agent with the chopper agent.

docs/agents/chopper_mc2000b.rst
```diff
@@ -0,0 +1,56 @@
...
+       'arguments': [['--mode', 'init'],
+                     ['--comport', 'COM3']]},
+
+Description
+-----------
+
+The ControllerAgent contains methods which control the chopper controller.
+Before the chopper can take measurements, the user needs to
+initialize the chopper controller using the init_chopper() function. This will connect to the
+chopper controller via a COM port (user specified argument in the OCS site config file).
```
BrianJKoopman, line 49: This is true, but in any currently available configuration happens automatically. Not sure we need this section unless more details are added.

docs/agents/chopper_mc2000b.rst
```diff
@@ -0,0 +1,56 @@
+.. highlight:: rst
+
+.. _chopper_mc2000b_agent:
+
+========================
+Chopper MC2000B Agent
```
BrianJKoopman, line 6: Can we rename this Agent to "Thorlabs MC2000B Agent"? That maintains consistency with most of our other Agent names in the form "<Manufacturer> <Model Number> Agent".

BrianJKoopman (changes requested):
This looks good! All the comments I have are pretty small things. A few questions about how the manufacturer's library works in there as well.

BrianJKoopman:
One more question that I just thought of. How fast does this end up sampling in the acq process loop?

agents/chopper_mc2000b/chopper_mc2000b_agent.py
sanahabhimani: Yes. `MC2000BGetFrequency()`/ `MC2000BGetReferenceOutFrequency()` both return a `ret` value that tells you whether it successfully queried the info you asked for (`0` for successful or `-1`) rather than returning the queried value. Hence my initial annoyance at the driver code.

agents/chopper_mc2000b/chopper_mc2000b_agent.py
sanahabhimani: I was going to ask you about this and completely forgot; thanks!

agents/chopper_mc2000b/chopper_mc2000b_agent.py
sanahabhimani: I just double-checked with Joe. It depends on the reference point you set for the chopper blade (and thus, the bladetype). If you've got a blade where you have small and large slits, you can pass floats for the small slits but not for the large ones.

agents/thorlabs_mc2000b/chopper_mc2000b_agent.py
```diff
@@ -0,0 +1,339 @@
...
+reference_mode_keys = {'internal': 0,
+                       'external': 1}
+
+reference_high_prec_mode = {'internalouter': 0,
+                            'internalinner': 1,
+                            'externalouter': 2,
+                            'externalinner': 3}
+
+
+class ControllerAgent:
```
BrianJKoopman, line 46: "Controller" is perhaps a bit overloaded of a word to use in "ControllerAgent". Other instances are "Motion Controller" in the fts aerotech agent, "PID Controller" in the HWP rotation agent, "Turbo Controller" in the Pfeiffer TC400 Agent, "Pysmurf Controller" in the pysmurf controller agent.

Along with the rename from "chopper_mc2000b" can we replace all instances of "ControllerAgent" with "ThorlabsMC2000BAgent"? This is here, where the class object is instantiated, in the call to `site_config.parse_args()`, and in the docs config example.

BrianJKoopman (changes requested):
Thanks for the updates! A few final comments, including a bit more renaming. I'm also going to push another change to address the pre-commit failure after the rename.

PR #390, opened 2023-01-17
Fix 372 upload cal curve timeout issue

### LINKED ISSUE #389: Timeout issue with uploading calibration curves to 372
For labs uploading rox calibration curves to 372's, a common timeout issue arises; this is because we're bombarding the 372 to upload ~200 temperature/resistance points very quickly. A time.sleep() fix in the 372 driver code should fix the timeout issue.

sanahabhimani:
## Description
Fixes timeout issue seen when uploading cal curves to 372. Also updates log statements for uploading curve points.

## Motivation and Context
Several labs see this timeout issue which in the past has been solved by continuously sending the same command until success. 
Fixes #389 

## How Has This Been Tested?
Tested in Yale's lab setup.

## Types of changes

## Checklist:


socs/Lakeshore/Lakeshore372.py
```diff
@@ -1316,11 +1320,13 @@ def set_curve(self, _file):
...
 
         for point in values:
-            print("uploading %s" % point)
+            logging.info("uploading %s" % point)
             self._set_data_point(point[0], point[1], point[2])
+            time.sleep(0.065)
 
         # refresh curve attributes
         self.get_header()
+        logging.info("checking that all points uploaded successfully...")
```
BrianJKoopman, line 1329: Move this to within the `_check_curve()` function.

socs/Lakeshore/Lakeshore372.py
```diff
@@ -1,10 +1,14 @@
 # Lakeshore372.py
 
+import logging
 import socket
 import sys
+import time
 
 import numpy as np
 
+logging.basicConfig(level=logging.INFO, format='%(levelname)s - %(message)s')
```
BrianJKoopman, line 10: This is a good addition I think. However, now we'll have two logging systems, `txaio`'s logging in the agent, and this standard logging module in the driver. This probably makes sense, since I wouldn't want to require `txaio` here. However, we should handle setting the log level similarly so they can be set simultaneously.

This is done in the agent via the `LOGLEVEL` environment variable:

https://github.com/simonsobs/socs/blob/702af44481d1cdbca34e3c217b73e51ff59fdb8f/socs/agents/lakeshore372/agent.py#L1112

Can we set it the same way here? You'll need to map the strings used to set the txaio logging to the appropriate `logging.LEVEL` from the `logging` module. `txaio` follows the log levels from twisted: https://docs.twistedmatrix.com/en/stable/core/howto/logger.html#log-levels

And here are the log levels from `logging`: https://docs.python.org/3/library/logging.html#logging-levels

socs/Lakeshore/Lakeshore372.py
```diff
@@ -1347,9 +1353,8 @@ def _check_curve(self, _file):
                 point = values[j - 1]
                 units = float(resp[0])
                 temperature = float(resp[1])
-                assert units == float(point[1]), "Point number %s not uploaded" % point[0]
-                assert temperature == float(point[2]), "Point number %s not uploaded" % point[0]
-                print("Successfully uploaded %s, %s" % (units, temperature))
```
BrianJKoopman, line 1352: Consider making this a `logging.debug()` level log if it was useful.

BrianJKoopman (changes requested):
Looks good! Just a couple small suggestions.

socs/Lakeshore/Lakeshore372.py
sanahabhimani: I'm not sure I understand exactly what the request is here. You're saying you want to make sure that the `logging` level `INFO` means the same type of logging that comes from `txaio`'s `info`? And if it doesn't, change it accordingly? In perusing those docs, it wasn't clear to me what the right `logging` level should be—I guess `debug`? (`info` says the arguments for `logging.info` are interpreted the same as `debug` which confuses me because it makes me think `info` is fine).

BrianJKoopman: The log levels in `txaio` are "debug, info, warn, error, critical" while the log levels in `logging` are "DEBUG, INFO, WARNING, ERROR, CRITICAL". My comment was in regards to 'warn' and 'WARNING' being different.

The documentation for `logging` specifies these numerical values and specifically mentions this is particularly useful if you need to define your own levels to override the defaults. This could be useful to make the log level names uniform between the two systems.

BrianJKoopman: > (`info` says the arguments for `logging.info` are interpreted the same as `debug` which confuses me because it makes me think `info` is fine).

That's just talking about the arguments to the function `logging.info()`. The docstring for `logging.debug()` has a long explanation of the `kwargs`. To save repetition they point you there, as the args at the same in `info()`.

sanahabhimani:
removed the logging statements

BrianJKoopman (approved):
Looks good, I merged the latest `main` in, will wait for tests to pass.

PR #414, opened 2023-02-16
suprsync: fix unformatted log message

mhasself:
This is a very tiny fix, in a code path that's rarely activated.


jlashner (approved):
Thanks for catching this!

PR #416, opened 2023-02-22
Add missing __init__.py for two agents

mhasself:
This would prevent the agent code for ibootbar and thorlabs_mc2000b from installing with the library.

As discussed with @BrianJKoopman  -- noticed that docker images did not have ibootbar code installed ... the `__init__.py` seems to make it install fully.


BrianJKoopman (approved):
Thanks!

PR #420, opened 2023-03-02
hwp composite agent

jlashner:
This is a proof of concept for a global HWP agent, that combines tasks and processes of all of the HWP subsystems.
This is a pretty simple implementation since it just composits the existing HWP agent classes, but uses a shared OCSAgent object between them. 

I'm curious to hear what others think of this organization, but I think it's a reasonable way to create a global HWP agent if we want to go that route.

## Motivation and Context
Creates global HWP agent that combines all subsystem operations. As discussed on the dev call, the major benefit of this is that it allows for tasks such as `shutdown` which requires coordination of the different subsystems to work even if the agent cannot connect to crossbar.

## How Has This Been Tested?

Has not been tested yet

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_composite/agent.py
+++ b/socs/agents/hwp_composite/agent.py
@@ -1,12 +1,14 @@
 
 import argparse
 import time
+
 from ocs import ocs_agent, site_config
 
-from socs.agents.hwp_rotation.agent import RotationAgent
 from socs.agents.hwp_encoder.agent import HWPBBBAgent
-from socs.agents.hwp_supervisor.agent import HWPSupervisorAgent
 from socs.agents.hwp_gripper.agent import GripperAgent
+from socs.agents.hwp_rotation.agent import RotationAgent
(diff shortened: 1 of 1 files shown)


mhasself:
Very cool.  Interesting how this supports an architecture where the "OCSAgent" is not a subclass of the Agent but actually just holds it as an instance variable.

I have seen in a few agents recently how all the register_task/process calls are in main() rather than in the Agent class `__init__.py`.  Is there specific recommendation to do one rather than the other in agent code?

jlashner:
Thanks! I think I've waffled on where operation registration should go... it's nice to have it in the `__init__` but that would make it difficult to handle things like operation name collisions, and operation overrides. Like here we replace the supervisor's `shutdown` operation with the one defined in this agent, which I'm not sure is possible if they were registered in the class init.

BrianJKoopman:
Going to close this since it's not the plan anymore. Feel free to reopen if needed.

PR #421, opened 2023-03-03
Split hwp rotation

ykyohei:
## Description

- Split hwp_rotation agent into two hwp_pmx and hwp_pid agents.
- Added a monitoring capability of current rotation freq and target rotation freq to hwp_pid.
- Changed and simplified the communication configuration (and also electronics configuration) of PMX kikusui.
  - Before: Ethernet -> Ethernet Serial Converter -> PMX
  - After: Ethernet -> PMX
- Added set_i_limit (current limit) and set_i to PMX

## Motivation and Context
Split hwp_rotation agent into two hwp_pmx and hwp_pid agents to simplyfy the hardware and DAQ.

## How Has This Been Tested?
At SAT2, TSAT, UT

## Types of changes

## Checklist:


socs/agents/hwp_pid/drivers/pid_controller.py
```diff
@@ -0,0 +1,348 @@
...
+        """Returns the target frequency of the CHWP."""
+        if self.verb:
+            print('Finding target CHWP Frequency')
+
+        responses = []
+        responses.append(self.send_message("*R01"))
+        if self.verb:
+            print(responses)
+
+        return self.target
```
frostomato, line 187: The get_target() function returns the value of self.target. However the value of self.target is not changed within the function itself which could lead to a situation where the value is inaccurate. From what I can tell, self.target only changes when the function return_messages() reads a responce which starts with "R01". This only occurs when self.verb=True and either tune_freq() or tune_stop() is run. If self.verb=False, then self.target will remain 0

bbixler500, line 187: The get_target() function returns the value of self.target. However the value of self.target is not changed within the function itself which could lead to a situation where the value is inaccurate. From what I can tell, self.target only changes when the function return_messages() reads a responce which starts with "R01". This only occurs when self.verb=True and either tune_freq() or tune_stop() is run. If self.verb=False, then self.target will remain 0

socs/agents/hwp_pid/agent.py
```diff
@@ -0,0 +1,379 @@
+import argparse
+import time
+
+from ocs import ocs_agent, site_config
+from ocs.ocs_twisted import TimeoutLock
+from twisted.internet import reactor
+
+import socs.agents.hwp_rotation.drivers.pid_controller as pd
```
bbixler500, line 8: This imports from hwp_rotation and not hwp_pid

socs/agents/hwp_pmx/agent.py
```diff
@@ -0,0 +1,277 @@
+import argparse
+import os
+import time
+
+import txaio
+from ocs import ocs_agent, site_config
+from ocs.ocs_twisted import TimeoutLock
+
+import socs.agents.hwp_rotation.drivers.PMX_ethernet as pmx
```
bbixler500, line 9: This imports from hwp_rotation and not hwp_pmx

socs/agents/hwp_pid/drivers/pid_controller.py
ykyohei: Thanks for pointing out.
I think I fixed this issue in https://github.com/simonsobs/socs/pull/421/commits/7eff0b1e0dca529d8b371e39002a4051b40ba6bd
Now the `self.target` will be updated every time we call `tune_freq()` or `tune_stop()` and `self.target` will always stay the latest value. Could you check if it looks good?

bbixler500 (approved):

ykyohei:
I finished reviewing the codes. Can you check if it looks good to merge? @BrianJKoopman

ykyohei:
I'm going to merge this PR this weekend because we only have relatively minor changes, which @bbixler500 approved, and all checks have passed. Let me know in case if there is any problems to merge.

BrianJKoopman:
> I'm going to merge this PR this weekend because we only have relatively minor changes, which @bbixler500 approved, and all checks have passed. Let me know in case if there is any problems to merge.

Apologies, I've been preparing to deploy to the site early next week and haven't gotten to this yet. I'll do so first thing in the morning.

socs/agents/hwp_pid/drivers/pid_controller.py
```diff
@@ -0,0 +1,351 @@
...
+            else:
+                pass
+
+        return output_array
+
+    def _decode_read(self, string):
+        read_type = string[1:3]
+        if read_type == '01':
+            self.target = float(int(string[4:], 16) / 1000.)
+            return 'Setpoint = ' + str(self.target)
```
BrianJKoopman, line 321: I don't think `@staticmethod` should be removed from this. The value that you're setting `self.target` to gets returned and passed up through to `return_messages()`, `self.target` should be set where you call `return_messages()` (like in `set_target` or `get_target` or both), not all the way down here.

I removed a similar structure from the direction reading back here: https://github.com/simonsobs/socs/commit/634ec8506a136aec6e50e21e2fd7b27346f1f8bf

BrianJKoopman (changes requested):
Hi @ykyohei, thanks for the PR. There are several things to address before merging. In addition to the comments inline below:

- The original HWP Rotation Agent should simultaneously be removed.
- I wrote a set of tests for [the original rotation agent code](https://github.com/simonsobs/socs/blob/main/tests/integration/test_hwp_rotation_agent_integration.py) and the [pid_controller drivers](https://github.com/simonsobs/socs/blob/main/tests/agents/hwp_rotation/test_pid_controller.py), these should be adapted for the split agent.

I'd also like to understand more about the change in the PMX setup, and whether that will also carry over to the wiregrid hardware. I commented on it below, but I just want to see if we need to maintain both methods of communication.

Again, sorry to be slow to respond to this. In the future it'd be helpful for the reviewer if code restructures occur separately from changes to the same code. It makes it difficult to follow what has changed.

socs/agents/hwp_pmx/agent.py
```diff
@@ -0,0 +1,277 @@
...
+                                'last_updated': current_time}
+                time.sleep(sleep_time)
+            self.agent.feeds['Rotation'].flush_buffer()
+        return True, 'Acquisition exited cleanly.'
+
+    def stop_acq(self, session, params=None):
+        """
+        Stop acq process.
+        """
+        self.dev.close()
```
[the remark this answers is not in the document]
ykyohei: Thanks for pointing out.
I added close method here https://github.com/simonsobs/socs/pull/421/commits/21a0a080e35c616c888e865d1faea2eae496e1df.

socs/agents/hwp_pid/drivers/pid_controller.py
ykyohei: Thanks for the suggestion. I fixed the structure of target freq reading. https://github.com/simonsobs/socs/pull/421/commits/3a9d36da49c1b3626c07f419cc8526fbe947e4fe

socs/agents/hwp_pmx/agent.py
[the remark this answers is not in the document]
BrianJKoopman: Ah, great. However, now if the acquisition process is stopped it will close the connection to the device and it will be required that `init_connection` be run before `acq` can run again. Is this the desired behavior?

I think `self.dev.close()` shouldn't be called in the `_stop_acq()` function. That way the connection stays open and doesn't need to be reinitialized. This is the behavior in other Agents.

socs/agents/hwp_pmx/drivers/PMX_ethernet.py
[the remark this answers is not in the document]
BrianJKoopman: Alright, let's keep the current structure then and look for an opportunity to use this Ethernet version for the wiregrid at a later time.

BrianJKoopman (commented):
Thanks for making the requested changes @ykyohei. After looking through the changes I have just one remaining question, see the comment here: https://github.com/simonsobs/socs/pull/421#discussion_r1190385415

I made some small fixes to the docs that I pushed to this branch. I also went ahead and removed the old HWP Rotation agent and split the tests for it to test the two new agents. That I haven't pushed to this branch. If you would like, please take a look at the changes in #440.

My hope is to merge this, work out the `self.dev.close()` behavior you would like, and then we'd be almost ready to merge. I need to do a rebase of this entire branch to avoid losing history on `pid_controller.py` (which was copied and not `git mv`'d to the new location in `socs/agents/hwp_pid/drivers/`.) I'll wait until everything else has settled to do that.

socs/agents/hwp_pmx/agent.py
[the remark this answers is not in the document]
ykyohei: I see. Thanks.
I simply removed `self.dev.close()` from `_stop_acq()` in the same way as other agents. https://github.com/simonsobs/socs/pull/421/commits/774008b49936eb78447fc0ae5e31e051dd0a7aa9

BrianJKoopman (approved):
Great, approved! Just need to wait on merging #440, and then I'll do the rebase to preserve history of `pid_controller.py`. Once that's done we can merge, but not until then.

Issue #385, opened 2022-12-20
pysmurf-controller needs to deal with tags in more concrete way.


kmharrington:
Part of the plan for implementing tags / observation tracking from the schedule on down involves the pysmurf-controller passing tags from sorunlib into sodetlib. (Whole plan outlined in the [flow to librarian document](https://github.com/simonsobs/tod2maps_docs)

We can technically do the pysmurf-controller --> sodetlib through the passing of kwargs, but I'd like it to be one of the default parameters (like duration in take_noise currently is). 

I'm assuming the sorunlib --> pysmurf-controller parts haven't been linked?

PR #424, opened 2023-03-07
adds pysmurf-controller tag params

jlashner:
Adds `tag` param to pysmurf-controller operations, which will append a user-provided tag to the end of the g3 stream tag.

This + https://github.com/simonsobs/sodetlib/pull/336 resolves https://github.com/simonsobs/socs/issues/385

## Motivation and Context
Helpful for the scheduler to be able to append tags to streams or other sodetlib operations.

## How Has This Been Tested?
Not tested yet. This should be tested with https://github.com/simonsobs/sodetlib/pull/336 before merging.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/pysmurf_controller/agent.py
+++ b/socs/agents/pysmurf_controller/agent.py
@@ -331,6 +331,7 @@ class PysmurfController:
     @ocs_agent.param("duration", default=None, type=float)
     @ocs_agent.param('kwargs', default=None)
     @ocs_agent.param('load_tune', default=True, type=bool)
+    @ocs_agent.param('tag', default=None)
     @ocs_agent.param('test_mode', default=False, type=bool)
     def stream(self, session, params):
         """stream(duration=None)
@@ -350,6 +351,9 @@ class PysmurfController:
         load_tune : bool
             If true, will load a tune-file to the pysmurf object on
             instantiation.
+        tag : Optional[string]
+            Tag (or comma-separated list of tags) to attach to the G3 stream.
+            This has precedence over the `tag` key in the kwargs dict.
 
         Notes
         ------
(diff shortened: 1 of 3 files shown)


socs/agents/pysmurf_controller/agent.py
```diff
@@ -331,6 +331,8 @@ def _stop_check_state(self, session, params):
     @ocs_agent.param("duration", default=None, type=float)
     @ocs_agent.param('kwargs', default=None)
     @ocs_agent.param('load_tune', default=True, type=bool)
+    @ocs_agent.param('stream_type', default='obs')
```
BrianJKoopman, line 334: Could use the `choices=['obs', 'oper']` argument here to limit the possible inputs.

BrianJKoopman (commented):
Changes look good! Just one additional suggestion from me.

BrianJKoopman (commented):
I have a new question about this. It looks like `g3_tag` on operations overrides default tags like "oper,bgmap". (Should we always be overriding these?)

I just want to confirm that adding a user defined tag on a stream doesn't override the default "obs,stream" tags.

jlashner:
Right, in conjunction with [this sodetlib PR](https://github.com/simonsobs/sodetlib/pull/336), user-defined tags are set through a separate kwarg from the stream type/subtype. So when we test we'll need to test both these PRs together.

socs/agents/pysmurf_controller/agent.py
kmharrington: I do vote for limiting the options here because that will prevent accidents / errors later down the chain.

kmharrington (approved):
Reviewing in combo with this one: https://github.com/simonsobs/sodetlib/pull/336 

Looks good to me modulo the one in-line comment. I was able to test passing subtypes to observations with stream with and w/o extra tags. I was also able to test extra tags while taking IVs and noise. The bias steps and bgmaps are throwing errors but I'm told that's due to a different error that has been fixed elsewhere. So assuming the tag implementation is the same between all the operations we should be good here!

BrianJKoopman:
Great, thanks for testing @kmharrington!

@jlashner can we merge https://github.com/simonsobs/sodetlib/pull/336 and then also bump the docker base image for the controller in this PR?

PR #425, opened 2023-03-14
File emulator quantization

jlashner:
Makes several changes to the Smurf File Emulator so that it's better for testing the bookbinder.

## Description

- Adds "timestamp quantization".  By clamping the start / stop times to integers, `np.arange` with an integer sample freq will create close to co-sampled data, even if the start / stop times are different between streams.
- Adds correct tags to the stream function
- Adds the ability to set a drop-chance, which will toss a subset of data points.
- Adds some params that tell the `stream` task to use the `stream_between` function of the streamer, instead of generating frames one-by-one, allowing for the task to run for a non-zero duration without sleeping.

## Motivation and Context

This is required to properly test the bookbinder on the emulated data.

## How Has This Been Tested?

Tested by creating sample data and running it all the way through to books.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/smurf_file_emulator/agent.py
+++ b/socs/agents/smurf_file_emulator/agent.py
@@ -491,8 +491,9 @@ class SmurfFileEmulator:
         if session is not None:
             session.data['noise_file'] = streamer.file_list[0]
 
+    @ocs_agent.param('sleep', default=True)
     def uxm_setup(self, session, params):
-        """uxm_setup(test_mode=False)
+        """uxm_setup(sleep=True)
 
         **Task** - Emulates files that might come from a general tune dets
         function. These are some of the files found on simons1 registered when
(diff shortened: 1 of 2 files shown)


socs/agents/smurf_file_emulator/agent.py
```diff
@@ -475,7 +505,8 @@ def uxm_setup(self, session, params):
         """
         self._run_setup(action='uxm_setup', action_time=time.time(),
                         session=session)
-        time.sleep(1)
+        if params.get('sleep', True):
```
BrianJKoopman, line 508: Create a `Parameters:` section in the docstring with the new 'sleep' parameter and the old 'test_mode' parameter.

(`@ocs_agent.param` decorators could also be used.)

BrianJKoopman (changes requested):
This looks good! Just one docstring related request.

socs/agents/smurf_file_emulator/agent.py
jlashner: Not sure where test_mode was copied from, but it's not actually implemented here... so I just removed from the spec line.

jlashner:
Great, thanks brian! Fixed the uxm_setup docstring.

BrianJKoopman (approved):
Awesome, looks good!

PR #427, opened 2023-03-20
Hwp supervisor

jlashner:
## Description
Another go of what the HWP supervisor agent might look like based on our conversation with the HWP group.

Based on that discussion, I think for the supervisor we want an agent that has a very robust `monitor` process, that compiles data from various HWP-related agents into a single object, that can be easily queried by HWP agents or OCS web-panels.

This agent tries to enumerate the essential HWP-related systems discussed on the call and compile relevant session-data into a single place.

Instance-ids for all of the hwp sub-systems will be passed in as args, so we can also use this agent to define high-level operations.

## Motivation and Context

Compiles HWP and telescope information for HWP subsystems to use to inform state-changes for emergency shutdown.

## How Has This Been Tested?

This has been tested locally, by passing in instance-ids that correspond to fake-data-agents running on my system. I've verified that the monitor task works for the following cases:
- instance-id points to fake-data agent running locally -- session-data of the `acq` task is copied in
- instance-id points to agent that is not running: writes the dict 
  ```
  'hwp_encoder': {'agent_id': 'abcd', 'op_name': 'acq', 'timestamp': 1679330788.8817987, 'data': None, 'status': 'op_not_found'}
  ```
- No instance id provided for specified agent:
  ```
  'hwp_rotation': {'agent_id': None, 'op_name': 'iv_acq', 'timestamp': 1679330788.8961837, 'data': None, 'status': 'no_agent_provided'}
  ```

## Types of changes

## Checklist:


jlashner:
This still needs docs, but wanted to check with @BrianJKoopman and @mhasself to see if this is the right path.

mhasself:
This looks good.  Remember this will also be a place for higher-level HWP operations that require multi-agent coordination ... e.g. setting it to spin at a certain speed.

Collecting misc session.data into one place is good ... should the assessment of this information also live here?  For example, the assessment of whether or not the temperature is "too high".  If it's not here, then it may need to be duplicated in multiple other places; that's not fun.

It seems inevitable that this agent will need a configuration file ... potentially many limits and handling details to track.  Maybe you prefer to have lots of command line args and keep it all in the SCF though.

jlashner:
I think the structure of this agent is far enough along that we can review it and think about merging. There are still details we need to figure out (that I could honestly use some help with) like:
- how to know that the HWP has stopped spinning. Will the encoder agent report a hwp_freq == 0, or will the encoder fail to readout at such low speeds?
- What should we be monitoring in the UPS agent and how do we use UPS session data to determine subsystem actions?

I personally don't feel I know enough about the ups / hwp subsystems to make this determination right now, but I think the current structure of the `monitor` task makes this relatively easy to update. I think this is a good starting point so I think we might want to review and merge, and then add more state logic as we're able to test.

I think the logic for the subsystem shutdown tests should be as follows (for the rotation agent as an example).
1.  query the supervisor session data.
2. check data['actions']['rotation']. 
    - If this is `'ok'`, continue 
    - If this is `'stop'`, stop the rotation and set a `shutdown` flag that needs to be manually turned off with a `reverse_shutdown` task before other operations can change the rotation rate
    - If this is `'no_data'` or if you can't connect to the supervisor, trigger a warning in grafana if X minutes have passed since last `ok`, and enter shutdown if Y minutes have passed since last ok, where X and Y are input params.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,492 @@
...
+            >>> response.session['data']
+
+                {'forward': True,
+                'commanded_freq': 2.0,
+                'pid_freq': 1.2}
+
+        Parameters
+        ------------
+        forward : bool
+            If True, will spin the HWP in the forward direction.
```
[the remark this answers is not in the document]
ykyohei: Thanks for pointing it out.

This is not described in the so coordinate system document yet, but it is [proposed](https://docs.google.com/presentation/d/1LKq5hsVltEiDbHQZ6quGuD-I8CXO6oDZtlnc_la0bN8/edit?usp=sharing) to define the "forward hwp rotation direction" as "counter-clockwise from the sky", so I think we should define it that way.

I think the "forward" and "reverse" for PID agent is fine as it is, because PID agent is just turning on or off one of the digital outputs to change rotation direction.

Because of the slight differences of the rotation mechanism of the three SATs, the definition of "forward" and "reverse" for PID agent is not physically consistent among three SATs.
So, I think we need a config file for hwp supervisor agent.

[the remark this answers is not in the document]
mhasself: Can I suggest that instead of separate arguments for direction and rate, that only a single argument is given: the rate, as a _signed float_.  So `target_freq=2`. would spin at 2Hz CCW, `target_freq=-2` would spin at 2 Hz CW.

Also ... some people think that we'll only ever spin the HWP one direction.  But not I.  I want to spin it both directions, equally.  So defaulting to forward, or not forward, is not what I would expect.

So the HWP supervisor needs to be configurable, with regards to how +ve / -ve spin becomes a direction for the PID agent... that can just be command line arguments to the HWP supervisor agent.  The "config file" is the OCS site config file.  (I'm not against having a separate config file ... it's just not totally needed here.)

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,492 @@
...
+                                  args=args)
+
+    agent, runner = ocs_agent.init_site_agent(args)
+    hwp = HWPSupervisor(agent, args)
+
+    agent.register_process('monitor', hwp.monitor, hwp._stop_monitor,
+                           startup=True)
+    agent.register_task('spin_up', hwp.spin_up)
+    agent.register_task('spin_down', hwp.spin_down)
+
```
mhasself, line 487: Currently spin_up and spin_down are tasks, and contain potentially long-running, uninterruptible loops.  That's a problem.  There's also just a lot of duplication in those two functions.

I propose that instead of separate tasks, a process be used to manage the spin.  The process would normally be running, and would be smart about special speeds (such as stopping).  There would be separate, ~instantaneous task(s) available to change the target frequency / request stop.

docs/agents/hwp_supervisor_agent.rst
```diff
@@ -0,0 +1,81 @@
...
+
+This agent has two main purposes:
+
+- Monitor HWP subsystems and related agents to make high level determinations
+  such as when subsystems should start their shutdown procedure
+- Serve as a host for high-level HWP operations that require coordinated control
+  of various HWP subsystems, such as "Begin rotating at 2 Hz"
+
+Right now only the first point is implemented, but operations can be added here
+as we need them.
```
mhasself, line 68: I think the second point is in progress too...

mhasself (changes requested):
See grab bag of comments ... I think the structure for safety stuff is pretty good.  spin_up / spin_down should probably be reworked.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,492 @@
...
+        - ``test_mode``: if ``test_mode=True``, this will be the status and no
+          operation will be queried.
+        - ``op_not_found``: This means the operation could not be found, likely
+          meaning the agent isn't running
+        - ``no_active_session``: This means the operation specified exists but
+           was never run.
+        - ``ok``: Operation and session.data exist
+    """
+    if log is None:
+        log = txaio.make_logger()  # pyright: ignore
```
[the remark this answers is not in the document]
jlashner: Oops, I was but can remove this

jlashner:
Hi all, thanks so much for the comments, they have been really helpful. Based on Matthew and Brian's comments, it seems that instead of having separate tasks / processes in the supervisor to do things like `spin_up`, we should be working towards the model where we have a central process with a state-machine that is responsible for issuing all commands to agents under its control, and  tasks and processes for behaviors like `pid_to_freq` that put the agent in a specific state and return instantly. I do agree that this seems like a much more robust way of handling control. In the most recent commit I tried to implement this, so I'd really appreciate it if everyone could take another look.

All of the logic for how to control the HWP is now in the `ControlState.update()` function, so @ykyohei / @mhasself if you could look at that and we can work on making sure that is inline with HWP control procedures that would be great!

Also @mhasself it would be really great (for testing in particular) if we had an ocs-web panel to let us monitor and set the state, so if you see anything that we could add to make that easier let me know!

I have not tested this on a real system, but tomorrow I may be able to squeeze in some time to test on the tsat.  Also I still have to implement a few of Matthew's comments on the pmx shutdown procedure, and update docs and tests.

socs/agents/hwp_supervisor/agent.py
ykyohei, line 1: Do we have plan to also manage synacc agent (powerstrip for hwp electronics inside hwp enclosure) by supervisor?

As a operating requirement from HWP team,
we would like to turn on the power of HWP encoder LED when 2nd stage is at stable operating temperature, because it is known that large temperature changes while biasing the LEDs will degrade the LEDs.
To turn off the encoder LED, we need to turn off the some of the outputs of power strip.
So, we would like to implement, for example, to turn off the leds every time after stopping hwp spin or after shutdown.

Also, I think we discussed that we would like to shut the entire power from power strip when necessary.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+                    elif time.time() - self.freq_within_thresh_start > self.freq_thresh_duration:
+                        self.set_state_type(ControlStateType.DONE, success=True)
+                else:
+                    self.freq_within_thresh_start = None
+            else:
+                self.freq_within_thresh_start = None
+
+        elif self.state_type == ControlStateType.CONST_VOLT:
+            # Set to constant voltage mode
+            clients.pmx.ign_ext()
```
ykyohei, line 404: `clients.pmx.ign_ext()`
This line looks unnecessary because it is called in next line.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+            # Check if we are close enough to the target frequency.
+            # This will make sure we remain within the frequency threshold for
+            # ``self.freq_thresh_duration`` seconds before switching to DONE
+            f = hwp_state.pid_current_freq
+            if f is not None:
+                # If within_thresh for freq_thresh_duration, switch to Done
+                if np.abs(f - self.target_freq) < self.freq_thresh:
+                    if self.freq_within_thresh_start is None:
+                        self.freq_within_thresh_start = time.time()
+                    elif time.time() - self.freq_within_thresh_start > self.freq_thresh_duration:
```
ykyohei, line 395: Question about the behavior of `freq_thresh_duration`
If the hwp did't reach close enough to the `target_freq` with in `freq_thresh_duration`

Does the hwp task end at that time? and is the task to be considered as ` success=True`?
I think the warning like "hwp didn't reach `target_freq` within `freq_thresh_duration`" will be useful.

jlashner: `freq_thresh_duration` is the amount of time that the freq has to be within the threshold in order for it to be considered a success. I added this so we don't consider it a success if the HWP frequency just passes through the target frequency but doesn't stop.

Right now there isn't a time limit for which we switch to the DONE state with success=False, but that is something we can easily add if we want.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+    @ocs_agent.param('freq_thresh_duration', type=float, default=10)
+    def pid_to_freq(self, session, params):
+        """pid_to_freq(target_freq=2.0, freq_thresh=0.05, freq_thresh_duration=10)
+
+        **Task** - Sets the control state to PID the HWP to the given ``target_freq``.
+
+        Args
+        -------
+        target_freq : float
+            Target frequency of the HWP (Hz). Positive values are CW, negative are CCW.
```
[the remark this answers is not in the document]
mhasself: I second that, please be very clear about convention.  "Positive values correspond to counter-clockwise motion, as seen when looking at the cryostat face from the outside."

mhasself:
> All of the logic for how to control the HWP is now in the `ControlState.update()` function, so @ykyohei / @mhasself if you could look at that and we can work on making sure that is inline with HWP control procedures that would be great!

I do like a state machine... that looks great.  Some comments (apologies if I missed some subtleties, wasn't a complete read):
- DONE and NONE are effectively the same thing.  I'd probably call that state "IDLE".
- You'll probably want to issue a log message on each state change.
- You might end up needing to split WAIT_FOR.. into two states, waiting while braking vs. waiting while spinning up.  (And maybe one for spin-down-by-friction?)  Seems like in the braking/stopping cases, a natural thing to do when you reach the target freq is to turn off the power; WAIT_FOR_STOP?
- You may want to record and report the time spent in the present state; especially for spinning up / braking it will be useful for an operator to know how long ago that state was entered.
- I was reassured to see "run_and_validate" wrapping all those client calls ... but there is woefully little error _validation in there :)

socs/agents/hwp_pmx/agent.py
```diff
@@ -260,6 +298,86 @@ def _stop_acq(self, session, params=None):
...
+        """
+        self.log.warn("INITIATING SHUTDOWN")
+
+        with self.lock.acquire_timeout(10, job='shutdown') as acquired:
+            if not acquired:
+                self.log.error("Could not acquire lock for shutdown.")
+                return False, "Could not acquire lock."
+
+            self.shutdown_mode = True
+            self.dev.ign_external_voltage()
```
bbixler500, line 314: Technically this line is unnecessary, as you can shut off the power in both run modes (with or without PID)

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+                raise ValueError("PID dir must be set for PID_FREQ type")
+            if self.freq_thresh is None:
+                raise ValueError("Frequency Threshold must be set for PID_FREQ type")
+            if self.freq_thresh_duration is None:
+                raise ValueError("Freq Threshold Duration must be set for "
+                                 "PID_FREQ type")
+
+        if self.state_type == ControlStateType.CONST_VOLT:
+            if self.voltage is None:
+                raise ValueError("Voltage must be set for CONST_VOLT type")
```
bbixler500, line 319: You might want to add a catch for not setting `self.pid_dir` here as well. Even if the PID is not in use to control the frequency, it is still used to control the rotation direction

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+            # Set to constant voltage mode
+            clients.pmx.ign_ext()
+            self.run_and_validate(clients.pmx.ign_ext)
+            self.run_and_validate(clients.pmx.set_v,
+                                  kwargs={'voltage': self.voltage})
+            self.set_state_type(ControlStateType.DONE, success=True)
+
+        elif self.state_type == ControlStateType.BRAKE:
+            # Actively set the PID to brake the HWP
+            self.run_and_validate(clients.pid.tune_stop)
```
bbixler500, line 412: The `clients.pid.tune_stop` function reverses the rotation direction, changes the pid parameters, and set the setpoint to 0 Hz. If this function is run unattended, the HWP will reach 0 Hz and then start accelerating in the opposite direction. To fully stop the HWP the power to the PMX needs to be turned off as the rotation frequency approaches 0 Hz.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+            ``target_freq`` to be considered successful.
+        """
+        if params['target_freq'] >= 0:
+            d = '0' if self.forward_is_cw else '1'
+        else:
+            d = '1' if self.forward_is_cw else '1'
+
+        self.control_state = ControlState(
+            state_type=ControlStateType.PID_FREQ,
+            target_freq=params['target_freq'],
```
bbixler500, line 658: I believe that this should be `target_freq=abs(params['target_freq'])` as the function it calls only accepts positive values

socs/agents/hwp_supervisor/agent.py
```diff
@@ -0,0 +1,773 @@
...
+                        'data': <session data for test.acq>,
+                        'op_name': 'acq',
+                        'status': 'ok',  # See ``get_op_data`` docstring for choices
+                        'timestamp': 1680273288.6200094},
+                    },
+                    'rotation': {see above},
+                    'temperature': {see above},
+                    'ups': {see above}},
+                # State data parsed from monitored sessions
+                'state': {
```
bbixler500, line 534: Is this docstring still up to date?

ykyohei:
Thank you so much for your hard work, and sorry for late response.

About the PID braking logic, I think we should use the rotation direction and rotation frequency of PID agent, not of the encoder.
This is because the quadrature of encoder returns wrong bool value at some rate because of the reference slot of encoder, also I think it is simpler using PID agent.
https://github.com/simonsobs/socs/pull/427/commits/396f9c878162d749c7f469035e40c741d25fadb2
I think we'll address this too in other PR or issues, so I approve this PR.

ykyohei (approved):

PR #428, opened 2023-03-24
Fixes sampling issues with file emulator

jlashner:
While looking at end-to-end testing data, found a couple of bugs which was causing large numbers of dropped samples, and ~20us variation between timestamps of different slots. This PR should fix both those issues.

## Description

- The "write_next" function will take the previous stop time as the next starting time.
- Use `np.linspace` instead of `np.arange` to fix floating point buildup

## Motivation and Context
better end-to-end test data.

## How Has This Been Tested?

Tested on simons1 to check this fixes both issues.

## Types of changes

## Checklist:


guanyilun (approved):
Looking good!

PR #431, opened 2023-04-19
Adds shutdown procedure and logic to HWP PMX agent

jlashner:
This PR implements a shutdown procedure for the HWP PMX agent.

## Description

This shows how one may use the hwp-supervisor session data to trigger a shutdown procedure, and how a shutdown might lock out specific tasks or processes until it is reversed.

This implements three additional operations:
- monitor_shutdown is a process that continuously monitors hwp-supervisor rotation action to determine whether or not a shutdown should be triggered. This will trigger a shutdown if either the action is 'stop', or it has been `no_data_timeout` since the last 'ok' action received. The action will be published to grafana, so an alarm can be set that triggers if the action is not 'ok' for a certain amount of time.
- The 'initiate_shutdown' task will set `shutdown_mode=True`, and will tell the PMX to cut power and to ignore external voltage from the PID.
- The `reverse_shutdown` task sets `shutdown_mode=False`

While `shutdown_mode` is True, any task that tries to change the PMX state in a way that might interfere with the shutdown (such as `set_on`, `set_i`, `set_v`...) will fail.

## Motivation and Context

Required for emergency HWP shutdown

## How Has This Been Tested?

This has not been tested yet

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -359,11 +359,10 @@ class HWPPMXAgent:
             }
 
             self.agent.publish_to_feed('gripper_action', data)
-            session.data =  {
+            session.data = {
                 'rotation_action': action,
                 'time': time.time()
             }
-
 
             time.sleep(0.2)
 
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -361,7 +361,11 @@ class HWPPMXAgent:
             }
 
             self.agent.publish_to_feed('gripper_action', data)
-            session.data = data['data']
+            session.data =  {
+                'rotation_action': action,
+                'time': time.time()
+            }
+
 
             time.sleep(0.2)
 
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -8,6 +8,7 @@ from ocs.ocs_twisted import TimeoutLock
 from twisted.internet import reactor
 
 import socs.agents.hwp_pmx.drivers.PMX_ethernet as pmx
+from socs.agents.hwp_supervisor.agent import get_op_data
 
 txaio.use_twisted()
 
@@ -20,9 +21,13 @@ class HWPPMXAgent:
         ip (str): IP address for the PMX Kikusui power supply
         port (str): Port for the PMX Kikusui power supply
         f_sample (float): sampling frequency (Hz)
+        supervisor_id (str): ID of HWP supervisor 
+        no_data_timeout (float): Time (in seconds) to wait between receiving 
+            'no_data' actions from the supervisro and triggering a shutdown
     """
 
-    def __init__(self, agent, ip, port, f_sample=1):
+    def __init__(self, agent, ip, port, f_sample=1, supervisor_id=None, 
+                 no_data_timeout=15*60):
 
         self.agent: ocs_agent.OCSAgent = agent
         self.log = agent.log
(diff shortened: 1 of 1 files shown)


socs/agents/hwp_pmx/agent.py
```diff
@@ -250,6 +285,83 @@ def _stop_acq(self, session, params=None):
...
+        self.log.warn("INITIATING SHUTDOWN")
+
+        with self.lock.acquire_timeout(10, job='shutdown') as acquired:
+            if not acquired:
+                self.log.error("Could not acquire lock for shutdown.")
+                return False, "Could not acquire lock."
+
+            self.shutdown_mode = True
+            self.dev.ign_external_voltage()
+            self.dev.stop()
```
bbixler500, line 301: I don't see a "stop()" method for self.dev as defined in "/socs/agents/hwp_pmx/drivers/PMX_ethernet.py". If you're intending to turn off the PMX output I would use "self.dev.turn_off()" instead.

jlashner: Ah great, thanks for catching this!

socs/agents/hwp_pmx/agent.py
```diff
@@ -250,6 +285,83 @@ def _stop_acq(self, session, params=None):
...
+                if not self.shutdown_mode:
+                    self.agent.start('initiate_shutdown')
+
+            data = {
+                'data': {'rotation_action': action},
+                'block_name': 'rotation_action',
+                'timestamp': time.time()
+            }
+
+            self.agent.publish_to_feed('gripper_action', data)
```
bbixler500, line 351: Should this be 'rotation_action' instead of 'gripper_action'?

socs/agents/hwp_pmx/agent.py
```diff
@@ -20,9 +21,13 @@ class HWPPMXAgent:
         ip (str): IP address for the PMX Kikusui power supply
         port (str): Port for the PMX Kikusui power supply
         f_sample (float): sampling frequency (Hz)
+        supervisor_id (str): ID of HWP supervisor
+        no_data_timeout (float): Time (in seconds) to wait between receiving
+            'no_data' actions from the supervisro and triggering a shutdown
```
ykyohei, line 26: typo 'supervisro'

ykyohei (approved):
Sorry for late review, and thank you for your work.
Shutdown procedure looks reasonable to me.

BrianJKoopman:
Sorry, I had to rebase the branch this was being PR'd onto (in https://github.com/simonsobs/socs/pull/421). This has messed up this branch a bit, as it still has all of those rebased commits. You should probably rebase on `main`.

socs/agents/hwp_pmx/agent.py
```diff
@@ -260,6 +295,83 @@ def _stop_acq(self, session, params=None):
...
+            # trigger shutdown
+            elif action == 'no_data':
+                if (time.time() - last_ok_time) > self.no_data_timeout:
+                    if not self.shutdown_mode:
+                        self.agent.start('initiate_shutdown')
+
+            # If action is 'shutdown', trigger shutdown
+            elif action == 'stop':
+                if not self.shutdown_mode:
+                    self.agent.start('initiate_shutdown')
```
BrianJKoopman, line 353: You should also pass `params` here, even if just `None`. (If `initiate_shutdown` had the params decorator, this as is would run into an error.) Same goes for other direct calls to `.start()`.

socs/agents/hwp_pmx/agent.py
```diff
@@ -260,6 +295,83 @@ def _stop_acq(self, session, params=None):
         else:
             return False, 'acq is not currently running'
 
+    def initiate_shutdown(self, session, params):
+        """
```
BrianJKoopman, line 299: These task docstrings all need the method signature override at the top, and to be marked with `**Task**` or `**Process**` appropriately. See the [example here](https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html#task-and-process-documentation).

BrianJKoopman (changes requested):
A few small ocs related requests. Also, after seeing the import from the supervisor agent, I do think we should merge this and #427 together before merging to `main`.

BrianJKoopman (approved):
Looks good to merge into [hwp_supervisor](https://github.com/simonsobs/socs/tree/hwp_supervisor).

PR #433, opened 2023-04-26
IFM Electronic SBN246 Flowmeter Agent

### LINKED ISSUE #66: Flowmeter agent
https://www.ifm.com/us/en/product/SBN246

sanahabhimani:
## Description
Agent for monitoring flow and temperature of cooling loops at the site. Uses a DAQ IO-Link DAQ device from IFM Electronic on top of the IFM electronic SBN246 flowmeter. Agent uses http requests to query flow and temp data. Requests are made to the IO-Link device for the port the flowmeter is connected to. 

## Motivation and Context
Closes #66 

## How Has This Been Tested?
Tested at Yale with physical hardware set up in the lab. Has been tested in and out of a docker container.

Need to write tests as well, separate issue will be added soon.

## Types of changes

## Checklist:


BrianJKoopman (changes requested):
This is look good! Most of my comments are documentation/wording related.

Do we know how many of these flowmeters will be on site? I debated for a while if it would be better to have an agent per IO-Link master, rather than per flowmeter, but I'm not sure what other IO-Link devices we have, if any.

I'm also still a bit worried that eventually a model of flowmeter that natively reads out metric units will make its way to the site. On initialization we could check `/iolinkmaster/port[N]/iolinkdevice/productname/getdata` to ensure it is an SBN246 we're talking to.

docs/agents/ifm_sbn246_flowmeter.rst
```diff
@@ -0,0 +1,94 @@
...
+--------------------------------
+After plugging into the IoT port on your IFM Electronic AL1340 DAQ IO-Link device,
+the IP address of the IO-Link device must be configured using the LR Device software.
+On the LR Device software, click the 'read from device' on the upper right (leftmost
+IOLINK button). Once the software has found your IO-Link device, click on the device name
+(AL1340) on the column on the left side (not the ports that may say P1, P2... or the name
+of the device connected to any of the 4 ports on the IO-Link device). You should now be
+able to change the IP address to by clicking the Network (DHCP) button.
+
+ - `LR Device Software <https://www.ifm.com/de/en/download/LR_Device>`_
```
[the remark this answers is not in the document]
sanahabhimani: I think we do need to go in and set it to DHCP? Memory's already a little foggy now but I seem to recall the LR software finds the device, and assigns the factory setting ip address until we go into network parameter and toggle it to DHCP even for the IoT. And I think that's how you got it from 169.254.X.X to 10.10.10.159? Am I remembering that correctly? I also think that was our big argument for PR'ing the http requests version of this agent since we have to do a network configuration via the IoT port for this device anyway

sanahabhimani:
I don't know the final number of flowmeters that will be in use for the site. 2 per cooling loop? More? I know there is a desire for Oriel to buy an IFM pressure gauge to hook into the IO-Link master, but I don't know if that's been done yet or how finalized that is. Thinking about it today, it may be worth considering writing an agent for the IO-Link device—there are cables long enough that you'd be able to connect multiple flowmeters into one IO-Link device. It does sort of make sense to go that way? And if we're possibly adding an IFM pressure gauge.

I added the check for the model number on initialization; not sure if you want the assertion error message to be a `self.log` message also? Added in the `session.data` output into the `acq()` docstring too. Let me know if the `session.data` structure should be different.

docs/agents/ifm_sbn246_flowmeter.rst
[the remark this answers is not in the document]
BrianJKoopman: I didn't change anything for the IoT ports network settings, so unless you did I think it's DHCP by default. It's sort of hard to tell from the documentation, which says:

> To configure the IP settings with DHCP, a DHCP server has to be active in the IT network. If
> no DHCP server can be reached in the IT network, an IP address is automatically assigned to
> the IoT port with the Zeroconfig protocol (address range: → Factory settings (→ p. 91)).

If it is DHCP by default, then yeah, we don't need to run the LR Software and it should just get an address and work. Can we test by factory resetting the device we have in the lab?

socs/agents/ifm_sbn246_flowmeter/agent.py
```diff
@@ -0,0 +1,172 @@
...
+
+        return True, 'Stopping acq process'
+
+
+def add_agent_args(parser_in=None):
+    if parser_in is None:
+        parser_in = argparse.ArgumentParser()
+    pgroup = parser_in.add_argument_group('Agent Options')
+    pgroup.add_argument("--ip-address", type=str, help="IP address of DAQ IO-Link device.")
+    pgroup.add_argument("--daq-port", type=int, help="Port on DAQ IO-Link device that IFM is connected to")
```
BrianJKoopman, line 148: "DAQ IO-Link device" -> "IO-Link master"

Also, "that IFM is connected to" -> "that the flowmeter is connected to"

socs/agents/ifm_sbn246_flowmeter/agent.py
```diff
@@ -0,0 +1,172 @@
...
+                                 agg_params=agg_params,
+                                 buffer_time=1
+                                 )
+
+    @ocs_agent.param('test_mode', default=False, type=bool)
+    def acq(self, session, params=None):
+        """
+        acq(test_mode=False)
+
+        **Process** - Fetch values from the flowmeter using the DAQ device
```
BrianJKoopman, line 78: "using the DAQ device" -> "using the IO-Link master."

socs/agents/ifm_sbn246_flowmeter/agent.py
```diff
@@ -0,0 +1,172 @@
...
+    Monitor the flowmeter.
+
+    Parameters
+    ----------
+    agent : OCSAgent
+        OCSAgent object which forms this Agent
+    ip_address: str
+        IP address of DAQ IO-Link device to make requests from
+    daq_port: int
+        Port on daq IO device that connects to flowmeter. Choices are 1-4
```
BrianJKoopman, line 44: More "DAQ IO-Link device" -> "IO-Link master"

docs/agents/ifm_sbn246_flowmeter.rst
```diff
@@ -0,0 +1,92 @@
...
+
+IO-Link Master Network Setup
+````````````````````````````````
+After plugging into the IoT port on your IO-Link master, the IP address of the IO-Link
+device can be configured using the LR Device software. Once the LR Device software
+is downloaded, click the 'read from device' on the upper right (leftmost IOLINK
+button). Once the software has found your IO-Link device, click on the device
+name (AL1340) on the column on the left side (not the ports that may say P1,
+P2... or the name of the device connected to any of the 4 ports on your IO-Link
+master). You can now change  the IP address by clicking the Network (DHCP) button.
```
BrianJKoopman, line 84: Still a few things here:

- The IP address is for the IO-Link master, not the individual devices.
- "click on the device name" is tricky because of the keywords "master" and "device" here, especially since we really want the "master" in this case. Also, do we need to even find the devices in order to change the IP address of the master?
- Extra space in the last sentence.

BrianJKoopman (changes requested):
The updates looks good, thanks! Still a few wording changes needed. Maybe we can clarify the network setup too after we test the process with the factory reset.

> I don't know the final number of flowmeters that will be in use for the site. 2 per cooling loop? More? I know there is a desire for Oriel to buy an IFM pressure gauge to hook into the IO-Link master, but I don't know if that's been done yet or how finalized that is.Thinking about it today, it may be worth considering writing an agent for the IO-Link device—there are cables long enough that you'd be able to connect multiple flowmeters into one IO-Link device. It does sort of make sense to go that way? And if we're possibly adding an IFM pressure gauge.

Ah, yeah I imagined there would be multiple flowmeters per AL1340, since there are 4 ports and the various flowmeters probably aren't that far apart from each other.

I can see it either way. Having an Agent per device is simple, but could result in adding a lot of agents (I remember just seeing a box full of AL1340's at the site.) Having an agent for the AL1340 reduces the number of agents, but adds a bit of complexity.

Thinking about it real quick though for each different type of device we add the AL1340 agent could detect the device type and then know how to handle the process data.

> I added the check for the model number on initialization; not sure if you want the assertion error message to be a `self.log` message also? Added in the `session.data` output into the `acq()` docstring too. Let me know if the `session.data` structure should be different.

Those both look good, thanks for adding them.

docs/agents/ifm_sbn246_flowmeter.rst
sanahabhimani: Yeah do tell me if you get to the software today before I do, but I think when you hover your mouse over the button that finds the IO-Link master, it says "read from device," which if true, is confusing for this documentation, I agree. 

Definitely the software doesn't need to find the devices plugged into the master; think it just does that anyway when it finds the IO-Link master. But my inclination was that we need to click that "read from device" button so that the software can actually find the IO-Link master and tell us its ip.

BrianJKoopman (approved):
Thanks! Looks good. I made a few small updates to the docs.

The test that is failing is unrelated, going to go ahead and merge.

PR #446, opened 2023-05-22
Updates to ibootbar Agent

### LINKED ISSUE #417: ibootbar -- lock some outlets
Can we add an option to annotate outlets in the ibootbar/synaccess agent so that they can be treated slightly differently (especially by ocs-web)?  The goal is to make it a bit difficult for someone to power off critical network infrastructure remotely (leading to an unrecoverable outage).  This could be a passive notation (the Agent simply provides information about what outlets are critical to network) or explicit prevention (the Agent provides that information but also rejects state change requests on those outlets) -- for ocs-web it's pretty much the same.

I'm imagining either a config file or simply cmdline args like "--lock-outlet login2-1 --lock-outlet switch3".  The config file might be desirable in the longer term, when we start having "access levels" and the ability to escalate privileges.  But cmdline args would be great for now.

### LINKED ISSUE #437: ibootbar agent requests
Some minor quality-of-life improvement requests:

1. When SNMP isn't working, the "update_cache" function records that connected=False, along with the timestamp.  However this information isn't propagated back to the acq process (session.data).  This info would be valuable for ocs-web.

2. Please add the IP or hostname of the target device to acq's session.data.

3. Please make the --mode arg optional, by setting default="acq".

4. When the acq loop starts, transition to status = 'running'.  (Someday this might be done automatically/internally, but for now has to be done at the top of all ops.)

(Also, #417 is still important, but a bit more work than the above.)

davidvng:
## Description

Various updates to the ibootbar agent. Includes logic change to handling connection drops, which is the same with UPS agent. Created new task `lock_outlet` that can lock/unlock a specified outlet. If an outlet is locked, other tasks such as `set_outlet` will fail. Agent can be initialized with locked outlets using args. For example, `["--lock_outlet", [1, 2, 3]]` in `default.yaml` or `--lock-outlet 1 2 3` on command line will startup agent with outlets 1, 2, and 3 locked.

## Motivation and Context

Resolves #417, resolves #437.

## How Has This Been Tested?

Tested on ibootbar at Yale.

## Types of changes

## Checklist:


mhasself:
I see this is a draft, but two immediate reactions:
- I think the presence of locked outlets need not prevent a "reboot".  In most cases, outlets will be set to reboot into their current state.
- It is essential that we can lock outlets on agent startup.  Either a config file or simply "--lock-outlet 1 --lock-outlet 3 --lock-outlet 8" or something like that.

Thanks!

socs/agents/ibootbar/agent.py
```diff
@@ -146,7 +151,7 @@ def update_cache(get_result, names, timestamp):
     except TypeError:
         oid_cache['ibootbar_connection'] = {'last_attempt': time.time(),
                                             'connected': False}
-        raise ConnectionError('No SNMP response. Check your connection.')
+        raise TypeError('Unable to interpret SNMP response.')
```
mhasself, line 154: The oid_cache is not returned if you raise an error.  So there's no point in setting it inside this "except" clause.

You need to decide whether you'll report errors by raising exceptions, or through the return value.

mhasself (changes requested):
Thank you!  I think there's still an issue with propagating session.data in some failure modes, see comment.

tests/integration/test_ibootbar_agent_integration.py
BrianJKoopman, line 1: Great to see added tests! You could also add a locked outlet to the config file so that it's locked at Agent startup, which would test lines 191-195, as well as toggle the lock off, which would test lines 440-441.

BrianJKoopman (changes requested):
Awesome, just a couple small suggestions from me.

mhasself (changes requested):
Thanks for these changes!

A few simple notes inline.  Also:  I see you have hard-coded the number 60, representing the sampling loop period in seconds, in a bunch of different places.  This is also used to modify lastGet, to trigger updates.  That "60" should become a constant declared at the top of the program.
-

mhasself:
Sorry, one more thing while you're in there -- can you add the "locked" information to acq's session.data?  This will help to make the display in ocs-web more informative.  E.g.
```
        "outletStatus_0": {"status": 1,                                                                        
                           "name": Outlet-1,
                           "locked": False,                                                                   
                           "description": "on"}
```

socs/agents/ibootbar/agent.py
```diff
@@ -316,7 +344,7 @@ def set_outlet(self, session, params=None):
         self.lastGet = self.lastGet - 60
 
         return True, 'Set outlet {} to {}'.\
-            format(params['outlet'] - 1, params['state'])
+            format(params['outlet'], params['state'])
 
     @ocs_agent.param('outlet', choices=[1, 2, 3, 4, 5, 6, 7, 8])
     @ocs_agent.param('cycle_time', default=10, type=int)
```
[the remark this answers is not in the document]
davidvng: The MIB file states that the OID for setting cycle time is type Integer32. I've tried setting it to float, and the ibootbar seems to automatically round to the nearest int and still cycle the outlet, however the task itself crashes due to type error.

socs/agents/ibootbar/agent.py
```diff
@@ -198,6 +201,7 @@ def __init__(self, agent, address, port=161, version=2, lock_outlet=None):
         self.snmp = SNMPTwister(address, port)
 
         self.lastGet = 0
+        self.sample_rate = 60
```
mhasself, line 204: Meaning of "period" vs "rate" is important for clarity.  Please change to sample_period.

socs/agents/ibootbar/agent.py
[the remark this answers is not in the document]
mhasself: Ah, ok.  Thanks for re-checking.  Int it is then!

mhasself (approved):
Thanks for those fixes.  I'm very happy with this now, except the rate -> period terminology issue where noted.

BrianJKoopman (changes requested):
The only thing from me is the testing changes we discussed in person @davidvng. To summarize those:

* Re-add test for "set state on".
* Test cycle task locks out when outlet locked.
* Make sure test names represent what they're testing (i.e. `test_ibootbar_lock_outlet` currently is more of a test of a unlocking an outlet.)

PR #447, opened 2023-05-23
ACU Exercisor mode

mhasself:
## Description

ACU Agent has an ability to command itself through a series of test motions, indefinitely.  The current motion type and a count of any errors are published to a feed.

## Motivation and Context

This has been useful for testing SATPs at site.  It is not the most polished code but I think for completeness it would be good to merge it into main line so it be run in the main OCS account and not in my screen session... 

## How Has This Been Tested?

This evolved from an independent client code that commanded the platforms through test motions for the past few months.  But since late April it has been run, in its current form, for many hours, on the SATPs at site.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -173,6 +173,7 @@ class ACUAgent:
         if exercise_plan:
             agent.register_process(
                 'exercise', self.exercise, self._simple_process_stop)
+            # Use longer default frame length ... very low volume feed.
             self.agent.register_feed('activity',
                                      record=True,
                                      buffer_time=0,
@@ -245,7 +246,8 @@ class ACUAgent:
               ...
             },
             "StatusResponseRate": 19.237531827325963,
-            "PlatformType": "satp"
+            "PlatformType": "satp",
+            "connected": True,
           }
 
         In the case of an SATP, the Status3rdAxis is not populated
@@ -1419,7 +1421,7 @@ class ACUAgent:
 
     @ocs_agent.param('starting_index', type=int, default=0)
     def exercise(self, session, params):
-        """exercise()
+        """exercise(starting_index=0)
 
         **Process** - Run telescope platform through some pre-defined motions.
 
@@ -1544,8 +1546,8 @@ def add_agent_args(parser_in=None):
     if parser_in is None:
         parser_in = argparse.ArgumentParser()
     pgroup = parser_in.add_argument_group('Agent Options')
-    pgroup.add_argument("--acu_config")
-    pgroup.add_argument("--exercise_plan")
+    pgroup.add_argument("--acu-config")
+    pgroup.add_argument("--exercise-plan")
     return parser_in
 
 
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -543,9 +543,10 @@ class ACUAgent:
                               }
         return True, 'Acquisition exited cleanly.'
 
+    @ocs_agent.param('auto_enable', type=bool, default=True)
     @inlineCallbacks
     def broadcast(self, session, params):
-        """broadcast()
+        """broadcast(auto_enable=True)
 
         **Process** - Read UDP data from the port specified by
         self.acu_config, decode it, and publish to HK feeds.  Full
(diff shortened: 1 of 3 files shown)


socs/agents/acu/agent.py
```diff
@@ -236,15 +236,29 @@ def monitor(self, session, params):
...
+        session.data = {'PlatformType': self.acu_config['platform'],
+                        'connected': False}
+
+        last_complaint = 0
+        while True:
+            try:
+                version = yield self.acu_read.http.Version()
+                break
+            except Exception as e:
+                if time.time() - last_complaint > 3600:
```
BrianJKoopman, line 274: Is complaining every hour about a dropout often enough?

socs/agents/acu/agent.py
```diff
@@ -236,15 +236,29 @@ def monitor(self, session, params):
...
-        self.log.info(version)
 
         # Note that session.data will get scanned, to assign data to
         # feed blocks.  Items in session.data that are themselves
         # dicts will parsed; but items (such as PlatformType and
         # StatusResponseRate) which are simple strings or floats will
         # be ignored for feed assignment.
-        session.data = {'PlatformType': self.acu_config['platform']}
+        session.data = {'PlatformType': self.acu_config['platform'],
+                        'connected': False}
```
BrianJKoopman, line 266: Add 'connected' field to the `session.data` example in the docstring.

socs/agents/acu/agent.py
```diff
@@ -1377,12 +1398,135 @@ def _run_track(self, session, point_gen, step_time, azonly=False,
...
+
+        return _exit_now(True, "Stopped run process")
+
 
 def add_agent_args(parser_in=None):
     if parser_in is None:
         parser_in = argparse.ArgumentParser()
     pgroup = parser_in.add_argument_group('Agent Options')
     pgroup.add_argument("--acu_config")
+    pgroup.add_argument("--exercise_plan")
```
BrianJKoopman, line 1548: The exercise plan arg is documented in the agent reference page as `--exercise-plan` with a hyphen. I see an underscore is probably used for consistency with `--acu_config`, but I think that should also use a hyphen for consistency across all of socs.

I'm fine with however you want to handle this at the moment, but should be consistent within this PR.

socs/agents/acu/agent.py
```diff
@@ -160,6 +169,17 @@ def __init__(self, agent, acu_config='guess'):
...
 
+        # Automatic exercise program...
+        if exercise_plan:
+            agent.register_process(
+                'exercise', self.exercise, self._simple_process_stop)
+            self.agent.register_feed('activity',
+                                     record=True,
+                                     buffer_time=0,
+                                     agg_params={
+                                         'frame_length': 600,
```
BrianJKoopman, line 180: Just to point out this frame length is 10 times longer than other feeds in this agent register with.

socs/agents/acu/agent.py
```diff
@@ -1359,12 +1417,135 @@ def _run_track(self, session, point_gen, step_time, azonly=False,
 
         return True, 'Track ended cleanly'
 
+    @ocs_agent.param('starting_index', type=int, default=0)
+    def exercise(self, session, params):
+        """exercise()
```
BrianJKoopman, line 1422: `exercise()` -> `exercise(starting_index=0)`

BrianJKoopman (changes requested):
Just a few small docstring related change requests, and some questions/comments.

socs/agents/acu/agent.py
mhasself: I think so, because during that time the session.data will accurately reflect connected=False, and the monitor data won't be updating... no motion commands will be executable.  It will be pretty obvious that the monitoring is not working.

socs/agents/acu/agent.py
mhasself: Added a note.

mhasself:
@BrianJKoopman Thanks -- I think I've processed your comments in the latest commit.

BrianJKoopman:
Great, I'm onboard with changing to `--acu-config`, but be aware you'll need to update the site config files that currently use `--acu_config` when updating to the latest version of this Agent.

I pushed a commit that catches some docs that still used `--acu_config`.

BrianJKoopman (approved):

mhasself:
> Great, I'm onboard with changing to `--acu-config`, but be aware you'll need to update the site config files that currently use `--acu_config` when updating to the latest version of this Agent.

Already done ...

> I pushed a commit that catches some docs that still used `--acu_config`.

Ah, right ... thanks!

PR #456, opened 2023-06-09
Suprsync CLI

jlashner:
This moves the suprsync script in `bin` to an socs entrypoint. It also adds an additional sub-command to add files from a local directory to the suprsync db.

Since the timecode dir finalization has already been reviewed, I figured we could review this in its own PR, and then merge both together to main  once tested on the e2e system.

## Motivation and Context

This is a nice user interface for checking the status of the suprsync db. The `add-local-files` command is useful for things like setting up hk sync or for the e2e testing system.

## How Has This Been Tested?

Tested locally.

## Types of changes

## Checklist:

### DIFF
--- a/socs/db/suprsync_cli.py
+++ b/socs/db/suprsync_cli.py
@@ -85,7 +85,7 @@ def main():
     add_local_files_parser = subparsers.add_parser(
         'add-local-files',
         help="Adds all files from a local directory to the suprsync db"
-    ) 
+    )
     add_local_files_parser.set_defaults(func=add_local_files_func)
     add_local_files_parser.add_argument(
         'local_root', help="Root directory to search for files")
--- a/socs/db/suprsync_cli.py
+++ b/socs/db/suprsync_cli.py
@@ -5,8 +5,9 @@ Utility script for interacting with the suprsync db.
 import argparse
 import os
 
-from socs.db.suprsync import SupRsyncFile, SupRsyncFilesManager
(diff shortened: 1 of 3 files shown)


socs/db/suprsync_cli.py
```diff
@@ -41,17 +41,30 @@ def list_func(args):
...
+    srfm = SupRsyncFilesManager(args.db, create_all=args.create_db)
+    known_files = srfm.get_known_files(args.archive_name)
+    known_paths = [f.local_path for f in known_files]
+    for root, _, files in os.walk(args.local_root):
+        for file in files:
+            path = os.path.join(root, file)
+            path = os.path.abspath(path)
+            if path not in known_paths:
+                remote_path = os.path.relpath(path, args.local_root)
+                srfm.add_file(path, remote_path, args.archive_name)
```
mhasself, line 54: Is there any checking on local_root, to ensure that it's the same local root used in the past with this DB?

For codes that work with the filesystem and a database, it's helpful to have a --dry-run switch -- which would cause the code to print out what it _would do_ without actually doing it.  In this case it would allow one to see all the files the config thinks are new; a good way to catch usage mistakes during initial setup (such as passing local_root="/path/to" instead of local_root="/path/to/stuff").

jlashner: Good idea, I added a `--dry` argument which will just print out some debugging info and up to 10 files that will be added to the database

Issue #455, opened 2023-06-07
PMX voltage and current flag monitoring


jlashner:
We need functionality in the PMX agent that allows us to monitor whether or not the voltage and current limit flags have been triggered. It would also be very nice to have a function in the PMX agent that lets us reset the flags, or a supervisor task that cycles power to the PMX using synaccess.

ykyohei:
Sound good. I started implementing them. We'll merge after testing it.
https://github.com/simonsobs/socs/tree/hwp_pmx

ykyohei:
We can query the error code and the error message, 
for example `(0, 'No error')`,  `(103,  'Operation denied while in PROTection state')`

Which level of monitoring is preferred in socs?
- publish only the error code to both influxDB and aggregator
- publish only the error code to both influxDB, error code and message to aggregator
- etc

PR #463, opened 2023-06-15
Add protection mode monitoring to pmx agent

d1ssk:
This adds protection mode and error monitoring to the pmx agent and adds task to clear protection mode.

## Description
With this change, the pmx agent will minitor for error and check if the pmx is in protection mode for every acq runs.
Information about errors and protection mode will be output to data along with voltage and current.
`clear_alarm()` method of the pmx agent can be used to exit the protection mode.

## Motivation and Context
#455 

## How Has This Been Tested?
Tested with chwp-daq pc in UTokyo

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -324,6 +324,7 @@ def main(args=None):
     agent.register_process('acq', PMX.acq, PMX._stop_acq)
     agent.register_task('set_on', PMX.set_on)
     agent.register_task('set_off', PMX.set_off)
+    agent.register_task('clear_alarm', PMX.clear_alarm)
     agent.register_task('set_i', PMX.set_i)
     agent.register_task('set_v', PMX.set_v)
     agent.register_task('set_i_lim', PMX.set_i_lim)
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -107,7 +107,7 @@ class HWPPMXAgent:
                     'Could not clear alarm because {} is already running'.format(self.lock.job))
                 return False, 'Could not acquire lock'
             self.dev.clear_alarm()
-            self.prot_mode = 0
+            self.prot = 0
         return True, 'Clear alarm'
 
     @ocs_agent.param('curr', default=0, type=float, check=lambda x: 0 <= x <= 3)
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -235,10 +235,16 @@ class HWPPMXAgent:
                 data['data']['current'] = curr
                 msg, volt = self.dev.meas_voltage()
                 data['data']['voltage'] = volt
+                msg, code = self.dev.check_error()
+                data['data']['err_code'] = code
+                data['data']['err_msg'] = msg
+                prot = self.dev.check_prot()
+                data['data']['prot'] = prot
 
                 self.agent.publish_to_feed('hwppmx', data)
                 session.data = {'curr': curr,
                                 'volt': volt,
+                                'prot': prot,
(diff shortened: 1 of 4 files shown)


d1ssk:
In testing, the last line of `socs/tests/integration/test_hwp_pmx_agent_integration.py` seems to cause error `6 == 5`.
Does anyone know why? @ykyohei @jlashner 
```
@pytest.mark.integtest
def test_hwp_rotation_acq(wait_for_crossbar, kikusui_emu, run_agent, client):
    responses = {'meas:volt?': '2',
                 'meas:curr?': '1'}
    kikusui_emu.define_responses(responses)

    client.init_connection.wait()  # wait for connection to be made
    resp = client.acq.start(test_mode=True)
    assert resp.status == ocs.OK

    resp = client.acq.wait(timeout=20)
    assert resp.status == ocs.OK
    assert resp.session['op_code'] == OpCode.SUCCEEDED.value
```

BrianJKoopman:
> In testing, the last line of `socs/tests/integration/test_hwp_pmx_agent_integration.py` seems to cause error `6 == 5`. Does anyone know why?

`6` represents a failure of the `acq` operation and `5` represents a successfully completed operation. (See the [OpCode Enum docs](https://ocs.readthedocs.io/en/main/api.html#ocs.base.OpCode).)

This is happening because the tests need to know how to respond to the new commands you're sending. The first one hit is when calling `check_error()`, that sends this command:
```
self.sock.sendall(b':system:error?\n')
```

The responses from the emulated hardware are defined per test, the `test_hwp_rotation_acq` test does that here:
https://github.com/simonsobs/socs/blob/b117bc54cb4cc52329fe927cfb6336f6af2f4d26/tests/integration/test_hwp_pmx_agent_integration.py#L149-L151

What you need to do is determine how the actual hardware responds to the `'system:error?'` command, and define a reasonable value in this response dict in the test.

The same will need to be done for the other added commands like `'stat:ques?'` and `'output:protection:clear'`.

There's a little bit of testing documentation here: https://socs.readthedocs.io/en/main/developer/testing.html#writing-tests

You can run the tests locally by going into the `tests/` directory and running:
```
$ python3 -m pytest integration/test_hwp_pmx_agent_integration.py -s
```

You can further restrict to a single test with `-k 'test_hwp_rotation_acq'`. If the device emulator receives a command it doesn't know the response to it'll print:
```
msg=':system:error?'
encountered error ':system:error?'
```

d1ssk:
Thank you for the detailed explanation! @BrianJKoopman 
I corrected that point.

socs/agents/hwp_pmx/agent.py
```diff
@@ -235,10 +250,19 @@ def acq(self, session, params):
                 data['data']['current'] = curr
                 msg, volt = self.dev.meas_voltage()
                 data['data']['voltage'] = volt
+                msg, code = self.dev.check_error()
+                data['data']['err_code'] = code
+                data['data']['err_msg'] = msg
+                prot_code = self.dev.check_prot()
+                if prot_code != 0:
+                    self.prot = prot_code
```
jlashner, line 258: I'm not sure if I fully understand how to interpret the prot_code, but why isn't `self.prot` reset to 0 if `check_prot` returns 0?

jlashner (approved):
I have one question inline, but this seems great and very useful. Thanks!!

socs/agents/hwp_pmx/agent.py
d1ssk: As a PMX specification, once the protection status code is read, the code is reset to 0 even though the protection mode is not actually cleared. The reason for this code is to continue to output the same error code unless clear_alarm() is done via the pmx agent.

d1ssk:
OK, I wil add documentation.

jlashner:
Looks great, thanks!

PR #467, opened 2023-06-26
Suprsync archive stats

jlashner:
## Description
This adds additional stats for each suprsync archive, which are added to session.data along with hk feeds. This can be used by G3tSmurf to determine the finalization time for each archive, and for grafana monitoring.

## Motivation and Context
Required for G3tSmurf to determine when an archive is complete and books can be created. Also useful for monitoring in grafana.

## How Has This Been Tested?

Tested by running the agent locally with fake data.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/suprsync/agent.py
+++ b/socs/agents/suprsync/agent.py
@@ -72,6 +72,7 @@ class SupRsync:
                                  agg_params={
                                      'exclude_aggregator': True,
                                  })
+        self.agent.register_feed('archive_stats', record=True)
 
     def run(self, session, params=None):
         """run()
@@ -148,8 +149,21 @@ class SupRsync:
                 self.log.error("rsync returned non-zero exit code! {e}", e=e)
                 op['error'] = 'nonzero exit'
                 counters['errors_nonzero'] += 1
-
+            
             now = time.time()
+
+            archive_stats = srfm.get_archive_stats()
+            for archive_name, stats in archive_stats.items():
+                self.agent.publish_to_feed('archive_stats', {
+                    'block_name': archive_name,
+                    'timestamp': now,
+                    'data': {
+                        f'{archive_name}_{k}': v
+                        for k, v in stats.items()
+                    }
+                })
+            session.data['archive_stats'] = archive_stats
+
             op['stop_time'] = now
             session.data['last_copy'] = op
             session.data['timestamp'] = now
--- a/socs/agents/suprsync/agent.py
+++ b/socs/agents/suprsync/agent.py
@@ -94,13 +94,13 @@ class SupRsync:
                     "files": [],
                     "stop_time": 1661284493.5398622
                   },
-                  'archive_stats': {
-                    'smurf': {
-                        'finalized_until': 1687797424.652119,
-                        'num_files': 3,
-                        'uncopied_files': 0,
-                        'last_file_added': '/path/to/file',
-                        'last_file_copied': '/path/to/file'
+                  "archive_stats": {
+                    "smurf": {
+                        "finalized_until": 1687797424.652119,
+                        "num_files": 3,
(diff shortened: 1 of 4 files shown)


socs/agents/suprsync/agent.py
```diff
@@ -93,6 +94,15 @@ def run(self, session, params=None):
                     "files": [],
                     "stop_time": 1661284493.5398622
                   },
+                  'archive_stats': {
+                    'smurf': {
```
kmharrington, line 98: will this always say "smurf" or is it "smurf" and "timestamps"?

kmharrington (approved):
I think this all looks good, I'd like to try building the G3tSmurf Updates off of this.

socs/db/suprsync.py
```diff
@@ -208,6 +212,70 @@ def __init__(self, db_path, create_all=True, echo=False):
...
+
+        query = session.query(SupRsyncFile).filter(
+            SupRsyncFile.archive_name == archive_name,
+        ).order_by(asc(SupRsyncFile.timestamp))
+
+        for file in query.all():
+            if file.ignore:
+                continue
+            if file.local_md5sum != file.remote_md5sum:
+                return file.timestamp - 1
```
BrianJKoopman, line 275: How close are timestamps between files generally? Is there any chance a "finalized" file has a timestamp that falls within the one second before the first failed file?

You could track the latest "good" timestamp and return that value, though then code that uses this needs to know that the timestamp means all files up to `t` are good. It seems like right now if a file has a matching timestamp it would mean it isn't finalized.

socs/db/suprsync.py
```diff
@@ -208,6 +212,70 @@ def __init__(self, db_path, create_all=True, echo=False):
...
+                'last_file_copied': last_file_copied,
+            }
+
+        return stats
+
+    def get_finalized_until(self, archive_name, session=None):
+        """
+        Returns a timetamp for which all files preceding are either successfully
+        copied, or ignored. If all files are copied, returns the current time.
+        """
```
BrianJKoopman, line 263: Docstring could use an `Args:` list.

socs/db/suprsync.py
```diff
@@ -208,6 +212,70 @@ def __init__(self, db_path, create_all=True, echo=False):
...
+                last_file_added = f.local_path
+                if (f.local_md5sum == f.remote_md5sum):
+                    last_file_copied = f.local_path
+
+                if (not f.ignore) and (f.local_md5sum != f.remote_md5sum):
+                    num_files_to_copy += 1
+
+                if finalized_until is None and not (f.ignore):
+                    if f.local_md5sum != f.remote_md5sum:
+                        finalized_until = f.timestamp - 1
```
BrianJKoopman, line 243: Same question here, are file guaranteed to be spaced out enough that `timestamp - 1` is a good cut off for "finalized until" times?

socs/db/suprsync.py
```diff
@@ -208,6 +212,70 @@ def __init__(self, db_path, create_all=True, echo=False):
         if create_all:
             Base.metadata.create_all(self._engine)
 
+    def get_archive_stats(self, session=None):
```
BrianJKoopman, line 215: I understand why this makes a dict that includes all archive names, but I just want to point out that the only use (outside of tests) is:
```
srfm.get_archive_stats().get(self.archive_name)
```

Which makes me think that it would make sense to have `get_archive_stats()` take an argument for `archive_name` like:
```
srfm.get_archive_stats(self.archive_name)
```

BrianJKoopman (commented):
This looks good, the one thing I'm wondering about is the use of `timestamp - 1` for a `finalized_until` time. Are files in an archive always at least 1 second apart?

BrianJKoopman (approved):

PR #473, opened 2023-07-12
Changes to allow for reading more than one continuous block of resgis…

DaveBoettger:
…ters at a time

Support reading of bit or bit range within individual registers needed for reading i.e. alarm conditions from the generator controller Configuration file loading switched to read an entire directory of .yaml files at a time

## Types of changes

## Checklist:


socs/agents/generator/config.yaml
davidvng, line 1: Although the configs won't live here, please include an example file that we can use as a template. Without one, I am unable to test this code since I don't know how the config files have been changed.

socs/agents/generator/agent.py
```diff
@@ -281,7 +381,7 @@ def acq(self, session, params=None):
 
             session.data = {"fields": {}}
 
-            pm = Pacemaker(1, quantize=True)
+            pm = Pacemaker(.1)
```
davidvng, line 384: This is a relatively low sample rate, is there a reason why we want this?

davidvng (changes requested):
Hi @DaveBoettger, thanks for the PR! I've added some quick comments below.

Also a few questions/concerns regarding the config file:

- From my understanding, you're reading in multiple config files that essentially have the same content, with only the `read_len` and `read_start` being changed. Is this correct? If so, is there any way to reduce the amount of repetition across the multiple config files?
- I do think I'll need to see example config files to be able to test this out, unless you've already tested on a generator yourself.

socs/agents/generator/agent.py
jlashner: I believe Packemaker accepts a sampling frequency, so this is 10 Hz

socs/agents/generator/agent.py
```diff
@@ -306,13 +406,13 @@ def acq(self, session, params=None):
...
                     regdata = self.generator.read_cycle()
-
                     if regdata:
                         for reg in regdata:
                             data['data'][reg] = regdata[reg]["value"]
-                            field_dict = {reg: regdata[reg]}
+                            field_dict = {reg: regdata[reg]['value']}
                             session.data['fields'].update(field_dict)
                         session.data.update({'timestamp': current_time})
+                        print(data)
```
jlashner, line 415: I assume this was for debugging, so we should remove

socs/agents/generator/agent.py
```diff
@@ -169,7 +267,7 @@ class GeneratorAgent:
         txaio logger object, created by the OCSAgent
     """
 
-    def __init__(self, agent, host='localhost', port=5021):
+    def __init__(self, agent, configdir, host='localhost', port=5021):
```
jlashner, line 270: Can you update the docstring to reflect new arg?

jlashner (changes requested):
I haven't gone deep into the cfg parsing, but changes to the agent look good to me, with a few small inline comments. I agree with David that adding a description (and example) of what the cfg file looks to the Generator docs page would be very helpful

socs/agents/generator/agent.py
```diff
@@ -1,5 +1,6 @@
 import argparse
 import os
+import os.path
```
jlashner, line 3: I'm pretty sure we don't need this if `os` is already imported

DaveBoettger: Agreed, removed.

socs/agents/generator/agent.py
DaveBoettger: Done.

socs/agents/generator/config.yaml
DaveBoettger: I've added some example config files.

jlashner (approved):
Made a couple of really small naming changes, and made sure docs can build. I think this is now good to go!

davidvng (approved):

PR #476, opened 2023-07-18
Separate generator fields into different blocks

jlashner:
Separates generator fields into separate blocks.

## Description

Sometimes the generator is not able to provide data for certain fields, in which case we don't want those fields to be published to grafana. The current version will simply remove those fields from the block, which starts throwing errors due to block structures not matching. To fix this, this PR separates fields into different blocks since they may not be co-sampled.

## Motivation and Context

Buxfix for current implementation of the generator agent.

## How Has This Been Tested?

Not tested yet

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/generator/agent.py
+++ b/socs/agents/generator/agent.py
@@ -424,7 +424,13 @@ class GeneratorAgent:
                     self.log.info('Trying to reconnect.')
                     continue
 
-                self.agent.publish_to_feed('generator', data)
+                for field, val in data['data'].items():
+                    _data = {
+                        'timestamp': current_time,
+                        'block_name': field,
+                        'data': {field: val}
+                    }
+                    self.agent.publish_to_feed('generator', _data)
 
             self.agent.feeds['generator'].flush_buffer()
 


socs/agents/generator/agent.py
```diff
@@ -424,7 +424,13 @@ def acq(self, session, params=None):
                     self.log.info('Trying to reconnect.')
                     continue
 
-                self.agent.publish_to_feed('generator', data)
+                for field, val in data['data'].items():
+                    _data = {
+                        'timestamp': current_time,
+                        'block_name': field,
+                        'data': {field: val}
```
davidvng, line 431: Just curious, does `val` return None here if the generator can't provide the data?

davidvng (approved):
Looks good to me!

socs/agents/generator/agent.py
jlashner: previously, the field that couldn't return data would just not be included in the data block, which is what caused errors when trying to publish.

davidvng: right, I see it now, the data block isn't updated because reading the register returns None on lines 169-170. makes sense!

PR #477, opened 2023-07-18
bump pyyaml

jlashner:
Docs builds are failing with the error described here: https://github.com/yaml/pyyaml/issues/601

Solution may be to bump the pyyaml version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ pyasn1==0.4.8
 
 # common dependencies - used by multiple agents
 numpy
-markupsafe==2.0.1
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
Bumping pyyaml is causing conflict with `pytest-docker-compose`, which depends on `docker-compose`. `docker-compose` via pip is actually a quite old docker-compose v1, which is deprecated, but nevertheless pins pyyaml<6, which is why this conflict is happening.

I think we'll need to drop `pytest-docker-compose`. More details in #482 as a draft right now. Closing this, thanks for starting the debugging!

PR #501, opened 2023-08-10
Bumps suprsync table version

jlashner:
Bumps the table version in suprsync to account for the `ignore` column added in https://github.com/simonsobs/socs/pull/467. 

## Description

Currently, if you try to run suprsync on an existing db file without altering the table as is shown [here](https://github.com/simonsobs/daq-discussions/discussions/27) the agent will crash. bumping the table version (and changing the table name) should prevent this from crashing while keeping data that was written before the change. Note that any un-copied files that were added before bumping the version will not be copied on their own.

## How Has This Been Tested?
has not been tested yet

## Types of changes

## Checklist:

### DIFF
--- a/socs/db/suprsync.py
+++ b/socs/db/suprsync.py
@@ -12,7 +12,7 @@ from sqlalchemy.orm import sessionmaker
 
 from socs.util import get_md5sum
 
-TABLE_VERSION = 0
+TABLE_VERSION = 1
 txaio.use_twisted()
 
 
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):

BrianJKoopman:
@jlashner, I think I held off on merging this to wait for testing. What's the status there? Should we go ahead and merge?

jlashner:
I am fine with closing this... We have already altered table version 0 on existing systems. I think a better way to go about this might be to provide script or function that alters tables if specific columns don't exist (at least for small changes in the table structure).

BrianJKoopman:
Great, closing.

It might be too much if these are just small changes, but I know on some SO projects folks use [Alembic](https://alembic.sqlalchemy.org/en/latest/) for managing those sorts of database updates.

PR #508, opened 2023-08-16
Adds archive_name filter to update timecode func

### LINKED ISSUE #507: Suprsync update_timecode_dirs not checking the archive name
Currently, the `update_all_timecode_dirs` does not filter by archive name: https://github.com/simonsobs/socs/blob/74cf6816b57138c70a35448a2828c463aa2a63b4/socs/db/suprsync.py#L446-L450

This means that it is possible for, say, the smurf-sync agent to "finalize" tcdirs in the `timestreams` archive. This happened on smurf-server so3, and now there are a bunch of finalization files that look like this:
```
/data/so/suprsync/16922/daq-sync-smurf-so3/1692207266_timestreams_16655_finalized.yaml
``` 
mismatched in that this is a finalization file for the timestreams archive, but it is in the suprsync folder with instance-id `daq-sync-smurf-so3`. ~Also... the finalization file has the archive_name `smurf` instead of `timestreams`~ (not sure if this is true).

I think this can be fixed by only updating / finalizing tcdirs with a matching archive name, but I'm wondering if this will mess up bookbinding for existing data.

jlashner:
Adds a filter on TimecodeDir.archive_name when updating tcdirs

## Description

Resolves #507 

## Motivation and Context

See issue #507 

## How Has This Been Tested?

not yet, but simple change.

## Types of changes

## Checklist:

### DIFF
--- a/socs/db/suprsync.py
+++ b/socs/db/suprsync.py
@@ -445,7 +445,9 @@ class SupRsyncFilesManager:
 
     def update_all_timecode_dirs(self, archive_name, file_root, sync_id):
         with self.Session.begin() as session:
-            tcdirs = session.query(TimecodeDir).all()
+            tcdirs = session.query(TimecodeDir).filter(
+                TimecodeDir.archive_name == archive_name,
+            ).all()
             for tcdir in tcdirs:
                 self._update_tcdir(tcdir, session, file_root, sync_id)
 


jlashner:
Tested on smurfsrv20

BrianJKoopman (commented):
This looks OK to me. Have you checked if this affects bookbinding for existing data? You expressed that concern in #507.

jlashner:
My concern was that bookbinding and the imprinter would not run properly without this fix, but it seems to be running ok... I don't think this change can break anything that is currently working in the bookbinder.

BrianJKoopman (approved):

PR #511, opened 2023-08-18
Add OCSClient crossbar address override to wg kikusui agent

BrianJKoopman:
## Description

This adds the ability for the Wiregrid Kikusui agent to pull the `SITE_HTTP` environment variable and use it for the `OCSClient` connection to the wiregrid encoder agent. It also adds an argument to the agent that allows overriding the env var.

## Motivation and Context

At site the docker host running the wiregrid kikusui agent has a host wide configuration in the SCF for the docker network based address for the crossbar server, while the container runs in network mode host and uses the `SITE_HTTP` env var to override the host wide config, allowing the kikusui agent to start.

However, it then crashes when trying to connect a client to interact with the encoder agent because that client connection tries to use the host wide configuration pointing to a name that can't be resolved.

As I write this I realize this could probably also be fixed by putting the internal docker domain name for the crossbar server in the hosts file...I'm going to try that too before merging this.

## How Has This Been Tested?

This was tested at site.

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/wiregrid_kikusui.rst
+++ b/docs/agents/wiregrid_kikusui.rst
@@ -42,6 +42,8 @@ An example site-config-file block::
   Communicating device is determined by the ethernet port number of the converter.)
 - encoder-agent is an instance ID of the wiregrid encoder agent (wiregrid-encoder).
   This is necessary to get the position recorded by the encoder for controlling the rotation.
+- crossbar-http is available (but not shown here) and can be used to set the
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
This is no longer really needed. I solved this problem by adding the internal crossbar network's domain name for the crossbar server to the host systems `/etc/hosts` file. This hosts file is used within all containers and allows resolution of the crossbar address for the OCSClient to connect to even when the container is in `network_mode: host`.

PR #521, opened 2023-09-05
SupRsync Stats Bug

kmharrington:
## Description
Found file trying to figure out what's going on the the LAT finalization at the site

## Motivation and Context
Looks like we were adding up the non-copied files in the way the logic actually works and then weren't actually reporting it. Using this `num_files_to_copy` variable will make the logic on what's going on with the finalization time match the number of files we need to copy number. Note, this won't actually fix the current issue with the LAT times.

## How Has This Been Tested?
I haven't actually tested it because I'm not sure how

## Types of changes

## Checklist:

### DIFF
--- a/socs/db/suprsync.py
+++ b/socs/db/suprsync.py
@@ -244,7 +244,7 @@ class SupRsyncFilesManager:
         stats = {
             'finalized_until': finalized_until,
             'num_files': len(files),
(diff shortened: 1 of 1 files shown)


jlashner (approved):
Thanks katie, this definitely looks right. This seems simple enough that it's probably fine to merge without testing...

PR #524, opened 2023-09-08
Small change to make vantage agent more robust to hardware dropouts

jlashner:
Some small changes to make sure the Vantage agent is robust to hardware dropouts. 

Unfortunately Weather station boots into a mode where it's uninitialized and returns junk data until someone holds down "Done" for two seconds... @RemingtonGerras do you have any idea if it's possible to automatically bypass this? We should also install batteries in the console regardless to try to avoid drop outs.

## Description

Makes the `acq` process attempts to re-initizlize connection on timeout

## Motivation and Context

Required so `acq` can recover after hardware dropout

## How Has This Been Tested?

Tested with hardware at site

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/vantagepro2/agent.py
+++ b/socs/agents/vantagepro2/agent.py
@@ -142,7 +142,6 @@ class VantagePro2Agent:
                 except TimeoutError as e:
                     self.log.warn(f"TimeoutError: {e}")
                     self.module = None
-                    time.sleep(30)
                     continue
 
                 self.agent.publish_to_feed('weather_data', data)
--- a/socs/agents/vantagepro2/agent.py
+++ b/socs/agents/vantagepro2/agent.py
@@ -46,7 +46,7 @@ class VantagePro2Agent:
         self.agent.register_feed('weather_data',
                                  record=True,
                                  agg_params=agg_params)
-    
+
     def _initialize_module(self):
         """Initialize the VantagePro2 module."""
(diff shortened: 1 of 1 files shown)


socs/agents/vantagepro2/agent.py
```diff
@@ -120,12 +126,25 @@ def acq(self, session, params=None):
...
                     'block_name': 'weather',
                     'data': {}
                 }
-                data['data'] = self.module.weather_daq()
+                try:
+                    data['data'] = self.module.weather_daq()
+                except TimeoutError as e:
+                    self.log.warn(f"TimeoutError: {e}")
+                    self.module = None
+                    time.sleep(30)
```
BrianJKoopman, line 145: Maybe if the connection has dropped we just try again right away to initialize then let the above 30 second sleep take effect?

BrianJKoopman (commented):
My only comment is on the 30 second wait. I'd probably err more on trying to connect more quickly. I know you mentioned if the device reboots it requires some manual intervention, but how does it handle network outages?

jlashner:
Thanks! Removed the 30s wait after timeout is thrown. I think network outages are fine, the console doesn't go into a shutdown state if the network disconnects, and with these changes the agent should be able to re-connect smoothly when it comes back online.

BrianJKoopman (approved):

PR #533, opened 2023-09-27
set acq session status appropriately

sanahabhimani:
## Description
Same as #528: Setting the acq process status to 'running' once a connection is established before the acq loop begins.

## Motivation and Context
So the agent doesn't remain in 'start' mode, which triggers alarms

## How Has This Been Tested?
It hasn't.

## Types of changes

## Checklist:


BrianJKoopman (approved):
Thanks!

Issue #531, opened 2023-09-23
Vantage agent changes


jlashner:
A few changes and additions we'd like in the vantage agent based on current usage: 
- UVI is off by a factor of ten
- UVI overflows if it goes above 127. We should have the agent automatically convert negative UVIs to the value they would be if they were an unsigned int.
- Temp is in fahrenheit but we'd like celsius
- We'd like an estimate of the wind-chill / wind-chill-temp. We may be able to just do this in grafana with existing data

BrianJKoopman:
Mostly solved by #539. If we're fine with estimating wind-chill / wind-chill-temp in Grafana this can be closed.

BrianJKoopman:
Wind chill was added in #550. I had to revert the UVI factor of 10 fix in #636, due to the fact that since an integer was already published, InfluxDB rejects the new float values. I think it would be fine to continue to publish integers and just divide by 10 for display in Grafana. Let me know if that's still a problem.

PR #539, opened 2023-10-06
Addressed (most of) Issue #531

RemingtonGerras:
## Description

- UVI is no longer off by a factor of 10 (divided by 10)
- UVI overflow issue is fixed by adding 2**7 to negative (overflowed) values
- Changed temperature from Fahrenheit to Celsius

## Motivation and Context

[Issue #531](https://github.com/simonsobs/socs/issues/531)

## How Has This Been Tested?

I ran python in terminal and created functions that take dummy input values and output the changed value. 
These functions reflected the changes I made in the agent.

## Types of changes

## Checklist:

There are comments in code that should be changed (instances of 'fahrenheit' to 'celsius')

### DIFF
--- a/socs/agents/vantagepro2/drivers.py
+++ b/socs/agents/vantagepro2/drivers.py
@@ -241,27 +241,27 @@ class VantagePro2:
         loop_data['packet_type'] = byte_data[4]
         loop_data['next_record'] = byte_data[5]
         loop_data['barometer'] = byte_data[6] / 1000.0
-        loop_data['temp_inside'] = byte_data[7] / 10.0
+        loop_data['temp_inside'] = (byte_data[7] / 10.0 - 32) * (5 / 9)
         loop_data['humidity_inside'] = byte_data[8]
-        loop_data['temp_outside'] = byte_data[9] / 10.0
+        loop_data['temp_outside'] = (byte_data[9] / 10.0 - 32) * (5 / 9)
         loop_data['wind_speed'] = byte_data[10]
         loop_data['avg_wind_speed'] = byte_data[11]
         loop_data['wind_dir'] = byte_data[12]
-        loop_data['extra_temp0'] = byte_data[13] - 90.0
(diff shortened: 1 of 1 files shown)


socs/agents/vantagepro2/drivers.py
```diff
@@ -312,6 +312,13 @@ def receive_data(self):
         loop_data['time_sunrise'] = byte_data[75]
         loop_data['time_sunset'] = byte_data[76]
 
+        # Correct UV Index by a factor of 10 and fix overflow
+        uvi = loop_data['UV']
+        if uvi < 0:
+            uvi += 2**7
```
mhasself, line 318: Great to see ... this should be 2**8 = 256, right?  i.e., -127 becomes +129.

RemingtonGerras: oops yes that makes sense

socs/agents/vantagepro2/drivers.py
```diff
@@ -241,27 +241,27 @@ def receive_data(self):
         loop_data['packet_type'] = byte_data[4]
         loop_data['next_record'] = byte_data[5]
         loop_data['barometer'] = byte_data[6] / 1000.0
-        loop_data['temp_inside'] = byte_data[7] / 10.0
+        loop_data['temp_inside'] = (byte_data[7] / 10.0 - 32) * (5 / 9)
```
mhasself, line 244: Please write a function that converts F to C, and use it on all of these.  This is like the classic example of when should use a function.

mhasself (changes requested):
Thanks, UV fix looks good.  Please use function for F_to_C.  It is then self-documenting and less prone to little typos.

socs/agents/vantagepro2/drivers.py
RemingtonGerras: you're right :(

mhasself (approved):

PR #541, opened 2023-10-10
Hwp site changes

bbixler500:
Changes to the hwp_gripper, hwp_pid, and hwp_pmx agents to improve functionality at the site

## Description

- hwp_gripper: fixed how arguments are passed to the agent and renamed a function for consistency
- hwp_pid: changed the socket communication to prevent agent crashes
- hwp_pmx: changed the socket communication to prevent agent crashes

## Motivation and Context

As it is currently, the hwp_gripper agent does not work. The changes made to that agent were made to get it running. The hwp_pid and hwp_pmx agents crash if the network communication is temporarily lost. The changes made to those agents are to prevent that crashing.

## How Has This Been Tested?

All three of these agents have been running on the dev environment overnight with no issue. The changes to the hwp_gripper agent are minor and only effect how the agent is initialized. The changes to the hwp_pid and hwp_pmx agents only change the socket communication between agent and device, and the agents have been continually sending data to grafana successfully.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pid/drivers/pid_controller.py
+++ b/socs/agents/hwp_pid/drivers/pid_controller.py
@@ -32,7 +32,7 @@ class PID:
         self.set_direction('0')
 
     @staticmethod
-    def _establish_connection(ip, port, timeout=5):
+    def _establish_connection(ip, port, timeout=2):
         """Connect to PID controller.
 
         Args:
@@ -54,9 +54,8 @@ class PID:
             try:
                 conn.connect((ip, port))
                 break
-            except (ConnectionRefusedError, OSError) as e:
+            except (ConnectionRefusedError, OSError):
                 print(f"Failed to connect to device at {ip}:{port}")
-            time.sleep(5)
         else:
             raise RuntimeError('Could not connect to PID controller')
         return conn
@@ -267,7 +266,7 @@ class PID:
                 time.sleep(0.5)  # Don't send messages too quickly
                 data = self.conn.recv(4096).decode().strip()
                 return data
-            except (socket.timeout, OSError) as e:
+            except (socket.timeout, OSError):
                 print("Caught timeout waiting for response from PID controller. "
                       + "Trying again...")
                 time.sleep(1)
--- a/socs/agents/hwp_pid/drivers/pid_controller.py
+++ b/socs/agents/hwp_pid/drivers/pid_controller.py
@@ -49,13 +49,16 @@ class PID:
         conn = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         conn.settimeout(timeout)
         # unit tests might fail on first connection attempt
-        while True:
+        attempts = 3
+        for attempt in range(attempts):
             try:
                 conn.connect((ip, port))
                 break
             except (ConnectionRefusedError, OSError) as e:
                 print(f"Failed to connect to device at {ip}:{port}")
             time.sleep(5)
+        else:
+            raise RuntimeError('Could not connect to PID controller')
         return conn
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


socs/agents/hwp_pid/drivers/pid_controller.py
```diff
@@ -45,18 +47,15 @@ def _establish_connection(ip, port, timeout=5):
 
         """
         conn = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        conn.settimeout(timeout)
         # unit tests might fail on first connection attempt
-        attempts = 3
-        for attempt in range(attempts):
+        while True:
```
jlashner, line 52: I don't love this being an infinite loop, since control is never passed back to the agent. Can we keep this to a finite number of attempts, and handle re-connection at the agent level?

jlashner (changes requested):
This looks ok to me for the most part, however I don't agree with the new `_establish_connection` logic. I think this needs to fail after a certain number of attempts, instead of entering an infinite loop.

davidvng:
Hi @bbixler500 , this fix looks great! Over at SATP3 we've been having trouble with our PID and gripper agents, and I've tested this branch and they work. I'd greatly appreciate getting this merged soon.

bbixler500:
At the site there was a strange behavior where the PID and PMX agents would consistently crash every night, requiring me to restart the agents every morning. These crashes occurred because of connection interruptions, which could last up to 3hrs (I was never able to determine the cause of the connection interruptions). Previously the agents were coded so that `_establish_connection` would throw an error after a set number of unsuccessful connections. Consequently, the number of unsuccessful attempts was always reached and the agent would always crash. The switch to the infinite loop solved the agents crashing issue.

davidvng:
> At the site there was a strange behavior where the PID and PMX agents would consistently crash every night, requiring me to restart the agents every morning. These crashes occurred because of connection interruptions, which could last up to 3hrs (I was never able to determine the cause of the connection interruptions). Previously the agents were coded so that `_establish_connection` would throw an error after a set number of unsuccessful connections. Consequently, the number of unsuccessful attempts was always reached and the agent would always crash. The switch to the infinite loop solved the agents crashing issue.

I think you can let the acq process be the "infinite loop" here: https://github.com/simonsobs/socs/blob/hwp_site_changes/socs/agents/hwp_pid/agent.py#L286
Then, whenever [these calls](https://github.com/simonsobs/socs/blob/fa26efa881e25498490bd64240281e71b4ea2ca9/socs/agents/hwp_pid/agent.py#L298-L300) fail, you can call `init_connection` again. I think this is what @jlashner meant by reconnecting at the agent level.

davidvng:
@bbixler500 this change looks good to me. If you could fix the pytest failure and pre-commit failure, we can get this approved and merged.

bbixler500:
I'm not sure what this pytest error means or how to solve it. It appears to be from the `test_hwp_pmx_agent_integration.py` file

socs/agents/hwp_pmx/drivers/PMX_ethernet.py
```diff
@@ -51,103 +78,95 @@ def check_output(self):
...
         """ Set PMX to ignore external voltage """
-        self.sock.sendall(b'volt:ext:sour none\n')
+        self.send_message(b'volt:ext:sour none\n', read=False)
         self.wait()
         return self.check_source()
 
     def set_current_limit(self, curr_lim):
         """ Set the PMX current limit """
-        self.sock.sendall(b'curr:prot %a\n' % curr_lim)
+        self.send_message(b'curr:prot %a\n' % curr_lim)
```
davidvng, line 159: do you need `read=False` here? I think this is what's causing the test to fail.

davidvng:
@jlashner @BrianJKoopman can we get this reviewed and merged this week? SATP3 would like to spin the HWP this weekend. thanks!

jlashner (approved):
Looks good to me! Thanks for the changes @bbixler500

PR #547, opened 2023-10-11
Improve ACU scanning fault tolerance and logging

mhasself:
## Description

A few important things pertinent to constant elevation scanning:
- Limit the turnaround acceleration according to the maximum jerk constraint described in the SATP ICD.
- Log the parameters passed in by the user, and computed params.  For forensic tracking on failed scans.
- Monitor key program track error bits, and state them in the log if they fire during scan.
- Return False from the Process, if it appears to have exited due to error.
- Clear Faults before starting a scan (-- and also before a go_to).

## Motivation and Context

These are all to make real usage smoother, and forensic analysis possible (why didn't the scan several hours ago work?).  And the max jerk constraint basically came out because it was very easy to trigger a "Turnaround time too short" fault.

## How Has This Been Tested?

I developed this actively on SATP2.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -1076,6 +1076,10 @@ class ACUAgent:
             if not acquired:
                 return False, f"Operation failed: {self.azel_lock.job} is running."
 
+            self.log.info('Clearing faults to prepare for motion.')
+            yield self.acu_control.clear_faults()
+            yield dsleep(1)
+
             ok, msg = yield self._check_ready_motion(session)
             if not ok:
                 return False, msg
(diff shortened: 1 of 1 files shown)


mhasself:
We will want this running on SATP1 before we next do overnight scans ...

socs/agents/acu/agent.py
```diff
@@ -1076,6 +1076,10 @@ def go_to(self, session, params):
             if not acquired:
                 return False, f"Operation failed: {self.azel_lock.job} is running."
 
+            self.log.info('Clearing faults to prepare for motion.')
+            yield self.acu_control.clear_faults()
+            yield dsleep(1)
+
```
BrianJKoopman, line 1082: Should we have a similar block before boresight motion?

BrianJKoopman (commented):
This looks good to me, just wondering if we should clear faults before boresight motion as well.

socs/agents/acu/agent.py
mhasself: yes, yes we should ...

BrianJKoopman (approved):
Great, thanks for adding. I'll merge once checks finish.

PR #550, opened 2023-10-17
Vantage windchill

RemingtonGerras:
Added wind chill function and report the wind chill temperature during acq process.

## Description

Above :)

## Motivation and Context

[531](https://github.com/simonsobs/socs/issues/531) (continued)
## How Has This Been Tested?

Put functions into a python terminal, checked results with various online calculators. 

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/vantagepro2/drivers.py
+++ b/socs/agents/vantagepro2/drivers.py
@@ -1,6 +1,7 @@
 import array as arr
 import struct
 import time
+import numpy as np
 
 from serial import Serial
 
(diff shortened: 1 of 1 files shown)


socs/agents/vantagepro2/drivers.py
```diff
@@ -58,6 +59,19 @@ def F_to_C(temp):
     return (temp - 32) * (5 / 9)
 
 
+def wind_chill(temp, wind):
+    """Function to calculate wind chill temperature. Only valid if temp < 50F.
+    Taken from https://www.calculator.net/wind-chill-calculator.html
+    If temp > 50F, need to use heat index instead...
```
BrianJKoopman, line 65: If this formula is not valid above 50F can you add a check for the input temperature and just return the input temp?

Just playing around with the function as is it looks like it might report strange numbers otherwise:
```
>>> # expect -5F windchill
>>> wind_chill(5, 5)
-4.6271537104386
>>>
>>> # what about at higher temps?
>>> wind_chill(105, 5)
112.82872791458931
```

BrianJKoopman (changes requested):
Thanks @RemingtonGerras! Just one suggestion about the 50F cutoff.

BrianJKoopman (approved):
I added that temperature limit. Will merge when checks complete.

PR #554, opened 2023-10-25
ACU Agent -- two features (LAT support)

mhasself:
## Description

1. By passing `--ignore-axis=...`, the agent will decline to command one or more axes, and pretend that all moves on the axis succeed.  This is useful in cases where an axis needs to be disabled for some reason, but you still want to run scans and stuff.
2. The "idle stow" feature of LAT ACU requires using the command interface every minute, to prevent Survival Mode.  This renames the process that does it, and makes sure to start that process if platform is LAT type.  Can be disabled with cmdline option.

## Motivation and Context

Both of these came up on LAT last week.

These changes will remove the need for me to run the LAT agent out of my user dir ... agent config should include args `--ignore-axis el` for now.

## How Has This Been Tested?

I've tested these on SATP2.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -54,13 +54,12 @@ class ACUAgent:
         startup (bool):
             If True, immediately start the main monitoring processes
             for status and UDP data.
-        ignore_axes (str or list of str):
+        ignore_axes (list of str):
             List of axes to "ignore". "ignore" means that the axis
             will not be commanded.  If a user requests an action that
             would otherwise move the axis, it is not moved but the
             action is assumed to have succeeded.  The values in this
-            list should be drawn from "az", "el", and "third".  A single
-            comma-delimited string (e.g. "az,el" is also accepted.
+            list should be drawn from "az", "el", and "third".
         disable_idle_reset (bool):
             If True, don't auto-start idle_reset process for LAT.
 
@@ -96,8 +95,6 @@ class ACUAgent:
         startup_idle_reset = (self.acu_config['platform'] in ['lat', 'ccat']
                               and not disable_idle_reset)
 
-        if isinstance(ignore_axes, str):
-            ignore_axes = [x.strip() for x in ignore_axes.split(',')]
         if ignore_axes is None:
             ignore_axes = []
         assert all([x in ['az', 'el', 'third'] for x in ignore_axes])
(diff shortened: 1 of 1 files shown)


socs/agents/acu/agent.py
```diff
@@ -1765,6 +1850,11 @@ def add_agent_args(parser_in=None):
     pgroup.add_argument("--exercise-plan")
     pgroup.add_argument("--no-processes", action='store_true',
                         default=False)
+    pgroup.add_argument("--ignore-axes",
+                        help="Comma-delimited list of axes to ignore "
+                        "(el, az, third).")
```
BrianJKoopman, line 1855: This can be limited to only accept "el", "az", and "third" by adding `choices=["el", "az", "third"], nargs="+"`. Passing args then looks more like this though:
```
--ignore-axes "el" "az" "third"
```

socs/agents/acu/agent.py
```diff
@@ -85,6 +93,18 @@ def __init__(self, agent, acu_config='guess', exercise_plan=None,
...
         self._set_default_scan_params()
 
+        startup_idle_reset = (self.acu_config['platform'] in ['lat', 'ccat']
+                              and not disable_idle_reset)
+
+        if isinstance(ignore_axes, str):
+            ignore_axes = [x.strip() for x in ignore_axes.split(',')]
+        if ignore_axes is None:
+            ignore_axes = []
+        assert all([x in ['az', 'el', 'third'] for x in ignore_axes])
```
BrianJKoopman, line 103: If `ignore_axes` isn't restricted in the argparse, perhaps put this in a try/except to catch the `AssertionError` and reraise with a useful error like `f"One or more of {ignore_axes} not in ['az', 'el', 'third']"`.

socs/agents/acu/agent.py
```diff
@@ -306,6 +336,7 @@ def monitor(self, session, params):
         session.data = {'PlatformType': self.acu_config['platform'],
                         'DefaultScanParams': self.scan_params,
                         'StatusResponseRate': 0.,
+                        'IgnoredAxes': self.ignore_axes,
```
BrianJKoopman, line 339: Add to session.data example in docstring.

BrianJKoopman (changes requested):
This looks good to me, a few small suggestions below.

socs/agents/acu/agent.py
mhasself: done

socs/agents/acu/agent.py
mhasself: Ya, ok ... I am suspicious of this syntax but I guess I can't find a logical problem with it.

BrianJKoopman (approved):
Thanks for the updates!

PR #558, opened 2023-10-28
Meinberg Syncbox Agent

### LINKED ISSUE #518: Meinberg Syncbox N2X Agent
The Syncbox N2X has an SNMP interface we can monitor.

davidvng:
## Description

Adds new agent to receive data from the Meinberg syncbox via SNMP.

## Motivation and Context

Resolves #518 .

## How Has This Been Tested?

Tested on `daq-dev` with syncboxes available on site.

## Types of changes

## Checklist:


davidvng:
@mhasself putting this on your radar as it seems relevant again. this had worked last year and I'm wondering if it'd be useful to update this and push for review?

docs/agents/meinberg_syncbox_agent.rst
```diff
@@ -0,0 +1,99 @@
...
+      - LOGLEVEL=info
+
+
+The ``LOGLEVEL`` environment variable can be used to set the log level for
+debugging. The default level is "info".
+
+Description
+-----------
+The Meinberg syncbox synchronizes to GPS and distributes timing mostly over
+the network using PTP.
```
mhasself, line 65: The M1000 does this, not the syncbox.

The syncbox syncs from PTP, and distributes signals to attached devices in various formats (IRIG, PPS).

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,513 @@
...
+        'block_name': blockname,
+        'timestamp': time,
+        'data': {}
+    }
+
+    for item in get_result:
+        field_name, oid_value, oid_description = _extract_oid_field_and_value(item)
+
+        if oid_value is None:
+            continue
```
mhasself, line 94: This construction can lead to invalid data blocks.  Once the set of fields has been written to a block once, those same fields always need to be written together (no new fields, no omissions).

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,513 @@
...
+                          accurateToWithin10ms(43),
+                          accurateToWithin25ms(44),
+                          accurateToWithin100ms(45),
+                          accurateToWithin250ms(46),
+                          accurateToWithin1s(47),
+                          accurateToWithin10s(48),
+                          accurateToGreaterThan10s(49)
+        """
+
+        session.set_status('running')
```
mhasself, line 348: No longer needed.

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,513 @@
...
+                self.log.info('Trying to reconnect.')
+
+            read_time = time.time()
+
+            # Check if sample period has passed before getting status
+            if (read_time - self.lastGet) < self.sample_period:
+                continue
+
+            main_get_list = []
+            get_list = []
```
mhasself, line 363: This is confusing.... why two lists?  It looks like all things (or almost all things?) are "get"ted twice.

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,513 @@
...
+
+            # Create the list of OIDs to send get commands
+            oids = ['mbgSyncboxN2XSerialNumber',
```
mhasself, line 366: This should not be inside the loop.  Probably it should be a global, class, or instance var.

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,513 @@
...
+                    'mbgSyncboxN2XPtpGrandmasterTimesource',
+                    'mbgSyncboxN2XPtpGrandmasterPriority1',
+                    'mbgSyncboxN2XPtpGrandmasterClockClass',
+                    'mbgSyncboxN2XPtpGrandmasterClockAccuracy',
+                    'mbgSyncboxN2XPtpGrandmasterClockVariance',
+                    'mbgSyncboxN2XPtpOffsetToGrandmaster',
+                    'mbgSyncboxN2XPtpMeanPathDelay']
+
+            for oid in oids:
+                main_get_list.append(('MBG-SYNCBOX-N2X-MIB', oid, 0))
```
mhasself, line 388: The string "MBG-SYNCBOX-N2X-MIB" is a constant; declare it at global / class / instance level and reuse.

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,492 @@
...
+        'block_name': 'syncbox',
+        'timestamp': time,
+        'data': {}
+    }
+
+    for item in get_result:
+        field_name, oid_value, oid_description = _extract_oid_field_and_value(item)
+
+        if oid_value is None:
+            oid_value = 'No SNMP response'
```
mhasself, line 92: But the usual value can be integer or float for some fields, right?  In that case you can't mix strings in (influx hates that ... and probably aggregator won't like it either).

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,492 @@
...
+            for i in range(outputs):
+                for oid in self.output_oids:
+                    get_list.append([(self.mib, oid, i)])
+
+            # Issue SNMP GET command
+            # The syncbox has a unique case this requires issuing GET commands
+            # one by one or else it will return the same data for each OID
+            get_result = yield self.snmp.get(get_list[0], self.version)
+            for get in get_list[1:]:
+                result = yield self.snmp.get(get, self.version)
```
mhasself, line 400: Why is the first get outside the loop?

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,492 @@
...
+                if result is None:
+                    self.connected = False
+                    session.data['syncbox_connection'] = {'last_attempt': time.time(),
+                                                          'connected': False}
+                    continue
```
mhasself, line 405: This "continue" applies to the "for get" loop.  Is that what you intend?

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -0,0 +1,492 @@
...
+                self.log.error('No SNMP response. Check your connection!')
+                self.log.info('Trying to reconnect.')
+
+            read_time = time.time()
+
+            # Check if sample period has passed before getting status
+            if (read_time - self.lastGet) < self.sample_period:
+                continue
+
+            get_list = []
```
mhasself, line 384: Again, it is confusing for this to be inside the loop.  The get_list should be static for the runtime of the agent.

mhasself (changes requested):
Thanks for the quick fixes -- still some issues with error handling I think.

I think the only way forward here is to make a list of expected fields (you have done that); then to query those on each iteration, and if *any* fields are missing, don't publish them and mark self as degraded.

If the "outputs" list is actually dynamic (depends on syncbox config), then you can have each output assigned to a different ocs feed -- that way it won't be an error for those to go in and out depending on syncbox configuration.

A better way, given that we want our syncbox to be in a particular configuration, would be to have command line args that _specify_ which outputs should be configured.  Then the syncbox can only monitor those, and become degraded if the status is not available.

davidvng:
> I think the only way forward here is to make a list of expected fields (you have done that); then to query those on each iteration, and if _any_ fields are missing, don't publish them and mark self as degraded.

Should I not update `session.data` as well as not publish to feed when this happens? And if so, do I need to do something like `session.set_status('stopping')`?

BrianJKoopman:
> Latest version looks good.

This seems good to merge, yeah? @davidvng any other changes incoming?

davidvng:
> > Latest version looks good.
> 
> This seems good to merge, yeah? @davidvng any other changes incoming?

@BrianJKoopman no other changes from me, I think it's good to merge

BrianJKoopman (approved):

PR #559, opened 2023-11-01
ACU Sun Avoidance

mhasself:
## Description

The ACU Agent monitors and reports the Sun position.  If active_avoidance is enabled, it will reject non-Sun-safe moves, and even slew to safe position if idle and the Sun gets too close.  "Too close" is controllable, to some extent, through command line args.

Note this currently the changes from #554.

## Motivation and Context

SATPs do not have built-in Sun Avoidance.  But we might run this on the LAT too, to have more fine-grained / easily configurable control.

## How Has This Been Tested?

I developed this on SATP2, and ran through a number of different situations.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -88,9 +88,9 @@ class ACUAgent:
             elevation lower limit.
         max_el (float): If not None, override the default configured
             elevation upper limit.
-        solar_avoidance (bool): If set, override the default Sun
+        avoid_sun (bool): If set, override the default Sun
             avoidance setting (i.e. force enable or disable the feature).
-        avoidance_radius (float): If set, override the default Sun
+        fov_radius (float): If set, override the default Sun
             avoidance radius (i.e. the radius of the field of view, in
             degrees, to use for Sun avoidance purposes).
 
@@ -99,7 +99,7 @@ class ACUAgent:
     def __init__(self, agent, acu_config='guess', exercise_plan=None,
                  startup=False, ignore_axes=None, disable_idle_reset=False,
                  min_el=None, max_el=None,
-                 sun_avoidance=None, avoidance_radius=None):
+                 avoid_sun=None, fov_radius=None):
         self.log = agent.log
 
         # Separate locks for exclusive access to az/el, and boresight motions.
@@ -144,8 +144,8 @@ class ACUAgent:
         if len(self.ignore_axes):
             agent.log.warn('User requested ignore_axes={i}', i=self.ignore_axes)
 
-        self.reset_sun_params(enabled=sun_avoidance,
-                              radius=avoidance_radius)
+        self.reset_sun_params(enabled=avoid_sun,
+                              radius=fov_radius)
 
         self.exercise_plan = exercise_plan
 
@@ -191,8 +191,8 @@ class ACUAgent:
                                self._simple_process_stop,
                                blocking=False,
                                startup=startup)
-        agent.register_process('solar_avoidance',
-                               self.solar_avoidance,
+        agent.register_process('monitor_sun',
+                               self.monitor_sun,
                                self._simple_process_stop,
                                blocking=False,
                                startup=startup)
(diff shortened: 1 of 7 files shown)


mhasself:
The ocs-web widget, coming soon, will look something like this:
image

BrianJKoopman:
Now that #554 is merged, I think a rebase would be good here. (This'll test whether I lose the couple of comments I have pending too...)

mhasself:
> Now that #554 is merged, I think a rebase would be good here. (This'll test whether I lose the couple of comments I have pending too...)

Yup, will do ... also one more commit coming...

BrianJKoopman (commented):
With the caveat that I don't really have a setup to test this in, this is looking good. One thing I would love to see (though let me know if it's going to hold this up) is some fixed time (i.e. repeatable) examples via some unit tests of the `avoidance.py` module.

Other requests are mostly documentation related below.

mhasself:
> With the caveat that I don't really have a setup to test this in, this is looking good. One thing I would love to see (though let me know if it's going to hold this up) is some fixed time (i.e. repeatable) examples via some unit tests of the `avoidance.py` module.
> 
> Other requests are mostly documentation related below.

Thanks -- I think I've dealt with that all.

I actually did have slightly better docs, and some tests, but somehow omitted those commits on my initial PR.  But I took another pass at it in any case.

socs/agents/acu/agent.py
```diff
@@ -2087,8 +2087,7 @@ def _notify_recomputed(result):
     @ocs_agent.param('avoidance_radius', type=float, default=None)
     @ocs_agent.param('shift_sun_hours', type=float, default=None)
     def update_sun(self, session, params):
-        """update_sun(reset, enable, temporary_disable, escape,
-                      avoidance_radius, shift_sun_hours)
+        """update_sun(reset, enable, temporary_disable, escape, avoidance_radius, shift_sun_hours)
```
BrianJKoopman, line 2090: Line wraps are possible here with a `\` if needed. (I noticed the wrapped line to unwrapped line change in the few commits.) (See the edit on the stream docstring in [this PR](https://github.com/simonsobs/socs/pull/553/files) for example.)

socs/agents/acu/avoidance.py
```diff
@@ -0,0 +1,685 @@
...
+        If raw=True, a debugging output is returned; see code.
+
+        Returns:
+          best_move (dict): The element of moves that is safest.  If
+            no safe move was found, None is returned.
+          decisions (list): The items in this list are dicts that
+            correspond one-to-one with the entries in moves.  Each
+            decision dict has entries 'rejected' (True or False) and
+            'reason' (string description of why the move was rejected
+            outright).
```
BrianJKoopman, line 539: The autodoc formatting for this is not quite right. It's rendering like:
image

The syntax is a bit different than for arguments, it's:
```
Returns:
    type: Description text.
    
    Multiple lines can be used indented to the same level as the first line.
```

So something like this would detect the type properly:
```
        Returns:
          (dict, list): (best_move, decisions)

          ``best_move`` -- the element of moves that is safest.  If no safe
          move was found, None is returned.

          ``decisions`` -- The items in this list are dicts that correspond
          one-to-one with the entries in moves.  Each decision dict has entries
          'rejected' (True or False) and 'reason' (string description of why
          the move was rejected outright).
```

This renders like:
image

BrianJKoopman (commented):
Great, thanks for adding the tests and updating the docs! Two small docs formatting suggestions, but otherwise this is ready to go.

I pushed a commit (https://github.com/simonsobs/socs/pull/559/commits/330be8644e35f51f5ec9be5f001cba674d698b1b) that includes the new dependencies in the build process and documents how to install them as well.

socs/agents/acu/avoidance.py
mhasself: Ah, great ... will do that.

mhasself:
Definitely squash this if you want ... lots of commits, mostly noise ...

BrianJKoopman (approved):

PR #563, opened 2023-11-07
PWV HTTP Requests Error 

sanahabhimani:
## Description
Addressing this error for the PWV agent:

`Unable to format event {'log_logger': <Logger 'ocs.ocs_agent.OCSAgent'>,
'log_level': <LogLevel=info>, 'log_namespace': 'ocs.ocs_agent. OCSAgent',
'10g_SOle': None, 'log_format': 'acq:0 CRASH: [Failure instance: Traceback: <class \'requests .exceptions .JSONDecodeError\'>: [Errno Expecting value] {"pwv": nan, "timestamp": nan}: `

## Motivation and Context
I'm 10% unsure that this is truly an HTTP requests failure because I don't know if I should expect/assume a 204 error to output `nan`'s. Implementing a try/except to loop through and make the http request again. Added a `self.log.info` message to query the requests status code the next time this happens in the agent to know more.

## How Has This Been Tested?
Hasn't been tested on the flask server at the site itself. 

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/ucsc_radiometer/agent.py
+++ b/socs/agents/ucsc_radiometer/agent.py
@@ -91,9 +91,7 @@ class UCSCRadiometerAgent:
                 self.last_published_reading = (last_pwv, last_timestamp)
 
             session.data = {"timestamp": last_timestamp,
-                            "pwv": {}}
-
-            session.data['pwv'] = last_pwv
+                            "pwv": last_pwv}
 
             if params['test_mode']:
                 break
--- a/socs/agents/ucsc_radiometer/agent.py
+++ b/socs/agents/ucsc_radiometer/agent.py
@@ -72,7 +72,8 @@ class UCSCRadiometerAgent:
                 r = requests.get(self.url)
             except ValueError:
                 self.log.info(f'Query to data failed due to request error: {r.status_code}')
-                pass
+                pm.sleep()
+                continue
             data = r.json()
             last_pwv = data['pwv']
             last_timestamp = data['timestamp']
--- a/socs/agents/ucsc_radiometer/agent.py
+++ b/socs/agents/ucsc_radiometer/agent.py
@@ -83,12 +83,9 @@ class UCSCRadiometerAgent:
                     }
 
             if self.last_published_reading is not None:
-                try:
-                    if last_timestamp > self.last_published_reading[1]:
-                        self.agent.publish_to_feed('pwvs', pwvs)
-                        self.last_published_reading = (last_pwv, last_timestamp)
-                except ValueError:
-                    pass
+                if last_timestamp > self.last_published_reading[1]:
+                    self.agent.publish_to_feed('pwvs', pwvs)
+                    self.last_published_reading = (last_pwv, last_timestamp)
             else:
                 self.agent.publish_to_feed('pwvs', pwvs)
                 self.last_published_reading = (last_pwv, last_timestamp)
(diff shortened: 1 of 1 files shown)


socs/agents/ucsc_radiometer/agent.py
```diff
@@ -68,7 +68,11 @@ def acq(self, session, params=None):
         session.set_status('running')
 
         while self.take_data:
-            r = requests.get(self.url)
+            try:
+                r = requests.get(self.url)
+            except ValueError:
+                self.log.info(f'Query to data failed due to request error: {r.status_code}')
+                pass
```
mhasself, line 75: If a ValueError is raised, this will log a message _and then continue on to execute the code below_.  If "r" has a value, e.g. from the previous loop iteration, that value will be recorded in the data stream.  This is duplicate data.

I think intead of "pass" here, you want a `pm.sleep()` and `continue`.

socs/agents/ucsc_radiometer/agent.py
```diff
@@ -68,7 +68,11 @@ def acq(self, session, params=None):
         session.set_status('running')
 
         while self.take_data:
-            r = requests.get(self.url)
+            try:
+                r = requests.get(self.url)
+            except ValueError:
```
mhasself, line 73: Will this catch the JSONDecodeError?  I see:
```
>>> import requests
>>> isinstance(requests.JSONDecodeError, ValueError)
False
```

mhasself (changes requested):
Forgive the random review but this caught my eye!  Couple other comments:

1. The JSONDecodeError you're seeing is because the server is working (it sends a json-like structure) but is sending "json" in a (non-)standard that the requests library doesn't like.  Specifically it's representing NaN values as "nan".  Probably you want to ignore those data, anyway, so it's fine to just catch the error like you're doing.

2. Later in the function, you have this:
```
            session.data = {"timestamp": last_timestamp,
                            "pwv": {}}
            session.data['pwv'] = last_pwv
```
That's a bit of an odd construction, to have pwv sometimes be an empty dict and sometimes be a float. It could cause trouble for anyone trying to use the session.data.  Can you change it to simply:
```
            session.data = {"timestamp": last_timestamp,
                            "pwv": last_pwv}
```
?

BrianJKoopman (commented):
Thanks for the review @mhasself. I just wanted to mention what I think is going on on the backend here. Looking at the most recent two crashes, this happens at the first query after 00:00 -- `2023-11-05T00:00:08+0000` and `2023-11-07T00:00:10+0000` are the two that I'm seeing.

The server that this agent queries takes the median of the last four data points, and I expect that code is running into some error when there are less than four points in the file its reading, so we end up with a `nan` response.

sanahabhimani:
thank you both for the input and catching my half-brained errors!

mhasself (commented):
Thanks for the fixes!

socs/agents/ucsc_radiometer/agent.py
BrianJKoopman, line 73: The error looks like it's raising `requests.exceptions.JSONDecodeError`, not a `ValueError`. I think we should be catching that instead here.

BrianJKoopman (changes requested):
Based on the logs it looks like the `except` statement needs to be handling the `JSONDecodeError`, not a `ValueError`. Where was the `ValueError` coming up?

socs/agents/ucsc_radiometer/agent.py
mhasself: (I did check that JSONDecodeError is a subclass of ValueError -- so this would catch it.)

BrianJKoopman: Ah, that makes sense.

BrianJKoopman (approved):

PR #564, opened 2023-11-07
Suprsync update tcdirs

### LINKED ISSUE #562: SupRsync adding finalize files every other day
@kmharrington  noticed that finalization files were being written by by suprsync every other day, instead of having a finalization file daily. Looking closely at the code, I realized that the reason is probably that the pysmurf-monitor is not using the `srfm.add_file` function to add files to the db:
https://github.com/simonsobs/socs/blob/b9609ac8b1b37cdc5f88ac3c8efe5157e5add8e5/socs/agents/pysmurf_monitor/agent.py#L210-L211
which is an issue since the `add_file` fn is what actually adds the tcdir for the new file. This means the _only_ files that are being added via `add_file` are previous finalization files.... which I think is what's causing this leap-frogging bug.

I think the easiest way to fix this is to have the `suprsync` agent regularly check and add the timecode dirs for recent files. I am going to have the agent query the last ~weeks worth of files to add timecode dirs like once an hour or so, and see if that fixes this bug.

jlashner:
Adds routine in suprsync main function to regularly update timecode-directories for files that have not been added with the `add_file` interface.

## Description

There is currently an issue with the suprsync finalization, where timecode directories are not being added for files added by the pysmurf-monitor, since it does not use the `add_file` function in suprsync. This fixes that issue, by having the suprsync agents regularly (like every 6 hrs) add the TimecodeDir if it does not exist for any file added in the last week.

## Motivation and Context
Required to fix finalization and smurf bookbinding.

Resolves #562

## How Has This Been Tested?
Tested with smurf-sync on smurf-srv20

## Types of changes

## Checklist:

### DIFF
--- a/socs/db/suprsync.py
+++ b/socs/db/suprsync.py
@@ -394,7 +394,7 @@ class SupRsyncFilesManager:
 
         return files
 
-    def get_known_files(self, archive_name, session=None):
+    def get_known_files(self, archive_name, session=None, min_ctime=None):
         """Gets all files.  This can be used to help avoid
         double-registering files.
 
@@ -404,12 +404,18 @@ class SupRsyncFilesManager:
                 Name of archive to pull files from
             session : sqlalchemy session
                 Session to use to query files.
+            min_ctime : float, optional
+
         """
         if session is None:
             session = self.Session()
 
+        if min_ctime is None:
+            min_ctime = 0
+
         query = session.query(SupRsyncFile).filter(
             SupRsyncFile.archive_name == archive_name,
+            SupRsyncFile.timestamp > min_ctime,
         ).order_by(asc(SupRsyncFile.timestamp))
 
         return list(query.all())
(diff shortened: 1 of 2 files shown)


socs/db/suprsync.py
```diff
@@ -404,12 +404,18 @@ def get_known_files(self, archive_name, session=None):
                 Name of archive to pull files from
             session : sqlalchemy session
                 Session to use to query files.
+            min_ctime : float, optional
```
BrianJKoopman, line 407: Missing a description.

BrianJKoopman (commented):
This looks good to me. One docstring comment. Also, can you confirm this resolves #562? (And link this PR to that issue by putting "Resolves #562" under the "Motivation and Context" section.)

jlashner:
Yep, confirmed

BrianJKoopman (approved):
Thanks!

PR #569, opened 2023-11-13
add connection error exception

sanahabhimani:
## Motivation and Context
Handles the `requests.exceptions.ConnectionError` 

## How Has This Been Tested?
Hasn't. Had deployed this version in my dev environment on the actual agent this past weekend, but there wasn't a connection issue in the last couple days.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/ucsc_radiometer/agent.py
+++ b/socs/agents/ucsc_radiometer/agent.py
@@ -73,6 +73,9 @@ class UCSCRadiometerAgent:
             except ValueError:
                 pm.sleep()
                 continue
+            except requests.exceptions.ConnectionError:
+                pm.sleep()
+                continue
             data = r.json()
             last_pwv = data['pwv']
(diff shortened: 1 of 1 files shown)


socs/agents/ucsc_radiometer/agent.py
```diff
@@ -73,6 +73,9 @@ def acq(self, session, params=None):
             except ValueError:
                 pm.sleep()
                 continue
+            except requests.exceptions.ConnectionError:
```
BrianJKoopman, line 76: Great, we should also include a `self.log.warn()` level message about the closed connection so that it's clear from the logs that the radiometer web server is unavailable.

BrianJKoopman (changes requested):
Looks good, just needs some logging.

BrianJKoopman (approved):
Thanks!

Issue #575, opened 2023-11-20
Pysmurf controller check_state failing out with ChannelAccessGetFailure


jlashner:
Pysmurf controller is sometimess failng out of the `check_state` process due to ChannelAccessGetFailure, as shown in logs below:
```
2023-11-20T17:24:35+0000 Warning! Not making output directories!This will break may things!
2023-11-20T17:24:35+0000 check_state:0 Status is now "running".
2023-11-20T17:24:40+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:24:55+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:25:10+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:25:25+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:25:40+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:25:55+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:26:10+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:26:25+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:26:40+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:26:56+0000 [ 2023-11-20 17:26:56 ]  Command failed: smurf_server_s7:AMCc:SmurfProcessor:Downsampler:DownsamplerMode
2023-11-20T17:26:56+0000 [ 2023-11-20 17:26:56 ]  Retry attempt 1 of 5
2023-11-20T17:27:01+0000 [ 2023-11-20 17:27:01 ]  Retry attempt 2 of 5
2023-11-20T17:27:08+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:27:23+0000 Could not connect to epics server! Waiting and then trying again
2023-11-20T17:27:38+0000 Could not connect to epics server! Waiting and then trying again
CA.Client.Exception...............................................
    Warning: "Virtual circuit unresponsive"
    Context: "localhost:45831"
    Source File: ../tcpiiu.cpp line 919
    Current Time: Mon Nov 20 2023 17:27:53.958065110
..................................................................
2023-11-20T17:27:53+0000 check_state:0 CRASH: [Failure instance: Traceback: <class 'epics.ca.ChannelAccessGetFailure'>: Get failed; status code: 192
/usr/lib/python3.8/threading.py:932:_bootstrap_inner
[...]

PR #576, opened 2023-11-20
Adds except for caget failure

jlashner:
Adds exception for epics channel access get failure in pysmurf-controller check-state. I believe this should make the check-state function more robust by preventing the behavior seen in #575 .

## Description

Excepts caget failures in check_state

## Motivation and Context

#575 

## How Has This Been Tested?

Not tested yet

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/pysmurf-controller.rst
+++ b/docs/agents/pysmurf-controller.rst
@@ -33,6 +33,12 @@ These can be installed via pip:
     $ python -m pip install 'sodetlib @ git+https://github.com/simonsobs/sodetlib.git@master'
     $ python -m pip install 'sotodlib @ git+https://github.com/simonsobs/sotodlib.git@master'
 
+Additionally, ``socs`` should be installed with the ``pysmurf`` group:
+
+.. code-block:: bash
+
+    $ pip install -U socs[pysmurf]
+
 Configuration File Examples
 -----------------------------------
 
--- a/docs/user/installation.rst
+++ b/docs/user/installation.rst
@@ -30,6 +30,8 @@ The different groups, and the agents they provide dependencies for are:
      - Magpie Agent
    * - ``pfeiffer``
      - Pfeiffer TC 400 Agent
+   * - ``pysmurf``
+     - Pysmurf Controller Agent
    * - ``smurf_sim``
      - SMuRF File Emulator, SMuRF Stream Simulator
    * - ``synacc``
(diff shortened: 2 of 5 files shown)


BrianJKoopman (changes requested):
This looks fine to me for catching in the `check_state` process, but seeing bias steps occasionally see similar behavior (https://github.com/simonsobs/daq-discussions/discussions/59) makes me wonder if this handling is needed throughout. Maybe that goes in a separate PR, but I'd be interested to hear your thoughts.

Also, this adds `pyepics` as a direct dependency of `socs` when dealing with the pysmurf controller agent. We should add it to `setup.py`. I'm happy to do that bit if you'd like. Either way it should be done before merging.

jlashner:
Ya... I think we want other operations to fail though in the event of an epics failure, where we don't really ever want the `check_state` process to stop running, so I don't think we should change the other tasks until we understand the source of the issue.

pyepics was already a dependency through sodetlib and pysmurf, so I don't think this actually adds any new dependencies... Dependencies for this agent were all handled because it is run in its own docker that is built on the so-smurf stack, rather than through setup.py

BrianJKoopman:
> pyepics was already a dependency through sodetlib and pysmurf, so I don't think this actually adds any new dependencies... Dependencies for this agent were all handled because it is run in its own docker that is built on the so-smurf stack, rather than through setup.py

Right, but now that we're directly importing it in socs code it makes sense to track here, rather than rely on it being pulled in by another dep. I added it with the needed docs in https://github.com/simonsobs/socs/pull/576/commits/c7792e6438225c27ac9e8bd555d6d130c8d413d2.

jlashner:
If that's the case, should we also uncomment the pysmurf, sodetlib, and sotodlib reqs?

BrianJKoopman:
> Ok, I'm fine w/ merging this but am still pretty confused by it. Maybe it makes sense for tracking / docs, but
> 
> ```
> pip install -U socs[pysmurf]
> ```
> 
> will never really be sufficient for running the agent outside of the docker

True, that's why there's the dependencies section on the agent page: https://socs--576.org.readthedocs.build/en/576/agents/pysmurf-controller.html#dependencies

This is pointed to from the installation page. I agree it's not ideal. But the best solution is to get those packages to publish to PyPI.

BrianJKoopman (approved):

PR #579, opened 2023-11-28
Fix bug in speed_rate limitation in wiregrid_actuator/drivers/Actuator.py

sadachi5:
## Description

Fix bug in speed_rate limitation in wiregrid_actuator/drivers/Actuator.py for high speed mode.
The limitation was changed from <=1.0 --> <=5.0.

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/wiregrid_actuator.rst
+++ b/docs/agents/wiregrid_actuator.rst
@@ -80,9 +80,6 @@ However, the motor power is not turned ON or OFF during the both functions.
 The parameter details are here:
 - speedrate: Actuator speed rate [0.0, 5.0] (default: 1.0)
 
-.. warning::
-  DO NOT use ``speedrate > 1.0`` if ``el != 90 deg``!
-
 **Test Functions**
  - check_limitswitch(): Check ON/OFF of the limit switches
  - check_stopper(): Check ON/OFF (lock/unlock) of the stoppers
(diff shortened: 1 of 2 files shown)


docs/agents/wiregrid_actuator.rst
```diff
@@ -77,6 +77,10 @@ The main functions are ``insert()`` and ``eject()``.
 In the both of the functions, after the inserting/ejecting, the stopper locks the actuators again.
 However, the motor power is not turned ON or OFF during the both functions.
 
+The parameter details are here:
+- speedrate: Actuator speed rate [0.0, 5.0] (default: 0.2)
+  DO NOT use speedrate > 1.0 if el != 90 deg!!
+
```
BrianJKoopman, line 83: This doesn't render well, and also duplicates information in the auto-documented API section. My suggestion would be to drop it. If you'd like to keep it, I'd suggest a `warning::` block like I added below.

BrianJKoopman (changes requested):
Looks good, though I think we're duplicating efforts with the docs page update. See below comment.

docs/agents/wiregrid_actuator.rst
sadachi5: Sorry for that. I missed that point in merging the main into my branch.
I fixed it and corrected the default parameter value for the insert/eject function.

BrianJKoopman (approved):

PR #588, opened 2023-12-06
Add hwp_pcu agent for the hwp phase compensation unit operations

17-sugiyama:
The hwp_pcu agent controls the phase compensation unit (PCU) for the CHWP rotation system.
The PCU will be installed in the CHWP system of all SATs. The SAT2 has already successfully installed this unit.
I would like to add this hwp_pcu agent to the socs.

PCU increases the motor efficiency of the CHWP motor.
PCU can also stop the CHWP rotation without disabling the motor driving system.
The hardware and software of PCU are described in [this slide.](https://docs.google.com/presentation/d/1jbdmJdoj98kQmAiez3OXtEXVjHxMvFFvU_FDXhJINuw/edit?usp=sharing)

This agent is tested in the lab, and will be tested with TSAT at site.
Hopefully this agent will be included in the hwp supervisor agent.


socs/agents/hwp_pcu/agent.py
```diff
@@ -0,0 +1,245 @@
...
+
+        return True, 'Connection to PCU established'
+
+    @ocs_agent.param('command', default='off', type=str)
+    def send_command(self, session, params):
+        """send_command(command)
+
+        **Task** - Send commands to the phase compensation unit.
+        off: The compensation phase is zero.
+        on_1:The compensation phase is +120 deg.
```
ykyohei, line 80: Why don't we change from 'on_1' to '＋120 deg', 'on_2' to '-120 deg'?

socs/agents/hwp_pcu/agent.py
```diff
@@ -0,0 +1,245 @@
...
+                # status = self.PCU.get_status()
+                status = self.status
+                data['data']['status'] = status
+
+                self.agent.publish_to_feed('hwppcu', data)
+
+                session.data = {'status': status,
+                                'last_updated': time.time()}
+
+                time.sleep(1)
```
ykyohei, line 185: We don't need high sample rate for this, maybe we can increase to 3 sec ?

socs/agents/hwp_pcu/agent.py
```diff
@@ -0,0 +1,245 @@
...
+        elif command == 'hold':
+            on_channel = [0, 1, 2, 5]
+            off_channel = [6, 7]
+            for i in on_channel:
+                self.PCU.relay_on(i)
+            for i in off_channel:
+                self.PCU.relay_off(i)
+            self.status = 'hold'
+            msg = 'Phase compensation operates "hold".'
+            return True, msg
```
jlashner, line 130: All of this must be indented one layer so that it is in the lock acquisition block

jlashner (changes requested):
Thank you for this really nice code! A few comments inline about usage of locking, and a couple of questions and requests:
- Do you think you can write up a docs page that includes the information from your slides about what this does / how to use it?  You can check out existing doc pages like [this](https://github.com/simonsobs/socs/blob/main/docs/agents/lakeshore240.rst) for how these should be structured
- This agent structure, with multiple threads accessing a single hardware object with lock-outs is not really the best way to design an agent like this, [as described in this discussion](https://github.com/simonsobs/socs/discussions/593). I won't request that this agent is restructured if you don't want to since this is already very close to being mergeable, however if you are interested I am very happy to help restructure in such a way!

socs/agents/hwp_pcu/agent.py
ykyohei: I think this might be fine as it is because "on_1" does not necessarily mean +120 deg. We intend to include the hwp_pcu in the hwp_supervisor, so more accurate compensated angle can be described in hwp_supervisor.

ykyohei:
I have finished the requested modification of the code.
@17-sugiyama please check if it looks good.
I would like to set this the agent up for satp3 system as soon as possible. We confirmed that the hardware is ready to use, by using un-official python scripts.

The hardware is connected to nuc-hk1-satp3 over usb and the port is `/dev/ttyACM0`.
I would like to ask admin to fix this port to `/dev/HWPPCU` .
https://github.com/simonsobs/site-ansible/blob/44cc4107243208bd5b9cb2ec1a72b7ee4b08ec61/hosts.txt#L191C19-L191C47

jlashner (approved):
Thanks for addressing all of the comments. Looks good to me!

ykyohei (approved):

PR #592, opened 2023-12-07
Restart UPS Agent every 60 minutes

davidvng:
## Description

Break the acq process loop and exit the agent after 60 minutes. Must add `restart: unless-stopped` to docker-compose file to automatically restart the docker container everytime this occurs.

## Motivation and Context

Temporary fix for UPS agent which is the biggest contributor to high memory usage among site agents. This releases that memory by exiting the agent completely, which is what we've been doing manually. A fix to pysnmp, by switching to using asyncio, will come later.

## How Has This Been Tested?

Tested in lab

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/ups.rst
+++ b/docs/agents/ups.rst
@@ -31,10 +31,13 @@ using all of the available arguments::
        'arguments': [['--address', '10.10.10.50'],
                      ['--port', 161],
                      ['--mode', 'acq'],
-                     ['--snmp-version', 1]]},
+                     ['--snmp-version', 1],
+                     ['--restart-time', 60]]},
 
 .. note::
     The ``--address`` argument should be the address of the UPS on the network.
+    The ``--restart-time`` argument should be set to number of minutes before
+    exiting the agent. Setting to 0 (default) will not exit the agent.
 
 Docker Compose
 ``````````````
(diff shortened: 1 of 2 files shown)


docs/agents/ups.rst
```diff
@@ -53,10 +53,13 @@ example docker-compose service configuration is shown here::
       - SITE_HUB=ws://127.0.0.1:8001/ws
       - SITE_HTTP=http://127.0.0.1:8001/call
       - LOGLEVEL=info
+    restart: unless-stopped
```
BrianJKoopman, line 56: Is this needed? Host Manager should also still have the 'up' state, which should automatically restart the agent as well (and has the added perk that if the agent isn't in a Docker container still works.) Though I guess then we need to also note this only works if managed by the Host Manager.

BrianJKoopman (commented):
A few more comments.

docs/agents/ups.rst
davidvng: I think this is still needed, the default is ``restart: no`` which the Host Manager would still read in from docker-compose

socs/agents/ups/agent.py
```diff
@@ -308,7 +310,10 @@ def acq(self, session, params=None):
 
         session.set_status('running')
         self.is_streaming = True
+        timeout = time.time() + 60 * self.restart  # exit loop after self.restart minutes
         while self.is_streaming:
+            if ((timeout != 0) and (time.time() > timeout)):
```
mhasself, line 315: I think this needs to be `if (self.restart != 0) and...`

BrianJKoopman (approved):
Thanks for the updates!

PR #604, opened 2023-12-27
TimeoutLock less HWP PMX agent

ykyohei:
## Description
TimeoutLockless HWP PMX agent

## Motivation and Context
In the operation of PMX at the site, we have many timeout lock errors. To avoid this error and make PMX control more robust, we modify the PMX agent in the same way as HWP PCU agent #600
I plan to do the same modification for the HWP PID agent as well.
The way I define actions may not be good. Suggestions for better coding are welcome.

## How Has This Been Tested?
Tested in PMX in UTokyo setup

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -1,17 +1,108 @@
 import argparse
-import os
 import time
+from dataclasses import dataclass
+from queue import Queue
 
 import txaio
-from ocs import ocs_agent, site_config
-from ocs.ocs_twisted import TimeoutLock
-from twisted.internet import reactor
-
-import socs.agents.hwp_pmx.drivers.PMX_ethernet as pmx
-from socs.agents.hwp_supervisor.agent import get_op_data
+from twisted.internet import defer
 
 txaio.use_twisted()
 
+from ocs import ocs_agent, site_config
+
+import socs.agents.hwp_pmx.drivers.PMX_ethernet as pmx
+
+class Actions:
+    class BaseAction:
+        def __post_init__(self):
+            self.deferred = defer.Deferred()
+            self.log = txaio.make_logger()
+
+    def process(self, *args, **kwargs):
+            raise NotImplementedError
+
+    @dataclass
+    class SetOn(BaseAction):
+        def process(self, module, log=None):
+            if log is None:
+                log = self.log
+
+            log.info("Setting PMX on...")
+            self.dev.turn_off()
+
+    class SetOff(BaseAction):
+        def process(self, module, log=None):
+            if log is None:
+                log = self.log
+
+            log.info("Setting PMX off...")
+            self.dev.turn_off()
+
+    class UseExt(BaseAction):
+        def process(self, module, log=None):
+            if log is None:
+                log = self.log
+
+            log.info("Setting PMX Kikusui to PID control...")
+            self.dev.use_external_voltage()
+        return True
+
+    class IgnExt(BaseAction):
+        def process(self, module, log=None):
+            if log is None:
+                log = self.log
+
+            log.info("Setting PMX Kikusui to direct control...")
+            self.dev.ign_external_voltage()
+        return True
+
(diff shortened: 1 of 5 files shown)


socs/agents/hwp_pmx/agent.py
```diff
@@ -52,83 +161,62 @@ def __init__(self, agent, ip, port, f_sample=1, supervisor_id=None,
...
-                return False, 'Could not acquire lock'
-
-            if self.shutdown_mode:
-                return False, "Shutdown mode is in effect"
+        if self.shutdown_mode:
+            return False, "Shutdown mode is in effect"
 
-            self.dev.turn_on()
+        action = Actions.SetOn(**params)
+        self.action_queue.put(action)
```
jlashner, line 172: Hi Kyohei, in the lockless agents, all tasks should be structured something like this:
```python3
@defer.inlineCallbacks
  def set_on(self, session, params):
    action = Actions.SetOn(**params)
    self.action_queue.put(action)
    session.data = yield action.deferred
    return True, f"Action {action} finished successfully"
```

I'd move the shutdown-mode check to either the `_process_actions` function, or to the action's `process` function, in which case it should be passed in along with the module.

They also need to be registered with `blocking=False`, like
```python3
agent.register_task('set_on', PMX.set_on, blocking=False)
```

socs/agents/hwp_pmx/agent.py
```diff
@@ -256,83 +291,87 @@ def acq(self, session, params):
                  'last_updated': 1649085992.719602}
 
         """
+        try:
+            PMX = pmx.PMX(ip=self.ip, port=self.port)
+            self.log.info('Connected to PMX Kikusui')
+        except BrokenPipeError:
+            self.log.error('Could not establish connection to PMX Kikusui')
+            reactor.callFromThread(reactor.stop)
```
jlashner, line 299: If we can test the hardware and failure modes, I'd personally opt for something a bit more robust than this. The LS240 restructure branch has an example of how we can structure this such that on connection failures, it will continue to try making the connection instead of just killing the reactor thread: 
https://github.com/simonsobs/socs/blob/adf12ffc29681d51e502fabeb6515d7c3cf64a01/socs/agents/lakeshore240/agent.py#L203-L223

jlashner (changes requested):
Hi Kyohei, 

Looks good, but there are a few things off about the task structure here that need to be fixed. Can you fix them, and then I will help to resolve any test errors that still exist? Thanks!

jlashner:
I'm looking at the test failures here, and it seems like the PMX integration tests rely strongly on the old agent structure, and I don't personally see a clear way to modify them to work with the new structure. I think we should remove them for the time being, and maybe add in tests later that use the [HWP Emulator](https://github.com/simonsobs/socs/pull/606). @BrianJKoopman what do you think?

ykyohei:
@jlashner I made requested changes, and confirmed that these are working in UTokyo setup. Please review.

jlashner:
looking good to me! I made one small change (moved the publish / setting of session data into the try-block for the _get_data function), and fixed integration tests so that they can run for me. Hopefully tests pass, and then we can merge!

jlashner:
Tests pass!! and integration tests were very informative in catching a race condition. I think we can merge now.

jlashner (approved):

socs/agents/hwp_pmx/agent.py
```diff
@@ -256,106 +278,101 @@ def acq(self, session, params):
...
-                    if not self.lock.release_and_acquire(timeout=10):
-                        self.log.warn(f"Failed to re-acquire lock, currently held by {self.lock.job}.")
-                        continue
-
-                data = {
-                    'timestamp': current_time,
-                    'block_name': 'hwppmx',
-                    'data': {}
-                }
+        self._clear_queue()
```
mhasself, line 284: @jlashner This needs to be put in the reactor too ... I suggest blockingCallFromThread.

PR #605, opened 2023-12-30
Add Agent to support RTSP compatible IP cameras

tskisner:
This agent adds support for acquiring snapshots and short recordings from cameras that can stream the RTSP protocol.

## Description

This new agent records frames to disk (default every 10s) and passes them through a motion detection algorithm.  If there is motion, a 60 second recording is taken.  Recordings can also be triggered directly as an agent Task.  Both images and video are stored to disk in circular buffers which by default hold the most recent 2 days of snapshots and 120 minutes of triggered videos.  Use of this agent requires the opencv / cv2 and imutils packages.

Note that each agent instance handles exactly one camera.  This enables using different settings for different camera placements.

## Motivation and Context

This agent was created to support the Dahua security cameras at the site.  These are interior cameras designed to capture unauthorized access or theft.

## How Has This Been Tested?

This agent has been tested on a local system (laptop) at the site.  It has been run for several hours and successfully detected motion in one of the containers as people went in and out.  Note that I only tested this running as a local agent on the host system.  I verified that the CPU and memory use are not too dramatic.  There is a brief spike at the end of a recording when the list of frames is encoded to mp4.  Here is an example of motion detection, which draws a green box around the areas where a change was detected:

img_2023-12-30T16:54:53+00:00
img_2023-12-30T16:55:03+00:00

## Types of changes

## Checklist:


tskisner:
Just a note:  this agent does require the `opencv` and `imutils` packages.  Let me know if you want me to add another "extra" key to setup.py for this case.

socs/agents/camera_rtsp/agent.py
```diff
@@ -0,0 +1,887 @@
...
+                # Grab an image
+                _ = cap.grab()
+            success, image = cap.retrieve()
+            if not success:
+                msg = "Failed to retrieve snapshot image from stream"
+                self.log.error(msg)
+                return False, "Broken stream"
+
+            # Motion detection.  We ignore the first few snapshots and also
+            # any changes that happen while we are already recording.
```
davidvng, line 565: Is there any way to increase the recording duration if motion is continuously being detected? I figure we'd want to have a continuous video of a break-in that occurs longer than 60 seconds rather than having to open many 60 second videos.

socs/agents/camera_rtsp/agent.py
```diff
@@ -0,0 +1,887 @@
...
+        password,
+        seconds=10,
+        port=554,
+        urlpath=None,
+        jpeg_quality=20,
+        max_snapshot_files=10000,
+        record_fps=20.0,
+        record_duration=60,
+        max_record_files=100,
+        disable_motion=False,
```
davidvng, line 435: Can you make a similar argument that disables motion based on some time range during the day? I figure we wouldn't need to record motion during work hours when people on site are continuously entering/exiting containers

socs/agents/camera_rtsp/agent.py
```diff
@@ -0,0 +1,887 @@
...
+    height = first.shape[0]
+    width = first.shape[1]
+    fourcc = cv2.VideoWriter_fourcc(*"mp4v")
+    out = cv2.VideoWriter(path, fourcc, fps, (width, height))
+    for frame in data:
+        out.write(frame)
+    out.release()
+
+
+class CircularMediaBuffer:
```
davidvng, line 94: I'm not too familiar with circular media buffers, pardon my ignorance here: what happens when the agent crashes/exits and is restarted? Does this gracefully pick back up the buffer and continue handling files correctly?

socs/agents/camera_rtsp/agent.py
```diff
@@ -0,0 +1,887 @@
...
+
+    def read(self):
+        _ = self.grab()
+        return self.retrieve()
+
+    def release(self):
+        pass
+
+
+class CameraRTSPAgent:
```
davidvng, line 387: Nitpicky, but could you update this and all other instances to `RTSPCameraAgent` for consistency with the `ACTiCameraAgent`?

davidvng (changes requested):
Thanks so much for writing this agent! Here's my initial review with some requests. Mainly name changes and core functionality requests.

I'm also wondering if the classes `CircularMediaBuffer` and `MotionDetector` should become external classes since they can also be used by the `ACTiCameraAgent` and any other similar agents in the future. Though this can be a discussion with others.

BrianJKoopman:
This is great. Thanks for reviewing @davidvng, and thanks for the PR @tskisner! 

> Just a note: this agent does require the `opencv` and `imutils` packages. Let me know if you want me to add another "extra" key to setup.py for this case.

This would be good to do. If these will be used in the functions that @davidvng suggested get pulled out for use in the ACTi camera agent then I don't mind them going in the main `install_requires`, else a new `extras_require` group would be great.

Since I assume you're busy with other things now that you're back from the site @tskisner, let's go ahead and merge this as is, and @davidvng will take over making these improvements in separate PRs. That'll allow us to stand-up the agent for the cameras already deployed on site, and move forward with these suggestions. Sound good?

socs/agents/camera_rtsp/agent.py
```diff
@@ -0,0 +1,887 @@
...
+
+    pgroup.add_argument(
+        "--password",
+        type=str,
+        required=True,
+        help="Password for camera access",
+    )
+
+    pgroup.add_argument(
+        "--seconds",
```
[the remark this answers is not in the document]
tskisner: That sounds reasonable, I'll make it more descriptive.

socs/agents/camera_rtsp/agent.py
tskisner: I'll see how difficult it is to change the logic.  In the worst case, we could increase the recording time to something longer and ensure that it is only enabled at night.

socs/agents/camera_rtsp/agent.py
tskisner: That would be a nice feature that I considered for a future update, however I guess the usefulness of this agent is somewhat limited if we are continuously triggering off motion during the day.  I'll work on adding a time of day start / stop setting.

socs/agents/camera_rtsp/agent.py
tskisner: Yes, on construction it scans the existing files and picks up where it left off.  I'll move these helper classes into `socs/common/camera.py` for general use.

socs/agents/camera_rtsp/agent.py
tskisner: Sure, I was aiming for `CategorySpecificThingAgent`, but looking across the codebase now I see most agents are `SpecificThingCategoryAgent`.  I'll change it.

tskisner: This is implemented.

socs/agents/camera_rtsp/agent.py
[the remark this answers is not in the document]
tskisner: Done

socs/agents/camera_rtsp/agent.py
tskisner: This is now implemented as `--motion_start` and `--motion_stop` options.  The format is an ISO style string with UTC offset.  For example, to only perform motion detection between 7pm and 6am Chile time, one would pass:
```
--motion_start 19:00:00-04:00
--motion_stop 06:00:00-04:00
```

socs/agents/camera_rtsp/agent.py
tskisner: This is quite challenging to implement with the current logic and I am not going to do this now.  Hopefully it is sufficient to increase the recording time to a few minutes and only enable motion detection during the night.

tskisner:
This is now ready for re-review @davidvng.

socs/agents/rtsp_camera/agent.py
```diff
@@ -0,0 +1,583 @@
...
+            self.log.debug(f"{data}")
+
+            message = {
+                "block_name": "cameras",
+                "timestamp": timestamp,
+                "data": {
+                    "address": self.address,
+                    "path": path,
+                },
+            }
```
davidvng, line 292: Can you add `movement` bool to this? We may want to monitor on Grafana and trigger alarms if the cameras see movement.

socs/agents/rtsp_camera/agent.py
```diff
@@ -0,0 +1,583 @@
...
+        while self.is_streaming:
+            # Use UTC
+            timestamp = time.time()
+            data = dict()
+
+            for iframe in range(frames_per_snapshot):
+                pmgrab.sleep()
+                # Grab an image
+                _ = cap.grab()
+            success, image = cap.retrieve()
```
davidvng, line 236: Does this `grab` have to be within the loop or can it just grab at the end? From my understanding, this is grabbing 200 images for example but only retrieving the last one.

socs/common/camera.py
```diff
@@ -0,0 +1,392 @@
...
+
+    Args:
+        blur (int):  The odd number of pixels for blurring width
+        threshold (int):  The grayscale threshold (0-255) for considering
+            image changes in the blurred images.
+        dilation (int):  The dilation iterations on the thresholded image.
+        min_frac (float):  The minimum fraction of pixels that must change
+            to count as a detection.
+        max_frac (float):  The maximum fraction of pixels that can change
+            to still count as a detection (rather than a processing error).
```
davidvng, line 245: Can you add these arguments as optional kwargs to the agent? We may want to change these if the camera FOV, room size, etc cause motion detection to fail.

davidvng (commented):
@tskisner Thanks for the updates! I added some comments below. Please also add this agent to [`plugin.py`](https://github.com/tskisner/socs/blob/rtsp_camera/socs/plugin.py).

tests/agents/test_camera_rtsp.py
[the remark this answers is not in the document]
BrianJKoopman: This sounds good, but I don't think we should let it hold up this PR at this point if it's not straightforward.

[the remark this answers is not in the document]
tskisner: Agreed, it is more important in my opinion to get this merged, so that at least there is a starting point for future development / extension.  I just rebased this branch and dealt with the setup.py to pyproject.toml transition.  Another example of why we should merge this if it does no harm.

socs/agents/rtsp_camera/agent.py
tskisner: This is a consequence of the way these cameras work.  You have to continuously grab images and extract the ones you want.

BrianJKoopman (approved):
Made some small packaging/docs changes. Will squash and merge when checks complete.

PR #606, opened 2024-01-04
HWP Emulation and PID agent lockless restructure

jlashner:
This PR begins to implement an emulation framework for the HWP, to be used for testing HWP agent operation. Also restructures the HWP_PID to use the lockless format.

## Description

During the dev call, and while testing out HWP supervisor status at the site, there are a number of things that we still need to update in many of the HWP agents before we can reliably use them to observe. Due to the limited access to HWP hardware, a relatively realistic HWP emulator will be extremely helpful for developing changes, and testing them before testing on real hardware. This PR begins to implement this emulation, and I have tested it running against PID and PMX agents on my local development setup.

The data being returned from the emulator isn't perfect (especially since the PID interface requires some more reverse engineering), but it is good enough such that agent tasks are able to run successfully.

In testing, I was finding many instances in which the PID agent could not successfully run operations due to the locking structure, so I refactored that agent to have the same structure as the new timeout-lockless agents, and tested with the emulator.

This PR contains the following changes:
- Start of the HWP emulation framework
- PID lockless restructure
- some logging modifications to the DeviceEmulator class to make it more usable by general emulators like this

## How Has This Been Tested?

Tested by running with the HWP PMX and PID agents locally.

## Types of changes

## Checklist:

### DIFF
--- a/simulators/hwp/hwp_emulator.py
+++ b/simulators/hwp/hwp_emulator.py
@@ -154,12 +154,12 @@ class HWPEmulator:
         with self.state.lock:
             # self.logger.debug(cmd)
             if cmd == "*W02400000":
-                self.state.pid.direction = "forward"
-                logger.info("Setting direction: forward")
-                return "asdfl"
-            elif cmd == "*W02401388":
                 self.state.pid.direction = "reverse"
                 logger.info("Setting direction: reverse")
+                return "asdfl"
+            elif cmd == "*W02401388":
+                self.state.pid.direction = "forward"
+                logger.info("Setting direction: forward")
                 return "asdfl"
             elif cmd.startswith("*W014"):
                 setpt = hex_str_to_dec(cmd[5:], 3)
@@ -169,7 +169,13 @@ class HWPEmulator:
             elif cmd == "*X01":  # Get frequency
                 return f"X01{self.state.cur_freq:0.3f}"
             elif cmd == "*R01":  # Get Target
-                return f"R{PID._convert_to_hex(self.state.pid.freq_setpoint, 3)}"
+                return f"R01{PID._convert_to_hex(self.state.pid.freq_setpoint, 3)}"
+            elif cmd == "*R02":  # Get Direction
+                if self.state.pid.direction == "forward":
+                    return "1"
+                else:
+                    return "0"
+
             else:
                 self.logger.info("Unknown cmd: %s", cmd)
                 return "unknown"
--- a/simulators/hwp/hwp_emulator.py
+++ b/simulators/hwp/hwp_emulator.py
@@ -141,9 +141,9 @@ class HWPEmulator:
                 return f"{self.state.pmx.voltage}\n"
 
             # Error codes
-            elif cmd == ':system:error?': # Error codes
+            elif cmd == ':system:error?':  # Error codes
                 return '0,"No error"\n'
-            elif cmd == 'stat:ques?': # Status Codes
+            elif cmd == 'stat:ques?':  # Status Codes
                 return '0'
             elif cmd == 'volt:ext:sour?':
                 return f'{self.state.pmx.source}\n'
--- a/simulators/hwp/hwp_emulator.py
+++ b/simulators/hwp/hwp_emulator.py
@@ -44,7 +44,7 @@ class HWPState:
     lock = threading.Lock()
 
 
-def _create_logger(name, log_level=logging.DEBUG):
+def _create_logger(name, log_level=logging.INFO):
     logger = logging.getLogger(name)
     logger.setLevel(log_level)
     formatter = logging.Formatter("%(name)s: %(message)s")
@@ -97,12 +97,58 @@ class HWPEmulator:
         cmd = data.split(" ")[0].strip()
         self.logger.debug(cmd)
         with self.state.lock:
-            if cmd == "meas:curr?":
+
+            # Output commands
+            if cmd == 'output':
+                val = int(data.split(" ")[1].strip())
+                self.logger.info("Setting output to %d", val)
+                self.state.pmx.output = bool(val)
+            elif cmd == 'output:protection:clear':
+                self.logger.info("Commanded to clear alarms")
+            elif cmd == 'output?':
+                return str(int(self.state.pmx.output))
+
+            # Current (limit) commands
+            elif cmd == 'curr':
+                val = float(data.split(" ")[1].strip())
+                self.logger.info("Setting current to %.3f", val)
+                self.state.pmx.current = val
+            elif cmd == 'curr:prot':
+                val = float(data.split(" ")[1].strip())
+                self.logger.info("Setting current limit to %.3f", val)
+                self.state.pmx.current_limit = val
+            elif cmd == 'curr?':
                 return f"{self.state.pmx.current}\n"
+            elif cmd == 'curr:prot?':
+                return f"{self.state.pmx.current_limit}\n"
+            elif cmd == "meas:curr?":
+                return f"{self.state.pmx.current}\n"
+
+            # Voltage (limit) commands
+            elif cmd == 'volt':
+                val = float(data.split(" ")[1].strip())
+                self.logger.info("Setting current to %.3f", val)
+                self.state.pmx.voltage = val
+            elif cmd == 'volt:prot':
+                val = float(data.split(" ")[1].strip())
+                self.logger.info("Setting voltage limit to %.3f", val)
+                self.state.pmx.voltage_limit = val
+            elif cmd == 'volt:prot?':
+                return f"{self.state.pmx.voltage_limit}\n"
+            elif cmd == 'volt?':
+                return f"{self.state.pmx.voltage}\n"
             elif cmd == "meas:volt?":
                 return f"{self.state.pmx.voltage}\n"
+
+            # Error codes
+            elif cmd == ':system:error?': # Error codes
+                return '0,"No error"\n'
+            elif cmd == 'stat:ques?': # Status Codes
+                return '0'
+            elif cmd == 'volt:ext:sour?':
+                return f'{self.state.pmx.source}\n'
             else:
-                self.logger.info("Unknown cmd: %s", cmd)
+                self.logger.info("Unknown cmd: %s", data)
                 if "?" in cmd:
                     return "unknown"
 
--- a/simulators/hwp/hwp_emulator.py
+++ b/simulators/hwp/hwp_emulator.py
@@ -6,8 +6,8 @@ import threading
 import time
 from dataclasses import dataclass
 
-from socs.testing import device_emulator
 from socs.agents.hwp_pid.drivers.pid_controller import PID
+from socs.testing import device_emulator
 
 
 def hex_str_to_dec(hex_value, decimal=3):
--- a/simulators/hwp/hwp_emulator.py
+++ b/simulators/hwp/hwp_emulator.py
@@ -1,11 +1,11 @@
 """
 HWP Emulation module
 """
-from dataclasses import dataclass
 import logging
+import threading
 import time
 from copy import deepcopy
-import threading
+from dataclasses import dataclass
 
 from socs.testing import device_emulator
 
--- a/simulators/hwp/hwp_emulator.py
+++ b/simulators/hwp/hwp_emulator.py
@@ -4,20 +4,15 @@ HWP Emulation module
 import logging
 import threading
 import time
-from copy import deepcopy
 from dataclasses import dataclass
 
 from socs.testing import device_emulator
+from socs.agents.hwp_pid.drivers.pid_controller import PID
 
 
 def hex_str_to_dec(hex_value, decimal=3):
     """Converts a hex string to a decimal float"""
     return float(int(hex_value, 16)) / 10**decimal
-
-
-def float_to_hex_str(value, decimal=3):
-    """Converts a decimal float to a hex string"""
-    return "0000" + str(hex(int(value * 10**decimal)))[2:]
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


ykyohei (commented):
The lockless restructure of hwp-pid agent looks good to me.
I really want to make HWP control more robust, so I really appreciate this work.

simulators/hwp/hwp_emulator.py
```diff
@@ -0,0 +1,192 @@
+"""
+HWP Emulation module
+"""
+import logging
+import threading
+import time
+from copy import deepcopy
```
BrianJKoopman, line 7: `deepcopy` is unused.

socs/agents/hwp_pid/agent.py
```diff
@@ -1,13 +1,70 @@
 import argparse
+import queue
 import time
 
+import txaio
 from ocs import ocs_agent, site_config
 from ocs.ocs_twisted import TimeoutLock
-from twisted.internet import reactor
```
BrianJKoopman, line 6: `reactor` is now unused.

socs/agents/hwp_pid/agent.py
```diff
@@ -27,88 +84,116 @@ def __init__(self, agent, ip, port, verbosity):
...
+    def _process_actions(self, pid):
+        while not self.action_queue.empty():
+            action = self.action_queue.get()
+            try:
+                self.log.info(f"Running action {action}")
+                res = action.process(pid)
+                action.deferred.callback(res)
+            except Exception as e:
+                self.log.error(f"Error processing action: {action}")
+                action.deferred.errback(e)
```
BrianJKoopman, line 122: I'm not sure I understand this construction, is it just passing the exception somewhere? I thought `e` had to be a `Callable` to be passed to `.errback()`.

socs/agents/hwp_pid/agent.py
```diff
@@ -27,88 +84,116 @@ def __init__(self, agent, ip, port, verbosity):
...
+        session.set_status("running")
+        while session.status in ["starting", "running"]:
+            now = time.time()
+            if now - last_sample > sample_period:
+                self._get_data_and_publish(pid, session)
+                last_sample = now
 
-        self._initialized = True
+            self._process_actions(pid)
+            time.sleep(0.2)
```
BrianJKoopman, line 160: Compute from `sample_period`?

Also, better to use a pacemaker?

socs/agents/hwp_pid/agent.py
```diff
@@ -339,49 +279,43 @@ def make_parser(parser=None):
...
-    agent.register_task('declare_freq', hwppid_agent.declare_freq)
-    agent.register_task('set_pid', hwppid_agent.set_pid)
-    agent.register_task('get_freq', hwppid_agent.get_freq)
-    agent.register_task('get_target', hwppid_agent.get_target)
-    agent.register_task('get_direction', hwppid_agent.get_direction)
-    agent.register_task('set_direction', hwppid_agent.set_direction)
-    agent.register_task('set_scale', hwppid_agent.set_scale)
 
+    if args.mode is not None:
+        agent.log.warn("--mode agrument is deprecated.")
```
BrianJKoopman, line 303: Since this is true, we should remove it from the example config block in the docs.

BrianJKoopman (changes requested):
Thanks for PRing this, and sorry it took me a bit to get to. Is the intention here to eventually run integration tests for the simulator with an instance of this stood up? Do you have any examples of these tests yet?

The biggest thing to me is the logging issues pointed out in the comment on `device_emulator.py`. Other comments are relatively minor.

socs/agents/hwp_pid/agent.py
```diff
@@ -27,88 +84,116 @@ def __init__(self, agent, ip, port, verbosity):
...
-                self.log.warn(
-                    'Could not tune stop because {} is already running'.format(self.lock.job))
-                return False, 'Could not acquire lock'
-
-            self.pid.tune_stop()
-
-        return True, 'Reversing Direction'
+        action = Actions.TuneStop(**params)
+        self.action_queue.put(action)
+        session.data = yield action.deferred
```
mhasself, line 179: Reminder to keep session.data as a dict.  Maybe standardize on `session.data['output'] = ...` or something like that?

Also -- what happens here if there's an errback?  Do you need to catch that?

jlashner: If there's an errback, the `yield` will just throw the error. I find this to be the easiest way to find and debug errors in ocs operations anyway, so I don't see the need to catch and handle manually.

socs/agents/hwp_pid/agent.py
jlashner: `sample_period` is meant to be the rate at which the PID variables are queried and written to feed. This sleep time is for the faster action processing loop, which is shorter than the `samle_period`. The fact that 0.2 sec is `1/5.0` is coincidence. The handling of the sampling period is done in the block above:
```
            if now - last_sample > sample_period:
                self._get_data_and_publish(pid, session)
                last_sample = now
```

I could add the pacemaker here for the .2 sec action loop, however I'm not sure if that is necessary because the daq/data publishing is handled separately (by the code above) and so that wouldn't necessarily regulate the publish rate.

simulators/hwp/hwp_emulator.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def hex_str_to_dec(hex_value, decimal=3):
+    """Converts a hex string to a decimal float"""
+    return float(int(hex_value, 16)) / 10**decimal
+
+
+def float_to_hex_str(value, decimal=3):
+    """Converts a decimal float to a hex string"""
+    return "0000" + str(hex(int(value * 10**decimal)))[2:]
```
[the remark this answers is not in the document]
jlashner: I realized this was a (incomplete) copy of [this function from the PID driver](https://github.com/simonsobs/socs/blame/fba087ec34401fde9251e62d072d9a1f081a8579/socs/agents/hwp_pid/drivers/pid_controller.py#L63-L70), so I think this can be replaced entirely. The driver function does not use zfill, but does take just 4 hex digits, which enforces a string length of 8. I'm not sure if this is an ideal implementation, but I think its probably a good idea to just use the driver fn.

jlashner:
Hi Brian, thanks for the comments, I think I addressed or responded to all of them.

My goal with this is not _necessarily_ to use this in tests, though it may be possible. Mainly the purpose of this is to allow myself and other HWP developers to run HWP agents for the PID, PMX, supervisor, etc. on their local systems, which is very helpful for developing new features and testing behavior.

socs/testing/device_emulator.py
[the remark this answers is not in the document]
BrianJKoopman: I know you said you couldn't reproduce this locally. Just writing to confirm that the problem is resolved with that handler check.

I was a bit confused at why this was happening, since the fixture creates a new device emulator for each test, but it must be that since `.getLogger()` returns the same logger each time, each creation of an emulator results in adding a new handler to the logger. Tricky!

BrianJKoopman (commented):
The recent changes look good to me. There's just the failing tests and docs builds now.

Merging in the most recent `main` should fix the docs builds. The tests will need fixing before we can merge though.

jlashner:
Hi Brian, so I wasn't originally intending to use the emulator for tests, but when trying to fix the tests for the PID agent restructure, it ended up being much easier to use the emulator, since the PID responses are difficult to write out manually.

With the emulator, the tests work well though so I moved it to the `socs.testing` sub-package. Let me know if this is still ok with you.

tests/integration/test_hwp_pid_agent_integration.py
```diff
@@ -24,67 +25,30 @@ def test_testing(wait_for_crossbar):
...
+    assert len(state.keys()) > 0
 
 
 @pytest.mark.integtest
-def test_hwp_rotation_set_direction(wait_for_crossbar, pid_emu, run_agent, client):
-    responses = {'*W02400000': 'W02\r',
-                 '*W02401388': 'W02\r'}
-    pid_emu.define_responses(responses)
-
-    client.init_connection.wait()  # wait for connection to be made
```
BrianJKoopman, line 65: This definition of the responses that the DeviceEmulator returns directly within the tests was an intended feature. It makes clear the intended communication between the agent and the emulator right up front.

I'd prefer this over burying the actual message and response within some other structure like the HWPEmulator, strictly for clarity when reading the test.

Since the tests have historically been written after the Agents, most commonly when I developed these I'd be running the test with no response configured locally, seeing what messages are being sent, and reverse engineer what a valid expected response would be, sometimes with access to the hardware such that I can directly test the message/response pattern.

It looks like the task count decreased, with the exception of adding `get_state`, so I would have thought most of these already defined responses would have worked, especially since the driver code didn't change. Was that not the case?

socs/testing/hwp_emulator.py
```diff
@@ -0,0 +1,223 @@
...
+            self.pmx_device.create_tcp_relay(self.pmx_port)
+
+        self.update_thread.start()
+
+    def shutdown(self):
+        """Shutdown TCP Sockets and update loop"""
+        self.run_update = False
+        self.pid_device.shutdown()
+        self.pmx_device.shutdown()
+        self.update_thread.join()
```
BrianJKoopman, line 96: This seems like it might shutdown the devices before the `update_loop` thread terminates. Wouldn't you want to `.join()`, ensuring any communication executed by the update thread finishes, before `.shutdown()`?

BrianJKoopman (commented):
I spent quite a bit of time today trying to track this down and still don't really know what's going on.

I can't reproduce the failing test locally -- I do, however, produce other failed tests that seem to stem from some race condition with the task completing and when the response is checked.

An example of two back to back local runs:
```
=================================================================== short test summary info ===================================================================
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_set_direction - assert 2 == 5
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_set_pid - assert 2 == 5
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_set_scale - assert 2 == 5
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_declare_freq - assert 2 == 5
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_tune_freq - assert 6 == 5
===================================================== 5 failed, 3 passed, 1 warning in 180.37s (0:03:00) ======================================================
```

```
=================================================================== short test summary info ===================================================================
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_set_scale - assert 2 == 5
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_declare_freq - assert 2 == 5
FAILED integration/test_hwp_pid_agent_integration.py::test_hwp_rotation_tune_freq - assert 2 == 5
===================================================== 3 failed, 5 passed, 1 warning in 192.34s (0:03:12) ======================================================
```

While we're not seeing the error on the CI runner, it concerns me these crop up locally. I suspect it's something to do with the new structure. Are you able to reproduce? (As a related aside, `task.wait()` works with this new structure, yes?)

I'll note I'm not testing with the newest ocs `main` branch with the updated OpCodes. And these examples were tested on Python 3.8 (like the CI runner uses.)

You probably saw that I re-ran the tests in this PR, which failed. I isolated just the PMX and PID tests in a separate run, and that [passed no problem](https://github.com/simonsobs/socs/actions/runs/7744864504/job/21119584133).

I've only really seen the failure here when running *all* tests. I'm also seeing the occasional failure of a test in `common/test_pmx.py`.

That said, recent runs of the CI workflow on other branches (like [this merge to `main`](https://github.com/simonsobs/socs/actions/runs/7732504623) from the other day) have run fine, so I'm not convinced this is just a flakey test that's suddenly cropping up, but instead something related to changes here.

socs/testing/hwp_emulator.py
jlashner: The update thread is just for state updates, so we don't really care about consistency or final execution on shutdown.

BrianJKoopman: Ah, yup, that makes sense. Thought it might have been a possible source of some communication after shutdown, which would result in that "bad file descriptor" we've seen.

socs/testing/hwp_emulator.py
```diff
@@ -0,0 +1,223 @@
...
+    direction: str = "forward"
+    freq_setpoint: float = 0.0
+
+
+@dataclass
+class HWPState:
+    """State of the HWP Emulator"""
+    cur_freq: float = 0.0
+    pmx: PMXState = PMXState()
+    pid: PIDState = PIDState()
```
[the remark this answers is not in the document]
jlashner: That is a good catch, that must be new to python3.11 since I'm running 3.10 and don't get those errors.  I don't feel the need to upgrade my system python right this moment but will probably do so pretty soon. I'll fix those defaults though

tests/integration/test_hwp_pid_agent_integration.py
BrianJKoopman: I've gone back and forth on this in thinking again about it. The thing I like about directly defining the responses like this within the test is that it clearly defines the state of the device within the test.

That said, our tests are largely independent of the state of the device, as most tasks just store a return value (if any at all) and those return values aren't checked in the tests. But if they were being checked, depending on the state of the emulator complicates the tests. 

For example, if you checked the current frequency after changing the set point in the PID agent, the emulator spends some time getting there, rather than just directly getting the value we defined here in the test.

But, again, since our tests don't really check these values, perhaps my point is moot. Marking as resolved.

socs/agents/hwp_pid/agent.py
```diff
@@ -247,87 +287,23 @@ def set_scale(self, session, params):
...
-    def acq(self, session, params):
-        """acq()
-
-        **Process** - Start PID data acquisition.
-
-        Notes:
-            The most recent data collected is stored in the session data in the
-            structure::
+    @defer.inlineCallbacks
+    def get_state(self, session, params):
```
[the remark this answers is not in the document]
BrianJKoopman: Ah, right, that makes sense, thanks!

BrianJKoopman (approved):
Approved! Will squash and merge.

PR #612, opened 2024-01-09
Implementation of stimulator encoder agent / driver.

dixilo:
## Description

This pull request adds reader for stimulator encoder.
This agent run in Kria KR260 inside the stimulator electronics box.
This reader accesses a custom-made IP inside the PL of ZynqMP device.
The firmware code can be found in https://github.com/dixilo/kr260_stm.
## Motivation and Context

These codes will be used to read encoder of the stimulator to know exact rotation speed and timing of the signal chopping.

## How Has This Been Tested?

We tested the functionality using the stimulator in Princeton.

## Types of changes

## Checklist:


socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+            return True, 'requested to stop taking data.'
+
+        return False, 'acq is not currently running.'
+
+
+def make_parser(parser=None):
+    if parser is None:
+        parser = argparse.ArgumentParser()
+
+    _ = parser.add_argument_group('Agent Options')
```
davidvng, line 110: Add `--path-dev` and `--path-lock` as arguments here.

socs/agents/stm_encoder/drivers.py
davidvng, line 1: Change the prefix of class names from `Stm`- to `Stim`-.

socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+        with self.lock.acquire_timeout(timeout=0, job='acq') as acquired:
+            if not acquired:
+                self.log.warn(
+                    f'Could not start acq because {self.lock.job} is already running')
+                return False, 'Could not acquire lock.'
+
+            session.set_status('running')
+
+            self.take_data = True
+            session.data = {"fields": {}}
```
davidvng, line 60: Can drop `"fields"` here.

socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+
+                ts_list = []
+                en_st_list = []
+
+                while not self._dev.fifo.empty():
+                    _d = self._dev.fifo.get()
+                    ts_list.append(_d.time.utc)
+                    en_st_list.append(_d.state)
+
+                if len(ts_list) != 0:
```
davidvng, line 77: Add a log to `self.log` if this condition is not satisfied.

socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+
+                    field_dict = {'stm_enc': {'ts': ts_list[-1],
+                                              'state': en_st_list[-1]}}
+
+                    session.data['fields'].update(field_dict)
+
+                    self.agent.publish_to_feed('stm_enc', data)
+                    session.data.update({'timestamp': current_time})
+
+                time.sleep(0.01)
```
davidvng, line 89: Use the `Pacemaker`.

socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+                ts_list = []
+                en_st_list = []
+
+                while not self._dev.fifo.empty():
+                    _d = self._dev.fifo.get()
+                    ts_list.append(_d.time.utc)
+                    en_st_list.append(_d.state)
+
+                if len(ts_list) != 0:
+                    data['timestamps'] = ts_list
```
davidvng, line 78: This should go into `data['data']['timestamps']`.

socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+
+            self.take_data = True
+            session.data = {"fields": {}}
+
+            self._dev.run()
+
+            while self.take_data:
+                # Data acquisition
+                current_time = time.time()
+                data = {'timestamps': [], 'block_name': 'stm_enc', 'data': {}}
```
davidvng, line 67: Change to standard timestamp key in the data dictionary, then move the list of timestamps into data (see below).
```suggestion
                current_time = time.time()
                data = {'timestamp': current_time, 'block_name': 'stm_enc', 'data': {}}
```

davidvng (changes requested):
Thanks for the PR! Looks good overall, I added some comments for consistency similar to my previous PRs.

socs/agents/stm_encoder/agent.py
```diff
@@ -0,0 +1,137 @@
...
+
+        agg_params = {'frame_length': 60}
+        self.agent.register_feed('stm_enc',
+                                 record=True,
+                                 agg_params=agg_params,
+                                 buffer_time=1)
+
+    def acq(self, session, params):
+        '''Starts acquiring data.
+        '''
```
BrianJKoopman, line 50: Follow the process docstring formatting recommendations here:
```suggestion
        '''acq()
        
        **Process** - Start acquiring data.
        '''
```
Also add `session.data` structure to the docstring.

socs/agents/stm_encoder/drivers.py
```diff
@@ -0,0 +1,283 @@
...
+from pathlib import Path
+from queue import Queue
+from threading import Thread
+from time import sleep
+
+import numpy as np
+
+ADDR_AXI = 0x80020000
+PATH_DEV_BASE = Path(f'/sys/devices/platform/axi/{ADDR_AXI:08x}.str_rd/uio')
+PATH_LOCK = Path(__file__).parent.joinpath('.lock')
```
BrianJKoopman, line 17: This puts the lock file in where the socs module is installed, which isn't necessarily writable. This should be moved somewhere else. I would suggest `/tmp/`.

BrianJKoopman (commented):
A couple of small additional comments.

mhasself:
Great to see this an apologies if these two concerns are premature:
- What's the data rate?  If very fast, then you need separate feeds for aggregator / influx.
- I don't think it's acceptable to have the UTC - TAI offset hard-coded to 37.  Either explicitly store TAI, or have a way of getting utc-tai dynamically (including regular updates), or have an assertion, on start-up, that the current time is before Jan 1 2026.

socs/agents/stm_encoder/agent.py
dixilo: Encoder data will not appear if the stimulator chopper is not rotating.
Becasue we will rotate the chopper only during calibration, this condition is not satisfied during usual observation.
So I think adding a log for this condition will make too much information

dixilo:
Thanks for the review!

> * What's the data rate?  If very fast, then you need separate feeds for aggregator / influx.
Data rate is around 250 Hz when rotated at the maximum speed.
The encoder will send data only during calibration time.
For the current calibration plan, data rate ~30 Hz to ~250 Hz continues for about 2 minutes during time constant calibration.

Do we need separate feeds for aggregator / influx for this situation?

dixilo:
> I don't think it's acceptable to have the UTC - TAI offset hard-coded to 37. Either explicitly store TAI, or have a way of getting utc-tai dynamically (including regular updates), or have an assertion, on start-up, that the current time is before Jan 1 2026.

Thanks for the comment. I'm planning to implement the routine to get utc-tai dynamically.

dixilo:
> I don't think it's acceptable to have the UTC - TAI offset hard-coded to 37. Either explicitly store TAI, or have a way of getting utc-tai dynamically (including regular updates), or have an assertion, on start-up, that the current time is before Jan 1 2026.

I decided not to include UTC - TAI conversion and provide raw TAI values since it is difficult to implement the behavior at the leap second insertion.

socs/agents/stimulator_encoder/agent.py
```diff
@@ -0,0 +1,173 @@
...
+                    if current_time - downsample_time > 0.1:
+                        data_downsampled = {'timestamp': current_time,
+                                            'block_name': 'stim_enc_downsampled',
+                                            'data': {
+                                                'timestamps_tai': ts_list[-1],
+                                                'state': en_st_list[-1]
+                                            }}
+                        self.agent.publish_to_feed('stim_enc_downsampled',
+                                                   data_downsampled)
+                        downsample_time = current_time
```
BrianJKoopman, line 117: Sorry, I think there's some confusion about the format for publishing to data feeds and for `session.data`.

For each `publish_to_feed` call, the format for the `message` argument needs to be either:
```
message = {
    'block_name': <Key to identify group of co-sampled data>
    'timestamp': <ctime of data>
    'data': {
        'field_name_1': <datapoint1>,
        'field_name_2': <datapoint2>
    }
}
```

if your fields are singular data points, or:
```
message = {
    'block_name': <Key to identify group of co-sampled data>
    'timestamps': [ctime1, ctime2, ... ]
    'data': {
        'field_name_1': [data1_1, data1_2, ...],
        'field_name_2': [data2_1, data2_2, ...]
    }
}
```

if you're publishing a list of points for each field. So 'timestamp' becomes 'timestamps' and is a list.

For the high rate data, as written currently, it looks like the format is:
```
message = {
    'block_name': 'stim_enc',
    'timestamp': current_time
    'data': {
        'timestamps_tai': [data1_1, data1_2, ...],
        'state': [data2_1, data2_2, ...]
    }
}
```

This might cause an issue in the HK aggregator. @mhasself any thoughts on the timestamps published being in TAI here?

BrianJKoopman (changes requested):
The changes mostly look good to me, except for the format for the message published to the high rate feed. See the comment below.

socs/agents/stm_encoder/agent.py
BrianJKoopman: This makes sense for the `session.data` object, but since `data` is being used in the `self.agent.publish_to_feed()` call 'timestamps' needs to be at this higher level. See https://ocs.readthedocs.io/en/main/developer/feeds.html#recorded-feed-message-format.

socs/agents/stm_encoder/agent.py
BrianJKoopman: Related to above, plural 'timestamps' is important for the publish to feed call, and when publishing a list of points, needs to also contain a list for the associated value.

socs/agents/stimulator_encoder/agent.py
mhasself: Putting the timestamps into a data field called `timestamps_tai` is a good solution.

As @BrianJKoopman says, you need to also record a full vector of timestamps, in message['timestamps'].  The most appropriate thing would be to grab the system time using `time.time()` in the `while not self._dev.fifo.empty()` loop.  (Those timestamps are not high precision but they're still useful and relevant... e.g. you can use them to make a very good guess of the [UTC - TAI] offset in processing, later.)

mhasself: I have two more general comments on the data feeds here.

In the "slow" feed, for grafana, it will not be that useful to have a heavily downsampled version of the raw data.  I would be *much more interested* for a regular published value of the spin frequency, computed from the last ~2 seconds of data.  Instead of 10 Hz, just do 0.5 Hz and do a pulse count / time delta.

Furthermore -- for that slow feed, it's useful to publish the frequency *even when the thing isn't spinning*.  That is helpful to distinguish between "there were no encoder pulses" and "the agent/process wasn't running".

If the slow feed has spin frequency, and is always running ... then I want it in the Aggregator / permanent HK data too!

So ya -- I think you should just record the fast data to aggregator, and then record a slower feed with spin frequency, including when freq=0, for both influx + aggregator.

(In the slow feed, instead of timestamps_tai you could record the most recent value of (timestamp_tai - timestamp) -- that will hover around the current UTC offset (37 or something).)

socs/agents/stimulator_encoder/agent.py
```diff
@@ -0,0 +1,173 @@
...
+                current_time = time.time()
+                data = {'timestamp': current_time, 'block_name': 'stim_enc', 'data': {}}
+
+                ts_list = []
+                en_st_list = []
+
+                while not self._dev.fifo.empty():
+                    _d = self._dev.fifo.get()
+                    ts_list.append(_d.time.tai)
+                    en_st_list.append(_d.state)
```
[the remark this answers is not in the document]
dixilo: Thank you for pointing out. We don't need 100 Hz here at all.

FIFO depth inside the FPGA firmware is 512 and the loop [here](https://github.com/simonsobs/socs/blob/a9c5c319c51696cd59e0f4d98ab2e4c11bc883b3/socs/agents/stimulator_encoder/drivers.py#L212) is taking it into account.
Since the fifo accessed from the agent is a Queue object on Python and the device has 4GB DDR4 memory, 1 Hz reading perfectly works.

I will simply reduce the loop time to 0.5 Hz so that we don't need to care about the different loop time between fast and slow feeds.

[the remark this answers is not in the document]
mhasself: Those changes look great to me!

BrianJKoopman (approved):
Thanks for all the updates on this one! Looks good to me, going to go ahead and merge.

PR #613, opened 2024-01-10
Remove incompatible network_mode from example config

BrianJKoopman:
## Description

"host" network mode is incompatible with port binding. The port binding example also shows a restriction to just accepting connections from localhost, which I don't think we want if the encoder is across the network and not running on the same machine that the wg-encoder agent is running.

## Motivation and Context

The daq-satp3 deployment configs used this combination "host" mode and port binding, which was preventing the wg-encoder agent container from coming online.

## How Has This Been Tested?

The satp1 and satp3 configs have been updated to use just the non-`localhost` port binding, with no 'host' network mode.

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/wiregrid_encoder.rst
+++ b/docs/agents/wiregrid_encoder.rst
@@ -54,17 +54,14 @@ An example docker-compose configuration::
       image: simonsobs/socs:latest
       restart: always
       hostname: ocs-docker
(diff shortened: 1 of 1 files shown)


sadachi5 (approved):
As for network mode, I understood your explanation. Please change as you have changed.

PR #614, opened 2024-01-11
Changed acquisition function to release the TimeoutLock when not actively querying data

bbixler500:
Changed acquisition function to release the TimeoutLock when not actively querying data

## Description

Currently the acq process always holds the TimeoutLock and periodically releases and reacquires the lock to allow other commands to process. These changes make it so that the acq process only holds the TimeoutLock when it is actively querying data and otherwise leaves the lock released.

## Motivation and Context

The acq process only releases the lock every 5 seconds, while most other commands only have a 3 second timeout for acquiring the TimeoutLock. Thus around 40% of all issued commands return false with a timeout error. 

## How Has This Been Tested?

This has not been tested yet. I mostly wanted to start the pull request so others could see the proposed changes and make comments/changes

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pid/agent.py
+++ b/socs/agents/hwp_pid/agent.py
@@ -273,24 +273,15 @@ class HWPPIDAgent:
                  'last_updated': 1649085992.719602}
 
         """
-        with self.lock.acquire_timeout(timeout=10, job='acq') as acquired:
-            if not acquired:
-                self.log.warn('Could not start iv acq because {} is already running'
-                              .format(self.lock.job))
-                return False, 'Could not acquire lock'
+        session.set_status('running')
+        self.take_data = True
 
-            session.set_status('running')
-            last_release = time.time()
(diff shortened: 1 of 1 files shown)


ykyohei:
Just for your information, this PR https://github.com/simonsobs/socs/pull/606 is also addresses the similar timeout lock issue more fundamentally.

jlashner:
Hi Bryce, I think this change looks good, so if you need this ASAP and are able to test, I'm happy to approve.

As Kyohei mentinoed though, this PR  https://github.com/simonsobs/socs/pull/606 restructures the PID agent to avoid Timeout lock entirely to fix this issue, so this change will be overwritten when that gets merged in.

bbixler500:
Currently I have a work around where I just disable the acq process whenever I want to issue commands to the agent so I don't need these changes merged immediately. I'm fine to just wait until the larger changes have been implemented.

PR #627, opened 2024-01-30
SynAccess agent add outlet names

### LINKED ISSUE #625: SynAccess agent should provide a way to assign meaningful labels to outlets
In the ibootbars, the devices themselves can store an outlet name (e.g. LS330).  But synaccess does not seem to have that.  So in ocs-web, or a client, all you know about the outlet that you're about to toggle is its index... 1,2,3... and you need to look up the outlet significance elsewhere.

It would provide a more informative interface to the hardware if the agent args could be used to configure labels for each outlet; e.g.
```
--label outlet2 LS330 --label outlet3 LS219
```
Or something more ocs-ey.

davidvng:
## Description

Adds outlet names as a parameter to SynAccess agent.

## Motivation and Context

Resolves #625 

## How Has This Been Tested?

To be tested with ocs-web on site.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/synacc/agent.py
+++ b/socs/agents/synacc/agent.py
@@ -33,7 +33,7 @@ class SynaccessAgent:
                 outlet_names.append('outlet%i' % (i + 1))
             self.outlet_names = outlet_names
         else:
-            for i in range(5):
+            for i in range(num_outlets):
                 if len(outlet_names) <= i:
                     outlet_names.append('outlet%i' % (i + 1))
             self.outlet_names = outlet_names
--- a/socs/agents/synacc/agent.py
+++ b/socs/agents/synacc/agent.py
@@ -17,13 +17,18 @@ class SynaccessAgent:
         password(str): Password credential to login to device.
     """
 
-    def __init__(self, agent, ip_address, username, password):
+    def __init__(self, agent, ip_address, username, password, outlet_names=None):
         self.agent = agent
         self.log = agent.log
         self.lock = TimeoutLock()
         self.ip_address = ip_address
         self.user = username
         self.passw = password
+
+        if outlet_names is None:
+            self.outlet_names = ['outlet1', 'outlet2', 'outlet3', 'outlet4', 'outlet5']
+        else:
+            self.outlet_names = outlet_names
 
         agg_params = {'frame_length': 60}
         self.agent.register_feed(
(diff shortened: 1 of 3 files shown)


socs/agents/synacc/agent.py
```diff
@@ -17,14 +17,19 @@ class SynaccessAgent:
...
         self.log = agent.log
         self.lock = TimeoutLock()
         self.ip_address = ip_address
         self.user = username
         self.passw = password
 
+        if outlet_names is None:
+            self.outlet_names = ['outlet1', 'outlet2', 'outlet3', 'outlet4', 'outlet5']
+        else:
+            self.outlet_names = outlet_names
```
mhasself, line 31: Should we check the length of `outlet_names` here?  I guess it's pretty easy to debug an indexing error later.

But I think it might be nice to pad the list, so a user only has to declare the You could even make it "pad" the list with any missing entries;
```
for i in range(MAX_POSSIBLE_OUTLETS):
  if len(outlet_names) <= i:
    outlet_names.append('outlet%i' % (i+1))
```

mhasself (commented):
Thanks for the quick action!  I'm ok with this (except see comment), but will leave approval to socs bosses.

I note this breaks the session.data interface.  That doesn't bother me (though it will require an `ocs-web` update).  Generally it would be good to take this opportunity to make this agent more like the ibootbar agent -- e.g. consider:
- change process name "status_acq" to "acq"?
- remove "synaccess: {}" level of the session.data -- don't need that nesting really

socs/agents/synacc/agent.py
```diff
@@ -17,14 +17,22 @@ class SynaccessAgent:
...
         self.log = agent.log
         self.lock = TimeoutLock()
         self.ip_address = ip_address
         self.user = username
         self.passw = password
 
+        if outlet_names is None:
+            self.outlet_names = ['outlet1', 'outlet2', 'outlet3', 'outlet4', 'outlet5']
+        else:
+            for i in range(5):
```
BrianJKoopman, line 31: Are all synaccess PDUs 5 outlets? (I suspect @mhasself was suggesting a constant be used that could be defined elsewhere, in the event that the outlet count differs.)

I think they do make different size units, but maybe all the ones we use are 5 outlets?

BrianJKoopman (commented):
This looks good, just one comment related to the padding that was recently added.

BrianJKoopman (approved):

PR #628, opened 2024-02-02
Supervisor actions

jlashner:
Provides various updates to the HWP Supervisor required for site use.

## Description

This PR adds some changes that are necessary for the Supervisor to be run through sorunlib and the scheduler. Mainly, we needed a better way to track if a user-commanded operation has finished, and to determine if it was successful.

Currently, the supervisor is done by a fairly simple state machine, where agent tasks would put the machine in a particular state, and states could put the machine into a new state depending on the update function.

This PR modifies this a bit, by grouping states together into "Actions", based around user requests. Now, when a user runs an operation like `pid_to_freq`, instead of just setting the PidToFreq state, this will request a new Action that begins in the PidToFreq state. This action can be monitored until it enters a "completed" state (i.e. `Done`, `Error`, or `Abort`) at which point the action will no longer be modified. Actions provide a bit more extra book-keeping, including a unique id, the current state, the state-history of the action, and whether the action is completed and successful.

Tasks now will create an action, and monitor it until the action is completed, at which point the task will finish.

An action can be aborted by running the `abort_action` task, which will put the current action into the "Abort" state at the next opportunity, and will put the state machine into an idle state. 

Note: This requires changes from the hwp-emulation branch, so I set that as the base. I can change that after hwp-emulator is merged in.

## Motivation and Context

Necessary behavior changes in order to add to next-line

## How Has This Been Tested?

Certain operations such as `pid_to_freq`, and `pmx_off` have been tested with the HWP emulator. We hope to be able to be able to test more on a real platform.

## Types of changes

## Checklist:

### DIFF
--- a/docs/agents/hwp_supervisor_agent.rst
+++ b/docs/agents/hwp_supervisor_agent.rst
@@ -109,7 +109,7 @@ or not. For example, to spin up to a particular frequency, you can run:
                       {'class': 'Done', 'msg': None, 'success': True}],
           'success': True}
 
-To stop an action while its running, you can use the ``abort_action`` task, 
+To stop an action while its running, you can use the ``abort_action`` task,
 which will set the state of the current action to ``Abort``, and put the
 supervisor into the Idle state.
 
@@ -127,7 +127,7 @@ supervisor into the Idle state.
     print("Result 2: ")
     print(res2.session['data']['action'])
 
-    >> 
+    >>
     Result 1:
     {'action_id': 1,
     'completed': True,
@@ -139,7 +139,7 @@ supervisor into the Idle state.
                         'target_freq': 2.0},
                       {'class': 'Abort'}],
     'success': False}
-    Result 2: 
+    Result 2:
     {'action_id': 3,
     'completed': True,
     'cur_state': {'class': 'Done', 'msg': None, 'success': True},
--- a/docs/agents/hwp_supervisor_agent.rst
+++ b/docs/agents/hwp_supervisor_agent.rst
@@ -67,12 +67,88 @@ This agent has two main purposes:
 - Serve as a host for high-level HWP operations that require coordinated control
   of various HWP subsystems, such as "Begin rotating at 2 Hz"
 
-Right now only the first point is implemented, but operations can be added here
-as we need them.
+For the first point, the supervisor agent implements the ``monitor`` process,
+which monitors HWP-related to compile full state info for the HWP, and uses
+that to make a determination if the HWP should be shutdown.
 
-HWP subsystems should implement a ``monitor_shutdown`` process that uses the
-``get_op_data`` function to get the hwp-supervisor's session data, and check
-the ``action`` field to determine if shutdown should be initiated.
+For high-level control, the HWP supervisor implements a state machine that
+is used to perform complex actions with HWP agents that depend on the global
+
+Running actions
+`````````````````
+HWP supervisor actions such as ``pid_to_freq``, ``set_const_voltage``,
+``brake``, and ``pmx_off`` request that the corresponding is run by the main
+``spin_control`` process.  Control tasks like these will sleep until the
+corresponding action enters an end- state such as Done, Error, or Abort.
+
+You can run an action and make sure it is complete by running the task and
+waiting for it to complete. The session-data that is returned will contain the
+encoded action, including the state chain, and whether the action was successful
+or not. For example, to spin up to a particular frequency, you can run:
+
+.. code-block:: python
+
+    supervisor = OCSClient("hwp-supervisor")
+
+    result = supervisor.pid_to_freq(target_freq=2.0)
+    print(result.session['data']['action'])
+
+    >> {'action_id': 6,
+        'completed': True,
+        'cur_state': {'class': 'Done', 'msg': None, 'success': True},
+        'state_history': [{'class': 'PIDToFreq',
+                            'direction': '0',
+                            'freq_tol': 0.05,
+                            'freq_tol_duration': 10.0,
+                            'target_freq': 2.0},
+                          {'class': 'WaitForTargetFreq',
+                            'freq_tol': 0.05,
+                            'freq_tol_duration': 10.0,
+                            'freq_within_tol_start': 1706829677.7613404,
+                            'target_freq': 2.0},
+                      {'class': 'Done', 'msg': None, 'success': True}],
+          'success': True}
+
+To stop an action while its running, you can use the ``abort_action`` task, 
+which will set the state of the current action to ``Abort``, and put the
+supervisor into the Idle state.
+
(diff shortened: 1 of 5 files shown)


BrianJKoopman:
Now that the tests in #606 are working can you rebase this?

socs/agents/hwp_supervisor/agent.py
```diff
@@ -883,16 +969,13 @@ def spin_control(self, session, params):
...
-            state_dict = asdict(s)
-            state_dict['state_name'] = s.__class__.__name__
-
             session.data = {
-                'state': state_dict,
-                'timestamp': time.time(),
+                'current_action': self.control_state_machine.action.encode(),
+                'action_history': [a.encode() for a in self.control_state_machine.action_history],
+                'timestamp': time.time()
             }
```
BrianJKoopman, line 976: Add `session.data` example to docstring.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -929,11 +1012,9 @@ def pid_to_freq(self, session, params):
...
             direction=d
         )
-        success = self.control_state_machine.request_state(state)
-        if success:
-            return True, f"Set state to {state}"
-        else:
-            return False, "Failed to update state"
+        action = self.control_state_machine.request_new_action(state)
+        action.sleep_until_complete(session=session)
+        return True, f"Completed with state: {action.cur_state}"
```
BrianJKoopman, line 1017: What happens when the requested action state fails? Does this return something like `True, f"Complete with state:" {'class': 'Error', 'msg': None, 'success': False}`?

socs/agents/hwp_supervisor/agent.py
```diff
@@ -929,11 +1012,9 @@ def pid_to_freq(self, session, params):
             freq_tol_duration=params['freq_tol_duration'],
             direction=d
         )
-        success = self.control_state_machine.request_state(state)
-        if success:
-            return True, f"Set state to {state}"
-        else:
-            return False, "Failed to update state"
+        action = self.control_state_machine.request_new_action(state)
+        action.sleep_until_complete(session=session)
```
BrianJKoopman, line 1016: Add an example of the `session.data` to the docstring.

(Do this everywhere `action.sleep_until_complete(session=session)` occurs.)

BrianJKoopman (commented):
This is look good. My comments/questions are mostly:
* Documentation change requests/comments
* Asking for clarification on how abort works
* Checking in on task success verification and wondering what happens when Supervisor agent tasks fail

socs/agents/hwp_supervisor/agent.py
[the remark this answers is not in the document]
jlashner: Is it ok if we write this as an issue and address it in a separate PR?

docs/agents/hwp_supervisor_agent.rst
```diff
@@ -67,12 +67,88 @@ This agent has two main purposes:
...
+                            'freq_tol': 0.05,
+                            'freq_tol_duration': 10.0,
+                            'freq_within_tol_start': 1706829677.7613404,
+                            'target_freq': 2.0},
+                      {'class': 'Done', 'msg': None, 'success': True}],
+          'success': True}
+
+To stop an action while its running, you can use the ``abort_action`` task,
+which will set the state of the current action to ``Abort``, and put the
+supervisor into the Idle state.
```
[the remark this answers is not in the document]
jlashner: Nope

socs/agents/hwp_supervisor/agent.py
```diff
@@ -480,14 +500,82 @@ class PmxOff:
...
-        freq_tol_duration: float = 20
-        start_time: float = field(default_factory=time.time)
+        success: bool = True
+
+    @dataclass
+    class Abort(Base):
+        """Abort current action"""
+        pass
+
+    completed_states = (Done, Error, Abort, Idle)
```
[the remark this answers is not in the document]
jlashner: This is at the bottom since the classes need to be defined in the namespace before the `completed_states` var is set.

socs/agents/hwp_supervisor/agent.py
jlashner: Yes... I can return `action.success` as the first param.

socs/agents/hwp_supervisor/agent.py
jlashner: Is it ok if I do something simple like:
```
>> session['data'] 
{'action': 
  {'action_id': 3,
  'completed': True,
  'cur_state': {'class': 'Done', 'msg': None, 'success': True},
  'state_history': List[ConrolState],
  'success': True}
}
```

Or do I need to go through and run each of the tasks and copy a realistic action for each task? It is a bit difficult to maintain correct session.data in per-operation docstrings, and there are already examples of what the encoded action looks like in the rst.

jlashner:
Hi Brian, updated docs and tried to address all of your comments, so if you could re-review that would be great.

In the RST I added a couple of paragraphs that try to explain what a Control Action is within the agent. They are pretty much one-to-one with OCS operations, so its _almost_ interchangeable, but not exactly the same thing.

socs/agents/hwp_supervisor/agent.py
BrianJKoopman: Thanks for adding. I think simple examples are fine, as long as they contain enough information that someone developing a control program can know what to expect from the object.

Since the `session.data` object is sort of our API for passing data to clients for programmatic checks in control programs, these interfaces are important. It's difficult when writing a control program to use the `session.data` if it's not well documented, especially so if you don't have a running instance of the Agent to develop against.

On that note, changing what's kept in `session.data` should be carefully considered and then called out in future PRs, since those changes potentially break control programs that depend on them.

socs/agents/hwp_supervisor/agent.py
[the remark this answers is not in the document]
BrianJKoopman: Yeah, I think that's fine. I'll open an issue and mark this comment as resolved.

socs/agents/hwp_supervisor/agent.py
BrianJKoopman: Yeah, I think we should do that. Otherwise you need to inspect the `ControlState` that's returned to know whether the task completed successfully, rather than just `response.session['success']`.

BrianJKoopman (changes requested):
(Sorry, botched the review -- GitHub's comment response options make that easy to do, more comments.)

Thanks for the docs updates, the new section on Actions cleared things up for me. I do think we need to return `action.success`, instead of just `True` all the time. Other than that, looks good!

jlashner:
Hey brian, just submitted a commit that addresses your last piece of feedback

BrianJKoopman (approved):

Issue #607, opened 2024-01-05
Race Conditions when using OpCodes and session.status


jlashner:
I'm encountering the following issue when trying to fix tests for the HWP PMX agent:

The following test is failing the assertion that `assert resp.session['op_code'] == OpCode.SUCCEEDED.value`
```python3
@pytest.mark.integtest
def test_hwp_rotation_set_on(wait_for_crossbar, kikusui_emu, run_agent, client):
    responses = {'output 1': '',
                 'output?': '1'}

    kikusui_emu.define_responses(responses)
    resp = client.set_on()
    print(resp)
    print(resp.session)
    assert resp.status == ocs.OK
    assert resp.session['op_code'] == OpCode.SUCCEEDED.value
``` 
with the following error message:
```
default response set to 'None'
msg='output 1'
response=''
msg='output?'
response='1'
OCSReply: OK : Operation "set_on" is currently not running (SUCCEEDED).
  set_on[session=2]; status=starting for 1.0 s
  messages (1 of 1):
    1704472073.685 Status is now "starting".
  other keys in .session: op_code, data
{'session_id': 2, 'op_name': 'set_on', 'op_code': 2, 'status': 'starting', 'success': True, 'start_time': 1704472073.6851099, 'end_time': None, 'data': None, 'messages': [[1704472073.6851099, 'Status is now "starting".']]}
F

================================================== FAILURES ==================================================
__________________________________________ test_hwp_rotation_set_on __________________________________________

wait_for_crossbar = None, kikusui_emu = <socs.testing.device_emulator.DeviceEmulator object at 0x7ff6fdd3bd60>
run_agent = None, client = OCSClient('hwp-pmx')

    @pytest.mark.integtest
    def test_hwp_rotation_set_on(wait_for_crossbar, kikusui_emu, run_agent, client):
        responses = {'output 1': '',
                     'output?': '1'}
    
        kikusui_emu.define_responses(responses)
        resp = client.set_on()
        print(resp)
        print(resp.session)
    
        # import time
        # time.sleep(1)
        # resp = client.set_on.status()
    
        # print(resp)
        # print(resp.session)
    
        assert resp.status == ocs.OK
>       assert resp.session['op_code'] == OpCode.SUCCEEDED.value
E       assert 2 == 5
E        +  where 5 = <OpCode.SUCCEEDED: 5>.value
E        +    where <OpCode.SUCCEEDED: 5> = OpCode.SUCCEEDED

[...]

PR #632, opened 2024-02-22
moves callbacks to reactor in PCU agent

jlashner:
Fixes race condition from callbacks in PCU agent

## Description

During HWP testing today Kyohei and I saw some strange behavior in when calling PCU agent operations through a client, which I am pretty sure are due to how [this race condition](https://github.com/simonsobs/socs/issues/607) interacts  [with the OCS timeout logic](https://github.com/simonsobs/ocs/blob/7badb58c6094fba6f1607284df3437ce20c06e91/ocs/ocs_agent.py#L818-L837). The behavior we were seeing was that anytime we called an operation through an OCS client with a timeout that wasn't None, it would return a TIMEOUT error even when successful. 

For testing, we changed the supervisor agent to not set any timeouts when interacting with the PCU agent, but I'm pretty sure this can be fixed by fixing that race condition in the agent re-structure.

## Motivation and Context
Important for HWP supervisor

## How Has This Been Tested?

This has not been tested.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pcu/agent.py
+++ b/socs/agents/hwp_pcu/agent.py
@@ -75,10 +75,10 @@ class HWPPCUAgent:
             try:
                 self.log.info(f"Running action {action}")
                 res = process_action(action, PCU)
-                action.deferred.callback(res)
+                reactor.callFromThread(action.deferred.callback, res)
             except Exception as e:
                 self.log.error(f"Error processing action: {action}")
-                action.deferred.errback(e)
+                reactor.callFromThread(action.deferred.errback, e)
 
     def _get_and_publish_data(self, PCU: pcu.PCU, session):
         now = time.time()


BrianJKoopman (approved):
This seems fine to me. Would having https://github.com/simonsobs/ocs/pull/371 help this too? That's merged upstream, but isn't released yet.

Do you want to test before merging?

jlashner:
Testing before merging might be difficult with how things are setup on satp3 ([see this issue](https://github.com/simonsobs/ocs-deployment-configs/issues/158)), and we'd need to find system time.... so if we don't need to I'd rather not, this seems simple enough.

Ya I think that OCS PR would also help. I'm not sure how urgent this is because we were able to work around it in recent supervisor changes and HWP spin-up, spin-down and brake ops seem to be working consistently. But I think when that new OCS version is ready to release we should get agents running on it.

BrianJKoopman:
> Testing before merging might be difficult with how things are setup on satp3 ([see this issue](https://github.com/simonsobs/ocs-deployment-configs/issues/158)), and we'd need to find system time.... so if we don't need to I'd rather not, this seems simple enough.

I'm fine with going ahead on this, since [it seemed to fix the problem](https://github.com/simonsobs/socs/issues/607#issuecomment-1881451004) when you tested initially fixing that race condition.

I saw that infrastructure issue go by, but haven't had time to look at the system really. Limited system time makes debugging that difficult too.
 
> Ya I think that OCS PR would also help. I'm not sure how urgent this is because we were able to work around it in recent supervisor changes and HWP spin-up, spin-down and brake ops seem to be working consistently. But I think when that new OCS version is ready to release we should get agents running on it.

Cool, we could release whenever. I'll move it up my list, but not to the top if this isn't urgent.

PR #635, opened 2024-02-28
Change check range of gripper distance

ykyohei:
Change check range of gripper distance from 10 mm to 20 mm.

## Description
Since the actuator distance of the SATP3 is greater than 10 mm, it is helpful to allow a distance of up to 20 mm for the operation of the SATP3 gripper.

## How Has This Been Tested?
No testing is required because this is a very minor change.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_gripper/agent.py
+++ b/socs/agents/hwp_gripper/agent.py
@@ -170,7 +170,7 @@ class HWPGripperAgent:
 
     @ocs_agent.param('mode', default='push', type=str, choices=['push', 'pos'])
     @ocs_agent.param('actuator', default=1, type=int, check=lambda x: 1 <= x <= 3)
-    @ocs_agent.param('distance', default=0, type=float, check=lambda x: -10. <= x <= 10.)
+    @ocs_agent.param('distance', default=0, type=float, check=lambda x: -20. <= x <= 20.)
     def move(self, session, params=None):
         """move(mode='push', actuator=1, distance=0)
 
--- a/socs/agents/hwp_gripper/agent.py
+++ b/socs/agents/hwp_gripper/agent.py
@@ -170,7 +170,6 @@ class HWPGripperAgent:
 
     @ocs_agent.param('mode', default='push', type=str, choices=['push', 'pos'])
     @ocs_agent.param('actuator', default=1, type=int, check=lambda x: 1 <= x <= 3)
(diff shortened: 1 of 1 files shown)


bbixler500 (approved):

BrianJKoopman (commented):
What happens if satp1 tries to move beyond the previous limit of 10 mm?

jlashner (approved):
I'm fine with this, I kind of dislike these hardcoded limits in general and think physical limits we want to enforce should probably come from the site cfg rather than the agent itself

ykyohei:
> What happens if satp1 tries to move beyond the previous limit of 10 mm?

If the actuator tries to move further than the grip distance (physical limit), the gipper motor reaches the current limit and raise an alarm and stop. But this is how we firmly grip hwp.
The grip distance is different for different actuators and different SATs.
https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/185434251/CHWP+main+page#Actuator-distance

I think it's also fine to just delete this limit.

bbixler500:
I agree that it is safe to remove the limit as well. Notably this was only a limit for a single move command (you could call two 10mm move commands in succession without issue to end up at 20mm). The limit was originally in place because I did not imagine there was ever a need to move more than 10mm at once. This is obviously not true anymore.

BrianJKoopman:
Great, let's just drop the `check` argument in that decorator then.

socs/agents/hwp_gripper/agent.py
```diff
@@ -170,7 +170,6 @@ def brake(self, session, params=None):
 
     @ocs_agent.param('mode', default='push', type=str, choices=['push', 'pos'])
     @ocs_agent.param('actuator', default=1, type=int, check=lambda x: 1 <= x <= 3)
-    @ocs_agent.param('distance', default=0, type=float, check=lambda x: -10. <= x <= 10.)
```
BrianJKoopman, line 173: We still need most of this, it should just be:
```
@ocs_agent.param('distance', default=0, type=float)
```

BrianJKoopman (changes requested):

socs/agents/hwp_gripper/agent.py
ykyohei: Oh, sorry I will fix it.

BrianJKoopman (approved):
Looks good, thanks for the update!

PR #641, opened 2024-03-12
hwp-pmx: reset protection code in clear_alarm

ykyohei:
## Description
Minor debug of hwp-pmx.
We need to reset the protection code (self.prot) in clear_alarm task.
This was implemented in the previous version of agent (following link), but I forgot to include this self.prot reset when we refactored the hwp-pmx agent to be TimoutLock less.

https://github.com/simonsobs/socs/blob/288fb5c2cac25d554bcfa161e6fb455b7eee1fd5/socs/agents/hwp_pmx/agent.py#L129

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -158,6 +158,7 @@ class HWPPMXAgent:
         action = Actions.ClearAlarm(**params)
         self.action_queue.put(action)
         session.data = yield action.deferred
+        self.prot = 0
         return True, 'Clear alarm'
 
     @defer.inlineCallbacks


d1ssk (approved):
Looks good.

jlashner:
Seems like this could also be done by allowing prot to be set back to 0 in the acq function:
https://github.com/simonsobs/socs/blob/288fb5c2cac25d554bcfa161e6fb455b7eee1fd5/socs/agents/hwp_pmx/agent.py#L297-L298

Is there a reason we don't do that? It's generally preferable to read from hardware when possible instead of setting value directly in the agent.

ykyohei:
Agree that it's generally preferable to read from hardware. But this is intentional.

When the pmx protection is triggered, pmx outputs protection code only once, you will get 0 from the second queries. However, pmx remains in protected state. If you run clear_alarm, pmx will go back to the normal state. If there is still problem, pmx will output protection code again, but only once.

For this reason, we made the agent to have `self.prot` for monitoring purposes.

PR #669, opened 2024-05-01
Moves pysmurf-controller functions that load lots of data to subprocess

jlashner:
This PR changes pysmurf-controller operations that load TODs such that instead of running in the main process, they call out to a subprocess.

## Description

This introduces the `smurf_subprocess_util` module, and a protocol for running functions inside a twisted subprocess in such a way that a general RunCfg can be passed to it, and a general RunResult can be returned. Any operation that involves TOD data loading is modified so that the loading and analysis occurs in a subprocess rather than in the main one.

## Motivation and Context

This should hopefully solve the memory leak, which I believe is coming from instances where TOD data is not being released to the system after it leaves scope. Moving TOD loading and analysis into subprocesses instead of the main one should enforce that this memory is returned to the OS when the subprocess exits. [This discussion](https://github.com/simonsobs/daq-discussions/discussions/81#discussioncomment-9286023) has some links that motivate this solution.

## How Has This Been Tested?

I have tested the subprocess protocol locally through the use of the `test` process (and the new `run_test_func` operation).
I will need to test all other operations with hardware on SATp1 or SATp3.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/pysmurf_controller/agent.py
+++ b/socs/agents/pysmurf_controller/agent.py
@@ -21,8 +21,9 @@ from ocs.ocs_twisted import TimeoutLock
 from sodetlib.det_config import DetConfig
 from sodetlib.operations import (bias_dets, bias_steps, bias_wave, iv,
                                  uxm_relock, uxm_setup)
-
-NBIASLINES = 12
+from socs.agents.pysmurf_controller.smurf_subprocess_util import (
+    RunCfg, RunResult, run_func_in_subprocess, NBIASLINES
+)
 
 
 class PysmurfScriptProtocol(protocol.ProcessProtocol):
@@ -76,6 +77,15 @@ class PysmurfScriptProtocol(protocol.ProcessProtocol):
 
         self.deferred.callback(rc)
 
+
+def set_session_data(session, result: RunResult):
+    """Sets session data based on a RunResult object"""
+    if not result.success or result.return_val is None:
+        return
+    if isinstance(result.return_val, dict):
+        session.data = result.return_val
+    else:
+        session.data = {'data': result.return_val}
 
 class PysmurfController:
     """
@@ -331,6 +341,25 @@ class PysmurfController:
     def _stop_check_state(self, session, params):
         """Stopper for check state process"""
         session.set_status('stopping')
+    
+    def run_test_func(self, session, params):
+        """
+        Task to run subprocess test function
+        """
+        cfg =RunCfg(
+            func_name='test',
+        )
+        result = run_func_in_subprocess(cfg)
+        if not result.success:
+            self.log.error("Subprocess errored out:\n{tb}", tb=result.traceback)
+
+        if isinstance(result.return_val, dict):
+            session.data = result.return_val
+        else:
+            session.data['data'] = result.return_val
+
+        return result.success, 'finished'
+        
 
     @ocs_agent.param("duration", default=None, type=float)
     @ocs_agent.param('kwargs', default=None)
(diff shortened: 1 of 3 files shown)


socs/agents/pysmurf_controller/smurf_subprocess_util.py
```diff
@@ -0,0 +1,290 @@
...
+    Args
+    -----
+    cfg: RunCfg
+        Configuration object to specify the function to run, and the arguments.
+    """
+    if not in_reactor_context():
+        return threads.blockingCallFromThread(
+            reactor, _run_func_in_subprocess_reactor, cfg)
+    else:
+        return _run_func_in_subprocess_reactor(cfg)
```
mhasself, line 264: Cool... I believe that if you _are_ in reactor context, then you can't block -- i.e you need to return a Deferred here, which would only be possible if `_run_func_in_subprocess_reactor(cfg)` returned `prot.deferred` or some wrapped thing that applies the Runresult.

jlashner: Good point... I changed to `run_func_in_subprocess_from_thread` and had it only handle the from-thread case

jlashner:
This has been tested on satp1 and it works! It seems to have fixed the large memory leak. Notes from my testing is here: https://simonsobs.atlassian.net/wiki/spaces/~55705801070cc775254537ab15f2dcf6702b78/pages/412778522/2024-05-02+pysmurf-controller+subprocessing+tests

Notably, memory returns to base-level after IVs and bias steps, which it did not previously: 
image

I want to note that there does still seem to be a memory leak caused by the functions that have _not_ been moved to subprocesses. I did not move the `stream` data function to a subprocess because it isn't loading TODs, however just the creation of the pysmurf object seems to be leaking memory, at a rate of ~15 MB per call:
image

I think we should probably move _all logic_ involving pysmurf to a subprocess, however I this this PR solves the most pressing issues and largest memory leaks, so I think we should merge this first.

BrianJKoopman (changes requested):
Glad the subprocess route is working! Can you address the unused imports caught by pre-commit: https://results.pre-commit.ci/run/github/186511668/1714677193.715HK9dWRnKVTIpCyOTUcg

socs/agents/pysmurf_controller/smurf_subprocess_util.py
```diff
@@ -0,0 +1,336 @@
...
+    }
+    quantiles = np.array([15, 25, 50, 75, 85])
+    data['quantiles'] = _process_quantiles(quantiles, arrays)
+    return data
+
+
+def test():
+    """Function for testing subprocess operation"""
+    print("HEREE")
+    return {'test': 10}
```
BrianJKoopman, line 193: Remove? Especially if subprocess task dropped from Agent code.

socs/agents/pysmurf_controller/smurf_subprocess_util.py
```diff
@@ -0,0 +1,336 @@
...
+    def processExited(self, status):
+        if self.result is None:
+            raise RuntimeError("No result received from child process")
+
+        self.deferred.callback(self.result)
+
+
+@defer.inlineCallbacks
+def _run_func_in_subprocess_reactor(cfg: RunCfg) -> RunResult:
+    """Helper function for run_func_in_subprocess, that can assume reactor context"""
```
BrianJKoopman, line 270: Is `run_func_in_subprocess` now `run_smurf_func`? If not, where is this function?

BrianJKoopman (commented):
This looks good to me, a few small comments/questions below.

socs/agents/pysmurf_controller/agent.py
```diff
@@ -332,6 +342,24 @@ def _stop_check_state(self, session, params):
...
+        if not result.success:
+            self.log.error("Subprocess errored out:\n{tb}", tb=result.traceback)
+
+        if isinstance(result.return_val, dict):
+            session.data = result.return_val
+        else:
+            session.data['data'] = result.return_val
+
+        return result.success, 'finished'
+
```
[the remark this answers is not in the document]
jlashner: I do want to keep this since it is helpful for testing locally, but I will correct docstrings.

BrianJKoopman (commented):
Addressed comments look good, merge conflicts still need resolving though.

BrianJKoopman (approved):

PR #683, opened 2024-05-29
Modification the max current setting of the kikusui agent (wire grid)

hnakata-JP:
## Description

In this pull request, I changed the max current setting about the wire grid kikusui agent. It was changed from 3.0 A to 4.5 A.
The kikusui power supply itself supports 18 V-5 A max in its specification. And this change will increase the motor's max torque, so the wire grid can rotate under lower temperatures.

## Motivation and Context

Now the wire grid used for SAT-MF2 has an issue in that it cannot rotate under low ambient temperature, even though the grease in its bearing has temperature resistance. When I checked the rotation directly, the rotator can move with a small additional torque support by hand. Then, to make the rotation robust, we want to change the max current setting to 4.5 A.

## How Has This Been Tested?

At the site, I changed the settings of the kikusui power supply manually, and confirmed the grid can rotate at 4.5 A max. Further, the over current protection in the hardware is now set at 4.9 A. So this change makes the rotation easy for the OCS operation as well.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/wiregrid_kikusui/agent.py
+++ b/socs/agents/wiregrid_kikusui/agent.py
@@ -289,14 +289,14 @@ class WiregridKikusuiAgent:
             return True, 'Set Kikusui off'
 
     @ocs_agent.param('current', default=0., type=float,
-                     check=lambda x: 0.0 <= x <= 3.0)
+                     check=lambda x: 0.0 <= x <= 4.5)
     def set_c(self, session, params):
         """set_c(current=0)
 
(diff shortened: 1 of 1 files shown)


sadachi5 (approved):

BrianJKoopman (approved):
This looks good to me, I'm assuming since you've tested this that 12V is high enough to drive the new higher current.

hnakata-JP:
12 V is the value determined by the company. It should be enough for the motor.

PR #684, opened 2024-05-30
Move HWP encoder timeout packet log to debug level

BrianJKoopman:
## Description

This PR changes the log level of the 'Received timeout packet.' message in the HWP Encoder agent from 'error' to 'debug'. If the HWP is not spinning, this packet is expected as normal behavior.

I also add the setup that allows changing the log level via environment variable, so it can be changed in the Docker Compose config. 

## Motivation and Context

When the HWP is not spinning, this packet is received every 10 seconds, filling the logs with this message. This makes it difficult to see any other messages that might be more relevant for debugging.

## How Has This Been Tested?

Not yet tested, though is a straightforward change. The log level selection is used in other agents.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_encoder/agent.py
+++ b/socs/agents/hwp_encoder/agent.py
@@ -52,6 +52,7 @@ HWPEncoder_full: separated feed for full-sample HWP encoder data,
 
 import argparse
 import calendar
+import os
 import select
 import socket
 import struct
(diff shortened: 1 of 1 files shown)


BrianJKoopman:
> This is okay, but can you explain exactly how we can change the log level through the Docker Compose config?

Yes, if you add this line to the environment block:
```
environment:
  - "LOGLEVEL=debug"
```

Also described here, for future reference: https://ocs.readthedocs.io/en/main/developer/agent_references/logging.html

PR #688, opened 2024-06-06
HWP ACU checks for spin-up

jlashner:
This PR adds a check of the ACU elevation before allowing for spinup of the HWP.

## Description

Adds ACU state information to the `HWPState` class. This state info is used by the control-state `update` function to check the ACU elevation and last update time before enabling spin-up.

## Motivation and Context

Motivation is described in Kyohei's discussion here: https://github.com/simonsobs/daq-discussions/discussions/96.

## How Has This Been Tested?

This has not yet been tested! @ykyohei can we test this on one of the SATs? To test at a safe elevation, we can raise the `acu_min_el` param.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -140,25 +140,40 @@ class IBootState:
 
 @dataclass
 class ACUState:
+    """
+    Class containg ACU state information.
+
+    Args
+    ------
+    instance_id : str
+        Instance ID of ACU agent
+    min_el : float
+        Minimum elevation allowed before restricting spin-up [deg]
+    max_el : float
+        Maximum elevation allowed before restricting spin-up [deg]
+    max_time_since_update : float
+        Maximum time since last update before restricting spin-up[sec]
+
+    Attributes
+    ------------
+    el_current_position : float
+        Current el position [deg]
+    el_commanded_position : float
+        Commanded el position [deg]
+    el_current_velocity : float
+        Current el velocity [deg/s]
+    last_updated : float
+        Time of last update [sec]
+    """
     instance_id: str
-
     min_el: float
-    "Minimum elevation allowed before restricting spin-up [deg]"
-
+    max_el: float
     max_time_since_update: float
-    "Maximum time since last update before restricting spin-up[sec]"
 
     el_current_position: Optional[float] = None
-    "Current el position [deg]"
-
     el_commanded_position: Optional[float] = None
-    "Commanded el position [deg]"
-
     el_current_velocity: Optional[float] = None
-    "Current el velocity [deg/s]"
-
     last_updated: Optional[float] = None
-    "Time of last reading"
 
     def update(self):
         op = get_op_data(self.instance_id, 'monitor')
@@ -760,7 +775,7 @@ class ControlStateMachine:
                 self.log.info("pid state: {data}", data=data)
                 return data
 
-            def check_acu_ok():
+            def check_acu_ok_for_spinup():
                 acu = hwp_state.acu
                 if acu is not None:
                     if acu.last_updated is None:
@@ -768,11 +783,12 @@ class ControlStateMachine:
                     tdiff = time.time() - acu.last_updated
                     if tdiff < acu.max_time_since_update:
                         raise RuntimeError(f"ACU state has not been updated in {tdiff} sec")
-                    if acu.el_current_position < acu.min_el_allowed:
-                        raise RuntimeError(f"ACU elevation is below minimum allowed: {acu.el_current_position} deg")
+                    if not (acu.min_el <= acu.el_current_position <= acu.max_el):
+                        raise RuntimeError(f"ACU elevation is {acu.el_current_pos} deg, "
+                                           f"outside of allowed range ({acu.min_el}, {acu.max_el})")
 
             if isinstance(state, ControlState.PIDToFreq):
-                check_acu_ok()
+                check_acu_ok_for_spinup()
                 self.run_and_validate(clients.pid.set_direction,
                                       kwargs={'direction': state.direction})
                 self.run_and_validate(clients.pid.declare_freq,
@@ -866,7 +882,7 @@ class ControlStateMachine:
 
             elif isinstance(state, ControlState.ConstVolt):
                 if state.voltage > 0:
-                    check_acu_ok()
+                    check_acu_ok_for_spinup()
                 self.run_and_validate(clients.pmx.set_on)
                 self.run_and_validate(clients.pid.set_direction,
                                       kwargs={'direction': state.direction})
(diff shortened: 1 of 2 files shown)


ykyohei:
Thank you so much for doing this quickly. Yes, we can test this.

I have a question. Do you plan to implement the constraints between gripper and ACU by implementing it directly in the gripper agent? Or do you plan to add gripper control in the supervisor function?

socs/agents/hwp_supervisor/agent.py
```diff
@@ -720,7 +776,20 @@ def query_pid_state():
                 self.log.info("pid state: {data}", data=data)
                 return data
 
+            def check_acu_ok_for_spinup():
+                acu = hwp_state.acu
+                if acu is not None:
+                    if acu.last_updated is None:
+                        raise RuntimeError(f"No ACU data has been received from instance-id {acu.instance_id}")
+                    tdiff = time.time() - acu.last_updated
+                    if tdiff < acu.max_time_since_update:
```
mhasself, line 785: Sign error

socs/agents/hwp_supervisor/agent.py
```diff
@@ -182,6 +236,8 @@ def from_args(cls, args: argparse.Namespace):
             self.gripper_iboot = IBootState(args.gripper_iboot_id, args.gripper_iboot_outlets)
         if args.driver_iboot_id is not None:
             self.driver_iboot = IBootState(args.driver_iboot_id, args.driver_iboot_outlets)
+        if args.acu_instance_id is not None:
+            self.acu = ACUState(args.acu_instance_id, args.acu_min_el, args.acu_max_time_since_update)
```
mhasself, line 240: Needs acu_max_el, right?

Also, isn't it generally wise to use kwargs?

socs/agents/hwp_supervisor/agent.py
```diff
@@ -1457,6 +1528,23 @@ def make_parser(parser=None):
         '--gripper-iboot-outlets', nargs='+', type=int,
         help="Outlets for gripper iboot power")
 
+    pgroup.add_argument(
+        '--acu-instance-id', help="Instance ID for the ACU agent"
+    )
```
mhasself, line 1533: Somewhere, maybe many places, you need to mention that passing this is necessary for the ACU check to be active.  I.e. it's not sufficient to specify min-el and max-el.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -138,6 +138,58 @@ def update(self):
         }
 
 
+@dataclass
+class ACUState:
+    """
+    Class containg ACU state information.
```
ykyohei, line 144: containing

socs/agents/hwp_supervisor/agent.py
```diff
@@ -182,6 +236,8 @@ def from_args(cls, args: argparse.Namespace):
             self.gripper_iboot = IBootState(args.gripper_iboot_id, args.gripper_iboot_outlets)
         if args.driver_iboot_id is not None:
             self.driver_iboot = IBootState(args.driver_iboot_id, args.driver_iboot_outlets)
+        if args.acu_instance_id is not None:
```
mhasself, line 239: I would want to see log messages for each of these (not just ACU):
"Checking ACU elevation range is [NOT active / active at instance_id={instance_id}]"
etc.

When something goes wrong, whether the agent though these things were on, or not, is very useful log output.

mhasself (changes requested):

ykyohei (commented):
`el_commanded_position` is monitored but not used for `check_acu_ok_for_spinup`. 
I think we should check this too.

jlashner:
@ykyohei My intention is to implement grip / ungrip procedures (or at least a wrapper for the gripper agent ops) in the supervisor, and to add the acu checks here. That will be the next PR.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -171,17 +223,41 @@ class HWPState:
...
             temp_field=args.ybco_temp_field,
             temp_thresh=args.ybco_temp_thresh,
             ups_minutes_remaining_thresh=args.ups_minutes_remaining_thresh,
         )
+
         if args.gripper_iboot_id is not None:
             self.gripper_iboot = IBootState(args.gripper_iboot_id, args.gripper_iboot_outlets)
+            log.info("Gripper Ibootbar id set: {id}", id=args.gripper_iboot_id)
+        else:
+            log.warning("Gripper Ibootbar id not set")
```
ykyohei, line 241: log.warn not log.warning, same typo in several lines later

socs/agents/hwp_supervisor/agent.py
```diff
@@ -138,6 +138,58 @@ def update(self):
...
+            return
+
+        d = op['data'].get("StatusDetailed")
+        if d is None:
+            return
+
+        self.el_current_position = d['Elevation current position']
+        self.el_commanded_position = d['Elevation commanded position']
+        self.el_current_velocity = d['Elevation current velocity']
+        self.last_updated = time.time()
```
mhasself, line 190: The timestamp should come from the ACU dataset.  Or at least be rejected if doesn't make sense.

I acknowledge the ACU session data isn't making that very easy... can you change this to
```
last_updated = d.get('timestamp_agent')
```
and we'll get that field added into the agent session.data separately?

mhasself (changes requested):
Just one more inline comment from me ... 

And my outline comment is that you should probably add some discussion of this feature to the docs, or else make the docstrings and argument descriptions clearer.  I can understand "Minimum elevation allowed before restricting spin-up" because I already know what this code is for... but I think there are clearer ways to say this.  ("Elevation below which HWP spin-up is not permitted", for example)

socs/agents/hwp_supervisor/agent.py
jlashner: Yep! Will `timestamp_agent` sometimes not exist / should I handle None values by just not changing `last_updated`?

mhasself: Right, sorry, I meant that you could then replace None with time.time().  That's a compromise until the ACU agent is updated.  Once ACU agent is updated, I would expect timestamp_agent to be either absent (during startup) or a float.

ykyohei:
I tested the acu_check in daq-dev and made small debugs. I believe it's working now.

jlashner:
Great, thanks Kyohei!

I just implemented Matthew's suggestion for acquiring the last_updated timestamp from ACU session data, and also added a brief section to the docs on ACU safety checks that we can expand as we add the gripper operations and checks.

mhasself (approved):
Thanks for the additional clarifications.

BrianJKoopman (commented):
This seems ready to merge, but there are some merge conflicts in the agent code. Can you resolve these @jlashner?

jlashner:
Looks like tests are failing due to this issue: https://github.com/simonsobs/ocs/issues/391

Do we need to pin setup-tools in socs or upgrade the ocs version?

BrianJKoopman:
> Looks like tests are failing due to this issue: [simonsobs/ocs#391](https://github.com/simonsobs/ocs/issues/391)
> 
> Do we need to pin setup-tools in socs or upgrade the ocs version?

I was thinking the later, but neither option is working...I might end up switching build systems, was trying to make that decision yesterday.

Issue #691, opened 2024-06-11
HWPSupervisor crashes on startup when using synaccess for `driver-iboot-id`


BrianJKoopman:
On satp3 after maintenance they're having trouble starting up the HWP Supervisor agent, it crashes immediately with:
```
Args: ['--instance-id', 'hwp-supervisor', '--site-hub', 'ws://127.0.0.1:8005/ws', '--site-http', 'http://127.0.0.1:8005/call']
Installed OCS Plugins: ['socs', 'ocs']
Renaming this process to: "ocs-agent:hwp-supervisor"
2024-06-11T20:37:41+0000 Using OCS version 0.11.0
2024-06-11T20:37:41+0000 ocs: starting <class 'ocs.ocs_agent.OCSAgent'> @ satp3.hwp-supervisor
2024-06-11T20:37:41+0000 log_file is apparently None
2024-06-11T20:37:41+0000 Setting state: ControlState.Idle()
2024-06-11T20:37:41+0000 transport connected
2024-06-11T20:37:41+0000 session joined: {'authextra': {'x_cb_node': '4e2f32e9b379-6',
               'x_cb_peer': 'tcp4:172.19.0.1:53870',
               'x_cb_pid': 13,
               'x_cb_worker': 'worker001'},
 'authid': 'XGSA-KE73-GR6Y-QNXJ-JNFL-WL4V',
 'authmethod': 'anonymous',
 'authprovider': 'static',
 'authrole': 'iocs_agent',
 'realm': 'test_realm',
 'resumable': False,
 'resume_token': None,
 'resumed': False,
 'serializer': 'cbor.batched',
 'session': 3582372078905339,
 'transport': {'channel_framing': 'websocket',
               'channel_id': {},
               'channel_serializer': None,
               'channel_type': 'tcp',
               'http_cbtid': None,
               'http_headers_received': None,
               'http_headers_sent': None,
               'is_secure': False,
               'is_server': False,
               'own': None,
               'own_fd': -1,
               'own_pid': 8,
               'own_tid': 8,
               'peer': 'tcp4:127.0.0.1:8005',
               'peer_cert': None,
               'websocket_extensions_in_use': None,
               'websocket_protocol': None}}
2024-06-11T20:37:41+0000 startup-op: launching monitor
2024-06-11T20:37:41+0000 start called for monitor
2024-06-11T20:37:41+0000 monitor:0 Status is now "starting".
2024-06-11T20:37:41+0000 startup-op: launching spin_control
[...]

Issue #681, opened 2024-05-29
HWPSupervisor: timestamps for encoder readings


mhasself:
In supervisor, the hwp_state is captured through certain data from hwp_encoder agent:
```
	self._update_from_keymap(op, {
            'enc_freq': 'approx_hwp_freq',
            'last_quad': 'last_quad',
            'last_quad_time': 'last_quad_time',
	})
```
In the case that `quad` and `last_quad_time` are not valid (ahem), then this dataset doesn't include a timestamp.

I believe the supervisor should capture (and check!!) the key `encoder_last_updated` ([from here](https://github.com/simonsobs/socs/blob/5bcf3e785d00527ea6391df81faab35ffbeed69c/socs/agents/hwp_encoder/agent.py#L670)).

I am going to change ocs-web panel to report the timestamp from encoder_last_updated, in anticipation of this being added ...

mhasself:
It looks like last_quad and last_quad_time are sometimes just not populated.  So I think they will be valid (but potentially stale), or absent.

PR #693, opened 2024-06-11
hwp-supervisor: make  IBootState compatible with synaccess.

ykyohei:
Make IBootState compatible with synaccess agent originally it was only compatible with ibootbar agent.

## Description
Make IBootState compatible with synaccess agent and add encoder timestamp
This PR addresses https://github.com/simonsobs/socs/issues/691 and https://github.com/simonsobs/socs/issues/681

## Motivation and Context
https://github.com/simonsobs/socs/issues/691 and https://github.com/simonsobs/socs/issues/681

## How Has This Been Tested?
Tested in daq-dev for satp3

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -139,7 +139,7 @@ class IBootState:
                 for outlet, label in self.outlet_labels.items()
             }
         elif self.driver_power_agent_type == 'synaccess':
-            self.outlet_labels = {o: str(o-1) for o in self.outlets}
+            self.outlet_labels = {o: str(o - 1) for o in self.outlets}
             self.outlet_state = {
                 outlet: op['data']['fields'][label]['status']
                 for outlet, label in self.outlet_labels.items()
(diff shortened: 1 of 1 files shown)


socs/agents/hwp_supervisor/agent.py
```diff
@@ -118,12 +118,12 @@ class HWPClients:
 class IBootState:
     instance_id: str
     outlets: List[int]
+    driver_power_agent_type: Literal['iboot, synaccess']
```
jlashner, line 121: Can we change this to just `agent_type`? The `driver-power-agent-type` arg is for specifying the power-agent type of the HWP driver board, so in the IBootState class we don't need those first two descriptors.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -179,9 +192,11 @@ def from_args(cls, args: argparse.Namespace):
             ups_minutes_remaining_thresh=args.ups_minutes_remaining_thresh,
         )
         if args.gripper_iboot_id is not None:
-            self.gripper_iboot = IBootState(args.gripper_iboot_id, args.gripper_iboot_outlets)
+            self.gripper_iboot = IBootState(args.gripper_iboot_id, args.gripper_iboot_outlets,
+                                            args.driver_power_agent_type)
```
jlashner, line 196: Here we'll need a new command-line arg `args.gripper_power_agent_type` which is not necessarily the same as the driver power type.

jlashner (changes requested):
Hi Kyohei, this seems good to me! I have a few requests inline. Do we know why this is suddenly not working for SATp3? As far as I know, nothing in the configuration has changed recently.

ykyohei:
Right, the configuration is not changed, and I'm confused why it was working before. HWP agents skips IBootState monitoring if synaccess agent is down when HWP agent is brought up, so we might have been in such an edge case. But I couldn't completely reproduce the previous state.

jlashner (approved):
Thanks for the changes!

PR #703, opened 2024-07-29
hw_pid: delete initial set_direction

ykyohei:
Delete initial hwp rotation direction setting of hwp_pid agent to resolve the hwp rotation direction ambiguities.

## Description
The "direction" in the hwp_pid memory register doesn't reflect the actual action of pid agent until hwp_pid.tune_freq is called. The default setting of "direction" on the agent startup lead to the ambiguity of hwp rotation direction. This PR delete the default setting. 

As long as we use `hwp_supervisor.pid_to_freq`, we can avoid this issue. https://github.com/simonsobs/socs/blob/4902d3be867b51ea77514825d5a301581fb1828b/socs/agents/hwp_supervisor/agent.py#L739-L750
However, if `hwp_supervisor.pid_to_freq` fails in the middle of its command sequence, we may still have a potential ambiguity in rotation direction, so maybe we should make it more robust?

## Motivation and Context
https://github.com/simonsobs/chwp-discussions/discussions/24

## How Has This Been Tested?
This has been tested by daq-dev for satp3. Confirmed that the restart of hwp_pid agent doen't lead to the direction ambiguity in both directions.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pid/drivers/pid_controller.py
+++ b/socs/agents/hwp_pid/drivers/pid_controller.py
@@ -26,11 +26,8 @@ class PID:
         self.ip = ip
         self.port = port
         self.hex_freq = '00000'
-        self.direction = None
-        self.target = 0
         # Need to setup connection before setting direction
         self.conn = self._establish_connection(self.ip, int(self.port))
-        self.set_direction('0')
 
     @staticmethod
     def _establish_connection(ip, port, timeout=2):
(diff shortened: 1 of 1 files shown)


mhasself:
Since you've removed the .direction attribute, update the class and function docstrings to not mention it.

BrianJKoopman:
The build error is unrelated. Working on a fix upstream.

bbixler500 (approved):

PR #705, opened 2024-07-31
Dump configs on IV

jlashner:
Dumps smurf configs on pysmurf-controller take_iv.
## Description

Dumps smurf configs on pysmurf-controller take_iv.

## Motivation and Context

During the smurf software town hall meeting I mistakenly claimed that smurf configs were dumped every operation. After checking the data, this turns out not to have been the case, and configs were only dumped each hammer. This changes it so that configs are dumped every IV, which I believe is roughly once a day. I think this is a good middle ground between every operation and every hammer.

## How Has This Been Tested?

This has not been tested.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/pysmurf_controller/smurf_subprocess_util.py
+++ b/socs/agents/pysmurf_controller/smurf_subprocess_util.py
@@ -83,7 +83,7 @@ def encode_dataclass(obj):
     return json.dumps(data).encode()
 
 
-def get_smurf_control():
+def get_smurf_control(dump_configs=False):
     """
     Get the SMuRF control object and sodetlib configuration object for the
     current slot
@@ -91,7 +91,7 @@ def get_smurf_control():
     slot = os.environ['SLOT']
     cfg = DetConfig()
     cfg.load_config_files(slot=slot)
-    S = cfg.get_smurf_control()
+    S = cfg.get_smurf_control(dump_configs=dump_configs)
     S.load_tune(cfg.dev.exp['tunefile'])
     return S, cfg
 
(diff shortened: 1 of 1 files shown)


msilvafe (approved):
We should pay attention to how the size of the logs folder grows in time

BrianJKoopman (commented):
This seems good to merge, but I was wondering if you planned on testing first.

jlashner:
It would be good to test, but may be difficult to get time at site. @dpdutcher or @tristpinsm, is there any chance we can test changes like this in a testbed at princeton or SLAC?

dpdutcher:
In theory yes, though right now isn't the best as we're about to have a 2-week DR shutdown due to building maintenance.
My software is also out-of-date, I may update it while the DRs are down.

tristpinsm:
We currently have a UFM in the DR at SLAC, and before heading down to Chile, I think @tpsatt had set up a smurf server with the SO software. It might be a bit of work to get a test setup running as we haven't been using OCS before (AFAIK), but I don't see why we couldn't do it there. Right now Toby and I are busy at site.

msilvafe:
> We currently have a UFM in the DR at SLAC, and before heading down to Chile, I think @tpsatt had set up a smurf server with the SO software. It might be a bit of work to get a test setup running as we haven't been using OCS before (AFAIK), but I don't see why we couldn't do it there. Right now Toby and I are busy at site.

FYI, I used OCS for testing at SLAC in the past, so hopefully not too much work to revive

dpdutcher:
Meh? It wouldn't hurt.
My preferred solution would be to have [sodetlib.util.get_metadata](https://github.com/simonsobs/sodetlib/blob/87211e780ce733fe4f7941658a773481aceaee22/sodetlib/util.py#L188) do a version of this instead, so all config info is present in the sodetlib data files to which they pertain. But merging this is easier.

msilvafe:
> Meh? It wouldn't hurt. My preferred solution would be to have [sodetlib.util.get_metadata](https://github.com/simonsobs/sodetlib/blob/87211e780ce733fe4f7941658a773481aceaee22/sodetlib/util.py#L188) do a version of this instead, so all config info is present in the sodetlib data files to which they pertain. But merging this is easier.

Ya happy to defer to @dpdutcher on this one.

tristpinsm:
I don't really know what this is about though, so I will need to read through and get up to speed.

PR #708, opened 2024-08-02
Adds grip and ungrip fns to hwp-supervisor with safety checks

jlashner:
Adds grip and ungrip fns to hwp-supervisor with safety checks.

This PR adds two additional operations to the HWP supervisor to grip/ungrip the HWP

## Description

For these two actions, the HWP and ACU states are first checked based on [kyohei's daq-discussion.](https://github.com/simonsobs/daq-discussions/discussions/96)
If checks pass, the gripper agent grip/ungrip operations will be called.

## Motivation and Context

This is to centralize the gripping procedure and to implement safety checks.

## How Has This Been Tested?

This has not been tested yet. @ykyohei any chance we can test at site sometime soon?

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -806,7 +806,7 @@ class ControlAction:
 
 
 @contextmanager
-def ensure_grip_safety(hwp_state: HWPState) -> Generator[None, None, None]:
+def ensure_grip_safety(hwp_state: HWPState, log: txaio.ILogger) -> Generator[None, None, None]:
     """
     Run required checks for gripper safety. This will check ACU parameters such
     as az/el position and velocity. If `use_acu_blocking` is set, this will
@@ -821,6 +821,7 @@ def ensure_grip_safety(hwp_state: HWPState) -> Generator[None, None, None]:
         Timeout for waiting for the ACU blockout before an error will be raised.
     """
     now = time.time()
+
     if hwp_state.pid_current_freq is None or hwp_state.pid_last_updated is None:
         raise RuntimeError("Cannot determine current HWP Freq")
 
@@ -830,6 +831,7 @@ def ensure_grip_safety(hwp_state: HWPState) -> Generator[None, None, None]:
 
     if np.abs(hwp_state.pid_current_freq) > 0.02:
         raise RuntimeError("Cannot grip HWP while spinning")
+    log.info("Rotation safety checks have passed")
 
     acu = hwp_state.acu
     if acu is None:
@@ -876,6 +878,7 @@ def ensure_grip_safety(hwp_state: HWPState) -> Generator[None, None, None]:
             f"boresight commanded angle of {acu.bor_commanded_position} deg, "
             f"above threshold of {acu.grip_max_boresight_angle} deg"
         )
+    log.info("ACU safety checks have passed")
 
     if not acu.use_acu_blocking:
         yield
@@ -883,12 +886,15 @@ def ensure_grip_safety(hwp_state: HWPState) -> Generator[None, None, None]:
 
     try:  # If use_acu_blocking, do handshake with ACU agent to block motino
         acu.set_request_block_motion(True)
+        log.info("Requesting ACU to block motion")
         while not acu.ACU_motion_blocked:
             if time.time() - acu.request_block_motion_timestamp > acu.block_motion_timeout:
                 raise RuntimeError("ACU motion was not blocked within timeout")
             time.sleep(1)
+        log.info("ACU motion has been blocked")
         yield
     finally:
+        log.info("Releasing ACU motion block")
         acu.set_request_block_motion(False)
 
 
@@ -1088,12 +1094,12 @@ class ControlStateMachine:
                 self.action.set_state(ControlState.Done(success=state.success))
 
             elif isinstance(state, ControlState.GripHWP):
-                with ensure_grip_safety(hwp_state):
+                with ensure_grip_safety(hwp_state, self.log):
                     self.run_and_validate(clients.gripper.grip)
                 self.action.set_state(ControlState.Done(success=True))
 
             elif isinstance(state, ControlState.UngripHWP):
-                with ensure_grip_safety(hwp_state):
+                with ensure_grip_safety(hwp_state, self.log):
                     self.run_and_validate(clients.gripper.ungrip)
                 self.action.set_state(ControlState.Done(success=True))
 
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -301,7 +301,7 @@ class HWPState:
         else:
             log.warn("Driver Ibootbar id not set")
 
-        if args.acu_instance_id is not None:
+        if not args.no_acu:
             self.acu = ACUState(
                 instance_id=args.acu_instance_id,
                 min_el=args.acu_min_el,
@@ -315,7 +315,7 @@ class HWPState:
             log.info("ACU state checking enabled: instance_id={id}",
                      id=self.acu.instance_id)
         else:
-            log.info("ACU state checking disabled.")
+            log.warn("The no-acu option has been set. ACU state checking disabled.")
 
         return self
 
@@ -886,7 +886,7 @@ def ensure_grip_safety(hwp_state: HWPState) -> Generator[None, None, None]:
 
 
 class ControlStateMachine:
-    def __init__(self):
+    def __init__(self) -> None:
         self.action: ControlAction = ControlAction(ControlState.Idle())
         self.action_history: List[ControlAction] = []
         self.max_action_history_count = 100
@@ -959,7 +959,7 @@ class ControlStateMachine:
                     if tdiff > acu.max_time_since_update:
                         raise RuntimeError(f"ACU state has not been updated in {tdiff} sec")
                     if not (acu.min_el <= acu.el_current_position <= acu.max_el):
-                        raise RuntimeError(f"ACU elevation is {acu.el_current_pos} deg, "
+                        raise RuntimeError(f"ACU elevation is {acu.el_current_position} deg, "
                                            f"outside of allowed range ({acu.min_el}, {acu.max_el})")
                     if not (acu.min_el <= acu.el_commanded_position <= acu.max_el):
                         raise RuntimeError(f"ACU commanded elevation is {acu.el_commanded_position} deg, "
@@ -1783,9 +1783,13 @@ def make_parser(parser=None):
         help="Type of agent used for controlling the gripper power")
 
     pgroup.add_argument(
-        '--acu-instance-id',
-        help="Instance ID for the ACU agent. This is required for checks of ACU "
-             "postiion and velocity before HWP commands."
+        '--acu-instance-id', default='acu',
+        help="Instance ID for the ACU agent."
+    )
+    pgroup.add_argument(
+        '--no-acu', action='store_true',
+        help="If set, will not attempt to connect to the ACU or perform any ACU "
+             "checks for grip and spin-up"
     )
     pgroup.add_argument(
         '--acu-min-el', type=float, default=48.0,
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Will take a look soon. In the meantime, merging in the latest `main` should fix the build error. See #709.

ykyohei (commented):
Thank you for doing this quickly, we will test this.

- We would like to have --acu-min/max-roll (or boresight, bs?) for gripper operation.
- We would like to monitor acu.roll_current/commanded_position for gripper operation.
- Can current hwp-supervisor provide information that the gripper functions are running to acu, so that acu can restrict its operation?

jlashner:
Hi Kyohei, I added the boresight checks and exposed the supervisor control state in several places so it can be checked by the ACU. I think the following is the one with the least amount of latency (~1 sec)...
```
# If this is GripHWP or UngripHWP, don't move the ACU
control_state_type = supervisor.spin_control.status().session['data']['current_state_type']
```
However I don't think this completely resolves the problem, since there is still a race condition if ACU movement and Grip/Ungrip are called at the same time (or within ~1 sec of each other)... there are ways that I can reduce the latency at which the control state session info is updated, but if this is something we really care about, we may want to thing about using a filesystem mutex to prevent simultaneous gripper/acu control (since I believe both agents are running on the same node). @mhasself do you think this may be worth doing or do you think that's overkill?

mhasself:
> Hi Kyohei, I added the boresight checks and exposed the supervisor control state in several places so it can be checked by the ACU. I think the following is the one with the least amount of latency (~1 sec)...
> 
> ```
> # If this is GripHWP or UngripHWP, don't move the ACU
> control_state_type = supervisor.spin_control.status().session['data']['current_state_type']
> ```
> 
> However I don't think this completely resolves the problem, since there is still a race condition if ACU movement and Grip/Ungrip are called at the same time (or within ~1 sec of each other)... there are ways that I can reduce the latency at which the control state session info is updated, but if this is something we really care about, we may want to thing about using a filesystem mutex to prevent simultaneous gripper/acu control (since I believe both agents are running on the same node). @mhasself do you think this may be worth doing or do you think that's overkill?

I would prefer to use an OCS-based handshake rather than some other system (e.g. filesystem).  That would mean:
- HWPSupervisor sets some flag "block_ACU_motion" flag, and waits...
- When ACU "monitor" process sees the flag, it quickly assesses whether it is stationary (i.e. move_to / scan aren't running), and if so it (a) locks out further motion and (b) sets an acknowledgment bit, e.g. "motion_blocked_by_hwp_supervisor: true".  If it can't block motion right now, it doesn't.  But as long as the "block_motion" flag is true, it will keep assessing + trying to block out motion.
- The HWPSupervisor waits for "motion_blocked_by_hwp_supervisor" flag to be true.  That might take 2 seconds.  It might never happen.  HWPSupervisor decides whether to give up, or keep waiting.
- When the HWPSupervisor is finished its activity, it lowers the "block_ACU_motion" flag.
- When ACU sees block_ACU_motion flag lowered, it resumes normal operation.

That seem good?

The ACU agent modifications are not too difficult, but we will need a testing platform for a few hours.

jlashner:
I think that seems good, but should probably go into a separate PR so that we can have some basic form of protection in the meantime. @ykyohei do you think we can test this version of the branch at site?

mhasself:
> I think that seems good, but should probably go into a separate PR so that we can have some basic form of protection in the meantime. @ykyohei do you think we can test this version of the branch at site?

Fine with me but you _could_ put in that "block_ACU_motion" flag right away, if it's not already in here somewhere. (But not insist on hand-shake completion.)  Then ACU Agent can evolve + start implementing the block-out; then you can come back and implement the handshake fully to eliminate the race condition.

jlashner:
Hi Matthew, in the latest commit I added the `request_block_ACU_motion` flag to the HWP state, which should be accessible from session data of the `monitor` process, which I think should be enough for you to develop with. In response we'll need an operation that we can query session data for a `ACU_motion_blocked` field. The `ensure_grip_safety` context manager should implement the HWP-side of the handshake, though it won't be used unless a site-config var is set and I add an update for the `ACU_motion_blocked` state var.

Hope this is enough for now!

socs/agents/hwp_supervisor/agent.py
```diff
@@ -626,6 +659,18 @@ class PmxOff:
         """
         success: bool = True
 
+    @dataclass
+    class GripHWP:
```
ykyohei, line 663: I think having a force option which will ignore ensure_grip_safety is helpful.
Obviously we should not use it usually but it might be helpful when we really need to grip it but the site is in unusual state.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -750,6 +795,89 @@ def sleep_until_complete(self, session=None, dt=1):
...
+    also attempt to block out ACU motion for the duration of the operation.
+
+    Args
+    ------
+    hwp_state : HWPState
+        HWP state object. In addition to reading ACU state vars, this will
+        set the `request_block_ACU_motion` flag if `use_acu_blocking` is set.
+    timeout: float
+        Timeout for waiting for the ACU blockout before an error will be raised.
+    """
```
ykyohei, line 812: We need to check pid.last_updated is None or not.
Also we need to define max_time_since_update for hwp_state and do the similar check as acu.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -750,6 +795,89 @@ def sleep_until_complete(self, session=None, dt=1):
...
+    if hwp_state.pid_current_freq is None:
+        raise RuntimeError("Cannot determine current HWP Freq")
+
+    if np.abs(hwp_state.pid_current_freq) > 0.02:
+        raise RuntimeError("Cannot grip HWP while spinning")
+
+    acu = hwp_state.acu
```
ykyohei, line 819: I'm concerned about this logic. At least we should have a warning?

socs/agents/hwp_supervisor/agent.py
```diff
@@ -750,6 +795,89 @@ def sleep_until_complete(self, session=None, dt=1):
             time.sleep(dt)
 
 
+@contextmanager
+def ensure_grip_safety(hwp_state: HWPState, timeout: float = 60.):
```
ykyohei, line 799: I would like to have some more logs in this function.
For example, 'hwp rotation frequency safety checks have passed', 'acu safety checks have passed', 'start blocking acu motion' etc.

socs/agents/hwp_supervisor/agent.py
```diff
@@ -750,6 +795,89 @@ def sleep_until_complete(self, session=None, dt=1):
...
+    if hwp_state.pid_current_freq is None:
+        raise RuntimeError("Cannot determine current HWP Freq")
+
+    if np.abs(hwp_state.pid_current_freq) > 0.02:
+        raise RuntimeError("Cannot grip HWP while spinning")
+
+    acu = hwp_state.acu
+    if acu is None:
+        yield
+        return
```
ykyohei, line 822: I'm concerned about this logic. At least we should have a warning?

BrianJKoopman: I'm also wondering about the behavior here. Does this just skip the position checks for the ACU? Doesn't that mean we can end up in an unsafe position?

BrianJKoopman (commented):
This looks mostly OK to me, just a question about behavior when `acu is None`.

socs/agents/hwp_supervisor/agent.py
jlashner: Yes, so I don't think an ACU should be required to operate the HWP supervisor, and when an ACU is not present I think skipping checks (without printing warning logs every check) is the correct behavior. Currently, I do log an info statement [here](https://github.com/simonsobs/socs/blob/811858ad075671b2306c16e9a3edcfff579c8797/socs/agents/hwp_supervisor/agent.py#L310) but this could definitely be hard to catch if your not looking for it...

What if we add an additional command line argument `--no-acu` which needs to be set to bypass ACU checks? Then I can throw an error if `--acu-instance-id` is not passed without the `--no-acu` flag, to prevent SATs from accidentally forgetting to set the instance id.

BrianJKoopman: Yeah, reading [this Instrument Safety page](https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/433684484/CHWP+commissioning#Instrument-Safety) linked from https://github.com/simonsobs/daq-discussions/discussions/96, it does seem like the only relevant thing for grip/ungrip is not to move the platform while doing it. So I agree that not defining the ACU shouldn't stop this from working.

> What if we add an additional command line argument `--no-acu` which needs to be set to bypass ACU checks? Then I can throw an error if `--acu-instance-id` is not passed without the `--no-acu` flag, to prevent SATs from accidentally forgetting to set the instance id.

This sounds okay. If platforms are going to forget to add `--acu-instance-id`, they'll probably also forget to add `--no-acu` -- so I guess default to requiring checks. I wouldn't be against setting the default `--acu-instance-id` to 'acu' either, as that's the instance-id across all platforms, and I don't really see that changing -- if we really are going to forget to set that flag after this is merged.

socs/agents/hwp_supervisor/agent.py
jlashner: I can add this, but just want to make sure, are we sure this is something we really want? If the site is an unusual state it is still be possible to go in and manually run the command from the gripper agent, or even the windows software.

I'm not entirely sure we want to provide non-experts an easy way to ignore safety checks.

jlashner:
Hi @ykyohei and @BrianJKoopman, I think I addressed the remaining issues in the latest commits, with the exception of the force_grip option, which I want to double check this is something we actually need and want before implementing.

Now, the ACU instance ID will always default to "acu", and checks will _only_ be skipped if the `--no-acu` argument is specified. This is to make it so that the default behavior for the SATs will be to run with safety checks without having to change anything in their config.

If we could test this on an instrument that would be great!

socs/agents/hwp_supervisor/agent.py
ykyohei: That's fair point. I agree not to add it in hwp-supervisor.

socs/agents/hwp_supervisor/agent.py
ykyohei: Let me ask about this.
Do you think there is possibilities that hwp_supervisor tries to grip hwp while telescope is moving or fails to block acu motion? If acu is None, the telescope is not moving and will not move later?
Also, can hwp_supervisor perform safety checks when acu crashes and it is restarted later?

jlashner: `acu` will only ever be None if `the `--no-acu` flag is set in the site-config file. If an ACU instance id is set, but the agent is not found or not responding, I believe one of the other checks will fail and you will be unable to grip with the supervisor until the ACU agent is working again.

At the moment I'm not sure if we want to allow the supervisor to grip when the ACU agent is not responding. I think we should not allow it for the time being (since there are other methods to grip) and if it is a problem we can always come back and change the behavior.

ykyohei:
Thank you so much Jack, I tested this and safety checks look working.
For the acu blocking. if I add `--use-acu-blocking`, then I get `RuntimeError: ACU motion was not blocked within timeout`.
This is an expected behavior because hwp-acu handshake is not completed yet, right?
We should not add `--use-acu-blocking` until hwp-acu handshake is completed. Is this correct?

jlashner:
Yes that's right, the ACU blocking will not work until its implemented on the ACU side. Thanks for testing!

Though testing worked successfully, I am realizing that there is not currently a good way to test this without risk of it not working and potentially putting the telescope in danger. I think for this reason I should add a "--dry" mode where gripper and hwp operations are not actually conducted, we just log statements. I think lets merge this PR and then I will implement this separately.

socs/agents/hwp_supervisor/agent.py
ykyohei: I see. Thanks for the explanation.

ykyohei (approved):

BrianJKoopman (approved):

PR #715, opened 2024-08-05
Adding support for direct Ethernet SCPI PSU Agent 

### LINKED ISSUE #313: SCPI PSU Agent should initialize and monitor at startup
The SCPI PSU Agent should be updated to run the `init` task on startup, and optionally start the `monitor` process (which should be the default.) This is done in several other Agents, for example the [PfeifferTC400Agent](https://github.com/simonsobs/socs/blob/develop/agents/pfeiffer_tc400/pfeiffer_tc400_agent.py) or the [Lakeshore372Agent](https://github.com/simonsobs/socs/blob/develop/agents/lakeshore372/LS372_agent.py).

CAWNoodle:
Adding support for direct Ethernet SCPI PSU Agent 

## Description

- Adds the ScpiPsuInterface class to the socs/agents/scpi_psu/drivers.py file to allow sending SCPI commands directly over ethernet with a specified IP address and port.

- Introduces code switching based on device Model number as returned by the '*idn?' SCPI command.

- Adds the '--port' and '--mode' options to the agent's arguments. An 'acq' mode was added to conform to allow for automated collection of measurement data, conforming to the pattern of other SOCS agents.

- Allows backwards compatibility with the older Prologix Adapter interface by omitting the '--port' option.

## Motivation and Context
Many modern power supplies come equipped with an ethernet port or wifi connection. A new SCPI PSU agent was required to interface with such devices while maintaining backward compatibility for the legacy GPIB-to-Ethernet adapters. Furthermore, some means of distinguishing between various models (e.g. those with 3 outputs vs. single channel devices, devices with varying voltage and current limitation, and devices supporting different subsets of the SCPI protocol,etc.) was necessary in order to provide functionality without loss of flexibility.

Resolves #313.

## How Has This Been Tested?
With no access to a Prologix GPIB-to-Ethernet adapter, regression testing was unable to be performed. However, testing proceeded with small incremental changes to the driver first. Then the agent code was built testing each change as it was made. Testing consisted of forcing a --no-cache build of a local image before running 'docker-compose up' for the full OCS stack and tracing container execution through the log files. This agent was attached to a Keithley 2280S power supply with which initialization, data acquisition, current measurement, voltage measurement, current setting, voltage setting, output enabling and disabling functions, as well as tests of the output state, and model self-identification were all tested -- first within the agent and then via an OCSClient object handle in python.

Other agent container run alongside this one during testing include:
The OCS host manager
Influxdb and publisher
crossbar
grafana
Aggregator 
FakeData
Labjack
CryomechCPA
and other test agents

These tests have been useful for unit testing, but they are insufficient for regression or integration testing. I expect further edits will be required if this change is to be adopted by the wider community.

## Types of changes

## Checklist:


socs/agents/scpi_psu/agent.py
```diff
@@ -41,14 +44,37 @@ def init(self, session, params=None):
...
+            else:  # Use the new direct ethernet connection code
+                try:
+                    self.psu = ScpiPsuInterface(self.ip_address, self.gpib_slot, port=self.port)
+                    self.idn = self.psu.identify()
+                except socket.timeout as e:
+                    self.log.error(f"PSU timed out during connect: {e}")
+                    return False, "Timeout"
+                except ValueError as e:
+                    if (e.args[0].startswith('Model number')):
+                        self.log.error(f"PSU initialization error: {e}. Suggest appending {e.args[-1]} to the list of known model numbers in scpi_psu/drivers.py")
```
BrianJKoopman, line 63: This is a good suggestion, but can you wrap the line to <100 characters?

socs/agents/scpi_psu/drivers.py
```diff
@@ -1,6 +1,150 @@
...
+
+
+class ScpiPsuInterface:
+    def __init__(self, ip_address, gpibAddr, port, **kwargs):
+        self.ip_address = ip_address
+        self.gpibAddr = gpibAddr
+        self.port = port
+        self.sock = None
+        self.model = None
+        self.numChannels = 0
```
BrianJKoopman, line 20: Sorry for the nitpick comment, but conventionally we're trying to use `lower_case_with_underscores` for attributes. Can you replace `numChannels` with `num_channels` everywhere? (And apologies for the confusing use of `gpibAddr` in the old code.)

socs/agents/scpi_psu/agent.py
```diff
@@ -68,6 +94,7 @@ def monitor_output(self, session, params=None):
                 Defaults to False.
 
         """
+        session.set_status('running')
```
BrianJKoopman, line 97: This no longer needs to be set as of [ocs v0.11.0](https://github.com/simonsobs/ocs/releases/tag/v0.11.0), remove.

socs/agents/scpi_psu/agent.py
```diff
@@ -104,6 +136,31 @@ def stop_monitoring(self, session, params=None):
...
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                data = {
+                    'timestamp': time.time(),
+                    'block_name': 'output',
+                    'data': {}
+                }
+
+                data['data']['Voltage_{}'.format(chan)] = self.psu.get_volt(chan)
+                session.data = data
```
BrianJKoopman, line 159: The structure here is the one required for publishing to an [ocs Feed](https://ocs.readthedocs.io/en/main/developer/feeds.html#recorded-feed-message-format). This isn't really required for the `session.data`, which is very flexible. Since this is just an instance of retrieving a single number I'd suggest simplifying this down to:
```
data = {
    'timestamp': time.time(),
    'voltage': self.psu.get_volt(chan),
}

session.data = data
```

Whatever the format of the `session.data` dict, I'd ask that you put an example in the docstring for the method. (A good reference on documentation in OCS here: https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html#task-and-process-documentation -- and again sorry that this old agent isn't up to these newer standards, so the local context was a bit lacking...)

socs/agents/scpi_psu/agent.py
```diff
@@ -125,6 +182,30 @@ def set_voltage(self, session, params=None):
...
+        chan = params['channel']
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                data = {
+                    'timestamp': time.time(),
+                    'block_name': 'output',
+                    'data': {}
+                }
+                data['data']['Current_{}'.format(chan)] = self.psu.get_curr(chan)
+                session.data = data
```
BrianJKoopman, line 204: Same comment as above in `get_voltage()`.

socs/agents/scpi_psu/agent.py
```diff
@@ -145,6 +226,26 @@ def set_current(self, session, params=None):
...
         return True, 'Set channel {} current to {}'.format(params['channel'], params['current'])
 
+    @ocs_agent.param('channel', type=int, choices=[1, 2, 3])
+    def get_output(self, session, params=None):
+        """get_output(channel)
+
+        **Task** - Check if channel ouput is enabled or disabled.
+
+        Parameters:
+            channel (int):
```
BrianJKoopman, line 236: This looks like it got dropped, but let's just use the same description as the other added methods:
```
channel (int): Channel number (1, 2, or 3).
```

socs/agents/scpi_psu/agent.py
```diff
@@ -145,6 +226,26 @@ def set_current(self, session, params=None):
...
+        enabled = False
+        with self.lock.acquire_timeout(1) as acquired:
+            if acquired:
+                enabled = self.psu.get_output(params['channel'])
+            else:
+                return False, "Could not acquire lock."
+        if enabled:
+            return True, 'Channel {} output is currently enabled.'.format(params['channel'])
+        else:
+            return True, 'Channel {} output is currently disabled.'.format(params['channel'])
```
BrianJKoopman, line 247: This is alright, but only really shows up in the logs. Can you also add the information to the `session.data` like you did above? That allows control programs to retrieve the response, as it's available to the `OCSClient` object.

socs/agents/scpi_psu/agent.py
```diff
@@ -80,8 +107,13 @@ def monitor_output(self, session, params=None):
...
                     for chan in params['channels']:
-                        data['data']["Voltage_{}".format(chan)] = self.psu.get_volt(chan)
-                        data['data']["Current_{}".format(chan)] = self.psu.get_curr(chan)
+                        if self.psu.get_output(chan):
+                            data['data']["Voltage_{}".format(chan)] = self.psu.get_volt(chan)
+                            data['data']["Current_{}".format(chan)] = self.psu.get_curr(chan)
+                        else:
+                            self.log.warn("Cannot measure output when output is disabled")
+                            self.monitor = False
+                            return False, "Cannot measure output when output is disabled"
```
BrianJKoopman, line 116: What's the response from the PSU when you query a disabled channel?

socs/agents/scpi_psu/agent.py
```diff
@@ -5,18 +5,21 @@
 from ocs import ocs_agent, site_config
 from ocs.ocs_twisted import TimeoutLock
 
-from socs.agents.scpi_psu.drivers import PsuInterface
+from socs.agents.scpi_psu.drivers import PsuInterface, ScpiPsuInterface
 
 
 class ScpiPsuAgent:
-    def __init__(self, agent, ip_address, gpib_slot):
+    def __init__(self, agent, ip_address, gpib_slot, **kwargs):
```
BrianJKoopman, line 12: Sort of along the same lines as the below comment about the `port` arg, there's the question of how to handle the `gpib_slot` arg. Since one or the other is required, not both. Perhaps we can also make `gpib_slot` optional, but throw an error if neither is submitted. This avoids having an agent instance that's going to use the new Ethernet only interface required to have a `--gpib-slot` argument to function.

BrianJKoopman (changes requested):
Thanks for the PR @CAWNoodle! See my inline comments below, and let me know if you have any questions/comments.

Out of curiosity, which model devices do you have access to?

CAWNoodle:
> Thanks for the PR @CAWNoodle! See my inline comments below, and let me know if you have any questions/comments.
> 
> Out of curiosity, which model devices do you have access to?

Hi Brian, Thanks for your work on this! I'll get right to work making these fixes. We have a Keithley 2280S-60-3 and a 2230G-30-1, but no GPIB adapter. I might end up making an agent to work with the 2230G over USB in the future.

requirements.txt
```diff
@@ -60,3 +60,17 @@ pyModbusTCP
...
+# Ensure compatability with Setuptools 71
+packaging>=24
+ordered-set>=3.1.1
+more_itertools>=8.8
+jaraco.text>=3.7
+importlib_resources>=5.10.2
+importlib_metadata>=6
+tomli>=2.0.1
+wheel>=0.43.0
+platformdirs>=2.6.2
```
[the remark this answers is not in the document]
CAWNoodle: These have been removed, but I left in pre-commit.

socs/agents/scpi_psu/drivers.py
```diff
@@ -1,6 +1,150 @@
...
 
+# append new model strings as needed
+ONE_CHANNEL_MODELS = ['2280S-60-3', '2280S-32-6']
+THREE_CHANNEL_MODELS = ['2230G-30-1']
+
+
+class ScpiPsuInterface:
+    def __init__(self, ip_address, gpibAddr, port, **kwargs):
+        self.ip_address = ip_address
+        self.gpibAddr = gpibAddr
```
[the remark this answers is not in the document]
CAWNoodle: gpibAddr has been removed. It was left over from when I initially wanted to keep the driver as one class, but ultimately it made more sense to make separate classes.

socs/agents/scpi_psu/drivers.py
```diff
@@ -1,6 +1,150 @@
...
+        if (self.numChannels != 1):
+            self.set_chan(ch)
+            self.write('OUTP:ENAB ON')
+
+    def disable(self, ch):
+        '''
+        disabled output from a channel (1,2,3). once called, enable must be
+        called to turn on the channel again
+        '''
+        self.write('OUTP:ENAB OFF')
```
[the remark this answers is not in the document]
CAWNoodle: Yes, at least the 2280S has a output enable/disable on the front panel that can also be controlled over SCPI. The documentation suggests that the single-channel device presumes the "CHAN" parameter is always "1".

Interestingly, it seems this function is never called for either of the drivers, nor is it referenced in the agent code, probably vestigial at this point. Should I remove it altogether (from both drivers)?

socs/agents/scpi_psu/drivers.py
```diff
@@ -1,6 +1,150 @@
...
+            return idn_response
+
+    def configure(self):
+        self.model = self.read_model()
+        if (self.model in ONE_CHANNEL_MODELS):
+            self.numChannels = 1
+        if (self.model in THREE_CHANNEL_MODELS):
+            self.numChannels = 3
+        if (self.numChannels == 0):
+            raise ValueError('Model number not found in known device models', self.model)
```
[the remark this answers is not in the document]
CAWNoodle: agreed, the reason I eventually made a whole separate class for this was honestly just laziness; I didn't want to deal with the PrologixInterface inheritance. In the agent code I choose which class to use based on the presence of the 'port' argument; if this driver class is chosen the ValueError is caught and the error text is logged.

It would probably be better to query the device for number of channels rather than maintaining a hard-coded list of model numbers. For now, should I just change the error message to a warning and set numChannels to 3 as the default?

socs/agents/scpi_psu/drivers.py
```diff
@@ -1,6 +1,150 @@
...
+                self.write('CHAN:OUTP ON')
+            else:
+                self.write('CHAN:OUTP OFF')
+        else:
+            if isinstance(out, str):
+                self.write('OUTP ' + out)
+            elif out:
+                self.write('OUTP ON')
+            else:
+                self.write('OUTP OFF')
```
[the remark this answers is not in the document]
CAWNoodle: I never actually tried to send the `CHAN:OUTP ON` command to the 2280S. It does seem like different SCPI devices support different sets of commands, and that seems to be intentional from the [SCPI Protocol](https://www.ivifoundation.org/downloads/SCPI/scpi-99.pdf)

socs/agents/scpi_psu/agent.py
```diff
@@ -188,15 +292,23 @@ def main(args=None):
...
-    p = ScpiPsuAgent(agent, args.ip_address, int(args.gpib_slot))
+    p = ScpiPsuAgent(agent, args.ip_address, int(args.gpib_slot), port=int(args.port))
 
-    agent.register_task('init', p.init)
+    agent.register_task('init', p.init, startup=init_params)
     agent.register_task('set_voltage', p.set_voltage)
     agent.register_task('set_current', p.set_current)
     agent.register_task('set_output', p.set_output)
 
+    # Need tasks for get_voltage, get_current, and get_output
```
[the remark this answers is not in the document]
CAWNoodle: removed

socs/agents/scpi_psu/agent.py
```diff
@@ -178,6 +279,9 @@ def make_parser(parser=None):
     pgroup = parser.add_argument_group('Agent Options')
     pgroup.add_argument('--ip-address')
     pgroup.add_argument('--gpib-slot')
+    pgroup.add_argument('--port')
```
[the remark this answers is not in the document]
CAWNoodle: Hmm.. This is actually a problem for backwards compatibility. If the `--port` option isn't set, then casting a NoneType object to an Int later in the ScpiPsuAgent instantiation (line 300) will be interpreted as a `self.port = 0` and the check on line 47 `if self.port is None:` will always be `False`.

I think I should set the default to `0` and then change the line 47 check to `if self.port == 0:` Then any older code will automatically use the older driver and anybody wanting to use the new driver just has to specify the correct port in their site configs.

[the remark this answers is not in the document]
CAWNoodle: or to better address the **kwargs comment, I could leave this with no default, and only cast to Int when assigning it: 
```
    self.port = None
    if port is not None:
        self.port = int(port)
```

socs/agents/scpi_psu/agent.py
```diff
@@ -5,18 +5,21 @@
...
         self.agent = agent
         self.log = agent.log
         self.lock = TimeoutLock()
 
         self.job = None
         self.ip_address = ip_address
         self.gpib_slot = gpib_slot
+        self.port = None
+        if ('port' in kwargs.keys()):
+            self.port = kwargs['port']
```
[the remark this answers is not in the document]
CAWNoodle: Ok, implemented this, with `__init__(self, agent, ip_address, gpib_slot, port=None)` and moved the int cast to the assignment statement. Still no default set in the parser, and kept line 47 as `if self.port is None:`

How's that?

socs/agents/scpi_psu/agent.py
```diff
@@ -30,7 +33,7 @@ def __init__(self, agent, ip_address, gpib_slot):
                                  agg_params=agg_params,
                                  buffer_time=0)
 
-    @ocs_agent.param('_')
+    # @ocs_agent.param('_')
```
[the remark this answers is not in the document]
CAWNoodle: Done

socs/agents/scpi_psu/agent.py
CAWNoodle: Fixed

socs/agents/scpi_psu/drivers.py
CAWNoodle: haha no worries. a simple `:%s/numChannels/num_channels/g` did the trick

socs/agents/scpi_psu/agent.py
CAWNoodle: In order to retain the channel information in the output, could I do something like
```
data = {
    'timestamp': time.time()
}
data['Voltage_{}'.format(chan)] = self.psu.get_volt(chan)
session.data = data
```

socs/agents/scpi_psu/agent.py
CAWNoodle: fixed

socs/agents/scpi_psu/agent.py
CAWNoodle: I added it to the session data and made the appropriate changes to the docstring.

socs/agents/scpi_psu/agent.py
CAWNoodle: at least with the 2280S, it sets an internal error flag and doesn't otherwise respond. So when you subsequently call `socket.recv()` to get the response it results in a timeout error -- less than an ideal way of handling that case, imo.

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
CAWNoodle: Yes, this was my original intention, but I ended up taking a shortcut instead of properly handling the PrologixInterface inheritance. It probably wouldn't take much at this point, I just wanted to reduce potential sources of conflict while testing.

CAWNoodle:
Oh also. Uh, how do I actually submit these changes once they're made?

BrianJKoopman:
> Oh also. Uh, how do I actually submit these changes once they're made?

When you're ready you can push your new commits to this same branch. Then once the changes are ready for re-review click the little circular arrows next to my name in the reviewers list to the right to "re-request review". That lets me know it's time to take a look/no more changes are expected.

PS - Thanks for the quick turnaround, will try to re-review soon!

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
BrianJKoopman: Having extra functionality in the drivers that the agent doesn't take advantage of (yet) is fine. Sometimes users end up commanding things directly via the drivers for debugging, so I wouldn't remove it. But I'd drop the check on `self.numChannels != 1` in the `enable()` method, since that'll prevent it from enabling single channel devices if they're ever disabled.

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
BrianJKoopman: Yeah, that sounds good to me. And so does querying for number of channels.

socs/agents/scpi_psu/agent.py
BrianJKoopman: I think I'd rather see it as a separate key/value pair, i.e.:
```
data = {
    'timestamp': time.time(),
    'channel' = chan,
    'voltage' = self.psu.get_volt(chan),
}

session.data = data
```

But yeah, still fine to include if you'd like.

socs/agents/scpi_psu/agent.py
BrianJKoopman: Ah, alright, then we should add some error handling for that timeout error, else a query on a disabled channel will crash this process.

socs/agents/scpi_psu/agent.py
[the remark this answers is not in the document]
BrianJKoopman: Ah, right, the existence of a valid port number is what changes the path for whether we use the GPIB interface or not. Yeah, that makes sense w/o a default in the parser.

BrianJKoopman:
Responded to inline comments, let me know if you need anything else from me, else I'll wait for your push and re-review request to take another look. No rush, just checking I'm not blocking on this.

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
CAWNoodle: Ok, changed the `enable()` code to
```
if (self.num_channels != 1):
    self.set_chan(ch)
self.write('OUTP:ENAB ON')
```
which is probably what I was trying to do originally.

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
CAWNoodle: I did a dive into querying for number of channels, and I can't see anyway to do it with the universally implemented SCPI commands. Each device and manufacturer implements different commands. It seems like the only way to query for number of channels would be to iterate through a set of functions, one for each device, and then check the error flags. Of course this would be impractical to implement for all possible SCPI power supplies that might be in use.

It's unfortunate that the set of required commands for SCPI do not include some way of checking for number of channels. I've retained the lists of known model numbers for now, changed the logged error into a logged warning and set num_channels to 3 if the model number is not found.

socs/agents/scpi_psu/agent.py
CAWNoodle: Ok, changed this for both the `get_voltage()` and `get_current()` and I updated the docstrings to reflect the new output.

socs/agents/scpi_psu/agent.py
CAWNoodle: done

socs/agents/scpi_psu/agent.py
CAWNoodle: As written, this will check to make sure the output is enabled before attempting a measurement in the `monitor_output()` process. I will add a similar check in the `get_voltage()` and `get_current()` functions. 

For the 2280S, anytime a command results in an error the next `socket.recv()` call will result in a timeout error. Ideally, the agent code should never send anything that would cause such an error (i.e. the agent should never send malformed commands or out-of-bounds parameter values, etc.). If a timeout error does occur, then it should indicate a networking issue rather than a bad command.

That said, I think the only time anyone should run into a timeout error caused by a bad command (e.g. a measurement command when the output is off, or setting the voltage on a channel that doesn't exist) should be if they are calling the driver functions directly. I also don't know if other SCPI power supplies respond the same way to bad commands.

CAWNoodle:
I think I've addressed all the changes. Our lab is running a set of experiments at the moment, but I should be able to test these changes over the weekend. 

I started reading manuals for various SCPI power supplies and I've come to the conclusion that supporting all of them with a general purpose SCPI driver is not feasible. Each manufacturer implements different commands to perform the same operations and they are usually not cross-compatible. For example, the 2230G 3-channel devices use the `:INSTrument:SELect <CH1|CH2|CH3|>` command to select a channel for subsequent commands. Whereas the 2280S devices do not even have a `:INST:` root command. And the BK Precision 9173 & 9174 dual-channel devices don't have a channel selection command at all. Instead they have separate commands for each channel such as `:MEAS:CURR?` for channel one and `:MEAS:CURR2?` for channel two. Even the command for enabling/disabling the output differs from model to model.

A generalized power supply agent could be constructed that would implement all the things you typically want to do with a power supply, but under the hood we would need a sort of implementation file for each model and/or manufacturer.

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
BrianJKoopman: That is unfortunate, and something I don't think we realized when we renamed this agent from "Keithley" to "SCPI PSU" agent way back when. Thanks for digging into the details though! Keithley PSUs are the main use case here, with one or two BK Precision PSUs I believe also being used. Though I don't have a good idea of how regularly this agent is used these days. That said, I wouldn't recommend spending more time generalizing this unless the functionality is in demand.

socs/agents/scpi_psu/agent.py
BrianJKoopman: That makes sense. My concern is with the possibility of network issues. 

We don't handle network errors well in many of the agents, but are [trying to fix this](https://github.com/simonsobs/socs/issues/721). There is a simultaneous effort to make this agent robust to network errors too: https://github.com/simonsobs/socs/pull/726

See this discussion for more context: https://github.com/simonsobs/socs/discussions/538

Essentially we'd like to keep the agent online and have it reconnect to the device. In its current form the `monitor` process would crash if there's a Timeout. You'd then have to run `init` and then restart `monitor` after resolving the networking issue.

socs/agents/scpi_psu/agent.py
BrianJKoopman: Similar to the `session.data` structure for `get_current` and `get_voltage`, can we separate the channel from the enabled state key? Something like:
```
>>> response.session['data']
{'timestamp': 1723671503.4899583,
 'channel': 1,
 'state': 'enabled'}
```

or
```
>>> response.session['data']
{'timestamp': 1723671503.4899583,
 'channel': 1,
 'enabled': True}
```

Both for consistency with those other tasks and for ease of use on the client end. (It's simpler to check `resp.session['data']['state']` for any given channel rather than needing to build the `'channel_{number}'` string to get the key right.)

socs/agents/scpi_psu/drivers.py
```diff
@@ -1,6 +1,161 @@
...
+        self.port = port
+        self.sock = None
+        self.model = None
+        self.num_channels = 0
+        self.conn_socket()
+        try:
+            self.configure()
+        except ValueError as err:
+            raise ValueError(err)
+        super().__init__(**kwargs)
```
BrianJKoopman, line 33: This `super().__init__(**kwargs)` line can be safely removed, since `ScpiPsuInterface` doesn't inherit from a parent class. (This was probably copied from the [`PrologixInterface` class](https://github.com/simonsobs/socs/blob/99854a71a54347801eb1e8758219218856595458/socs/common/prologix_interface.py#L12) where it also isn't needed...)

socs/agents/scpi_psu/drivers.py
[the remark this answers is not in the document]
BrianJKoopman: Alright, maybe we leave the class inheritance for later then?

socs/agents/scpi_psu/agent.py
```diff
@@ -188,15 +351,22 @@ def main(args=None):
                                   parser=parser,
                                   args=args)
 
+    init_params = False
+    if args.mode == 'acq':
+        init_params = {'auto_acquire': True}
     agent, runner = ocs_agent.init_site_agent(args)
 
-    p = ScpiPsuAgent(agent, args.ip_address, int(args.gpib_slot))
+    p = ScpiPsuAgent(agent, args.ip_address, int(args.gpib_slot), port=int(args.port))
```
BrianJKoopman, line 359: This'll hit a `TypeError` when either `--gpib-slot` or `--port` are not passed:
```
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType
```

This is what the tests are failing on, as they don't supply `--port`, only `--gpib-port`.

You could move the cast into the agent's `__init__` after the check for `None`.

BrianJKoopman (changes requested):
Thanks for pushing the updates! I responded to the previous review's comments and made a couple additional small comments in this re-review.

It is too bad that the commanding isn't more uniform across manufacturers. Like I said in one of the comment responses, I believe most users have Keithley devices (this used to be named after one of those devices) -- I wouldn't recommend spending the time to generalize this further unless the demand is there for other PSUs.

BrianJKoopman:
Hi @CAWNoodle, just wanted to check in on this. #726, which introduces re-connection logic for when there's a network interruption. That should handle my concern I commented on in https://github.com/simonsobs/socs/pull/715#discussion_r1734925589. Can you merge in the latest `main`? There's a small merge conflict to be handled.

Other than that the biggest thing to fix is the `TypeError` from my recent re-review: https://github.com/simonsobs/socs/pull/715#discussion_r1734975380

Then this should be good to merge. Let me know if there's anything I can do to help.

CAWNoodle:
Sorry for the delay, The Fall semester has just begun and I have been trying to adjust to the new schedule. I should have time to get this done during this week.

BrianJKoopman:
> Sorry for the delay, The Fall semester has just begun and I have been trying to adjust to the new schedule. I should have time to get this done during this week.

Great, no rush on my end, mostly wanted to let you know about the #726 merge.

CAWNoodle:
So it looks like the changes from #726 are catching the timeout error from trying to measure when the output is off. If there is a networking issue it would only be caught when running `monitor_output`. clearing the error queue also indicates that this is primarily addressing timeout errors that result from the PSU's command execution rather than networking problems. 

Concurring with suggestions in #538, I think the best place to handle connectivity issues would be in the drivers. Likewise, the drivers should ensure that commands sent to the devices do not trigger internal errors, that way agent code doesn't have to be aware of any implementation details.

I've essentially left the #726 changes alone in the merge as they're not doing any harm. I've made they requested edits as well. I'm just going to run some tests to make sure everything is still working as expected and I'll push everything.

CAWNoodle:
Ok, I've pushed my changes. I had to comment out the call to `_initialize_module()` at the bottom of the `init()` function. It kept resulting in a timeout error, and I suspect it is because the PSU already had a connection established on the same port. So trying to connect a second time resulted in a timeout.

socs/agents/scpi_psu/agent.py
CAWNoodle: Done

socs/agents/scpi_psu/agent.py
CAWNoodle: fixed

socs/agents/scpi_psu/agent.py
CAWNoodle: This should be fixed now

socs/agents/scpi_psu/drivers.py
CAWNoodle: Almost forgot this one, but it's fixed now

BrianJKoopman (commented):
> Ok, I've pushed my changes. I had to comment out the call to `_initialize_module()` at the bottom of the `init()` function. It kept resulting in a timeout error, and I suspect it is because the PSU already had a connection established on the same port. So trying to connect a second time resulted in a timeout.

Thanks, this is looking good. I have a [few commits](https://github.com/TexMExLab/socs/compare/UTAustin...simonsobs:socs:koopman/UTAustin) I'd like to push before merging that fix the problems in the tests. On your PR can you check ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)? (I could also PR onto your branch, though this is a bit more straightforward if possible.)

CAWNoodle:
Hmm... I do not appear to have that checkbox shown in ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). Any ideas on how I can fix that?

CAWNoodle:
> > Hmm... I do not appear to have that checkbox shown in ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). Any ideas on how I can fix that?
> 
> No clue. I just opened [TexMExLab#3](https://github.com/TexMExLab/socs/pull/3). If you merge that they should show up here.

Done

BrianJKoopman (approved):
Great, thanks for all the work on this! I'm checking with @agthomas-uc to see if he can test this on the hardware with the GPIB converter to double check old behavior is unaffected. Assuming that'll work out I'm going to wait on that test to merge. Thanks again!

agthomas-uc:
I verified the new agent works on our hardware with a GPIB converter. I performed three reconnect tests of variable length, making sure the longest was at least a minute to ensure secondary timeout errors did not occur. Grafana view and corresponding log showing flawless performance are attached. 
image
image

PR #720, opened 2024-08-12
HWP supervisor agent brake bugfix

bbixler500:
This PR switches a couple lines in the brake functionality of the hwp-supervisor agent.

## Description

In the logic of the brake function I switched the order of lines so that the setting of the voltage occurs after switching the pmx to direct voltage control.

## Motivation and Context

Currently the brake function of the hwp-supervisor has an argument `brake_voltage` which does not work because of a bug. To brake, the hwp uses a PMX power supply. This power supply can be configured to use an external voltage to control the output. This configuration is enabled during rotation and takes a PID feedback as an input. To brake however, this configuration is disabled and a static voltage is used instead. The issue is that while the external voltage configuration is enabled, commands to change the static voltage are ignored. Thus as the code was previously, commands to change the brake voltage would be ignored, and the PMX would simply use the last commanded static voltage value as the brake voltage (for satp1 this was 31V and for satp3 this was 10V).

## How Has This Been Tested?

I tested these changes with the satp2 hwp. After implementing the changes the brake issue was resolved

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -957,10 +957,10 @@ class ControlStateMachine:
                 self.run_and_validate(clients.pid.set_direction,
                                       kwargs=dict(direction=new_d))
                 self.run_and_validate(clients.pid.tune_stop)
-                self.run_and_validate(clients.pmx.set_on)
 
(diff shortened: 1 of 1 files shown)


ykyohei (approved):
Looks good to me!

jlashner (commented):
Thanks for the fix, this looks good to me!

PR #722, opened 2024-08-13
HWP supervisor iboot indexing bugfix

bbixler500:
Changes how the supervisor indexes iboot outlets when querying their state. This change only effects how the outlet states are reported in the `monitor` session data (actual functions using the iboot outlets are uneffected)

## Description

Subtracts one from the supplied outlet number, as the iboot outlets are indexed 0-7 while the outlet input argument is indexed 1-8.

## Motivation and Context

This fix ensures that the session data is accurate. Additionally, the supervisor agent would previously crash with an index error whenever trying to use outlet 8 (which is the case for the satp2 gripper power supplies)

## How Has This Been Tested?

I tested this fix with the satp2 hwp. 
Outlet states:
iboot outlet state

Before change:
supervisor old outlet index

After change:
supervisor new outlet index

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -133,7 +133,7 @@ class IBootState:
             return
 
         if self.agent_type == 'iboot':
-            self.outlet_labels = {o: f'outletStatus_{o}' for o in self.outlets}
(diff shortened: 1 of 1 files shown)


jlashner (approved):
Looks good to me, thanks bryce!

Issue #725, opened 2024-08-15
Make ScpiPsuAgent robust to connection dropouts


agthomas-uc:
As per Brian's request in [socs/issues/721](https://github.com/simonsobs/socs/issues/721), I am making this issue to track keep track of the fact I am working on this problem. I believe I understand the general structure of how to solve this issue and would be willing to modify and test additional agents afterwards.

BrianJKoopman:
Resolved by #726.

Issue #721, opened 2024-08-13
Make Agents robust to connection dropouts


BrianJKoopman:
Related to discussion https://github.com/simonsobs/socs/discussions/538, we need to make sure agents are robust against connection dropouts, whether that's a network interruption, serial connection dropout, or otherwise.

I'm making a single issue for this to avoid spamming 50+ individual issues. If you would like to contribute, read on! 

## Contributing
If you would like to work on "robustifying" an agent, please create an issue with a title like "Make \<AGENT NAME\> robust to connection dropouts" and assign yourself. This will help us keep track of which agents are actively being worked on, and which are up for grabs.

When you're ready, PR your code changes and [link the associated issue](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests). Once merged I'll update the checklist below.

(If you know any of the agents to already be robust, please comment here.)

## Robust Agents Checklist
* [ ] ACTiCameraAgent
* [ ] ACUAgent
* [ ] BlueforsAgent
* [ ] CrateAgent
* [ ] CryomechCPAAgent
* [ ] FPGAAgent
* [ ] FlowmeterAgent
* [ ] FTSAerotechAgent
* [ ] GeneratorAgent
* [ ] Hi6200Agent
* [ ] HWPBBBAgent
* [ ] HWPGripperAgent
* [ ] HWPPCUAgent
* [ ] HWPPicoscopeAgent
* [ ] HWPPIDAgent
* [ ] HWPPMXAgent
* [ ] HWPSupervisor
* [ ] ibootbarAgent
* [ ] LabJackAgent
* [ ] Lakeshore240Agent
* [ ] Lakeshore336Agent
* [ ] Lakeshore370Agent
* [ ] Lakeshore372Agent
* [ ] Lakeshore425Agent
* [ ] LATRtXYStageAgent
* [ ] MagpieAgent
* [ ] MeinbergM1000Agent
* [ ] MeinbergSyncboxAgent
* [ ] PfeifferAgent
* [ ] PfeifferTC400Agent
* [ ] PysmurfController
* [ ] PysmurfMonitor
* [ ] ScpiPsuAgent
* [ ] SmurfFileEmulator
* [ ] SmurfStreamSimulator
* [ ] SmurfTimingCardAgent
* [ ] SupRsync
* [ ] SynaccessAgent
* [ ] SynthAgent
* [ ] TektronixAWGAgent
[...]

PR #726, opened 2024-08-15
Scpi psu reconnect

agthomas-uc:
I have implemented a robust reconnecting agent to re-establish the `PsuInterface`, clear the PSU's event register, and begin reading out data. Let me know if I can help work with Daniel to get these PRs synced up. As my code is ignorant to a direct ethernet connection, there will be a compatibility issue around `self.port` which will require me to modify `_initialize_module `agent.py in the same way Daniel did to the main body of in his PR.

## Description
Implemented in the agent with one new driver function, these changes prevent the agent from crashing due to any disconnect after initialization and from block errors due disconnects during data readout. These changes attempt to reconnect the agent on a 5 second cadence and log each attempt and whether it was successful. Upon reconnecting, the event register is cleared to ensure previously `MEAS?` commands are not outputted to incorrect channels.

Long form description (copied from [simonsobs/daq-discussions#106](https://github.com/simonsobs/daq-discussions/discussions/106)).
1. As Brian suspected, implementing a reliable reconnection required clearing the PSU. To do so I created a clear command in the driver which is called once the agent reconnects. This ensures no queries from when the PSU was disconnected are sent and no bad measurements are received. Attempting to query leftover data from a disconnect can return 2 or 3 values in one query separated by \n characters. 
- I initially resolved these bad measurements by adding `.splitlines()` and attempting to use this data. However, reading data with a newline character creates a driver bug where measuring ch1 voltage returns 0 (integer zero), measuring ch1 current returns ch1 voltage, measuring ch2 voltage returns ch1 current, ..., measuring ch3 current returns ch3 voltage, and ch3 current is unrecoverable. Repeating this process does not turn additional channels to zero and instances of two \n characters still only "cycled" the data by 1. Querying the power as well did not stop the cycle. I tried many iterations of driver code and read the 2230G user's manual but did not determine why this behavior would occur.
2. The block structure error would occur when a disconnect (`socket.timeout()`) caused the data acquisition for loop to break without reading all values to data. Since psu is set to None when a disconnect occurs, I prevent block structure errors by checking that psu is still initialized before publishing.
- I initially implemented a more heavy handed solution comparing the length of `data['data']` to twice the number of channels. However, I believe checking if psu is initialized accomplishes the same goal more efficiently. I can revert if I missed a case.
3. The initialize module sees two types of Exceptions. When the module fails to initialize because it is still disconnected, a `socket.timeout()` occurs after 5 seconds (as set in prologix_interface.py#L10). If the connection has not been established within one minute, OSError [Errno 113] "No route to host" occurs. I deal with both by excepting any Exception to enable reconnect to attempt indefinitely. 

## Motivation and Context

This change is required because our PSU disconnects on a regular basis, between hourly and weekly, and previously needed to be manually power cycled to be restored. This prevented us from using the PSU to heat the focal plane and has been a standing issue in our lab for over 2 years.

This addresses one agent listed here [socs/issues/721](https://github.com/simonsobs/socs/issues/721) and the issue I created to break this out [socs/issues/725](https://github.com/simonsobs/socs/issues/725).

## How Has This Been Tested?

I tested routinely during the creation of this branch and documented issues in [simonsobs/daq-discussions#69](https://github.com/simonsobs/daq-discussions/discussions/69) and ([simonsobs/daq-discussions#75](https://github.com/simonsobs/daq-discussions/discussions/75). After attempting to resolve these issues, I disconnected and reconnected the ethernet cable irregularly over a 20 minute period. 11/11 Reconnections were successful as shown below. 

On the final test I made sure to disconnect the agent during the reconnection process multiple times and disconnected and reconnected the ethernet as fast as possible for 5 seconds. As I figured, this seemed to severely confuse the device. However, it recovered after ~2.5 minutes.
image

I am now running a duration test to ensure the connection stays reliable for an extended period of time and have remained connected for 5 days. I will update this PR if any disconnects and automatic reconnects occur.

## Types of changes

- [x] New feature (non-breaking change which adds functionality), the driver command to CLR may carry additional functionality.

## Checklist:


socs/agents/scpi_psu/agent.py
```diff
@@ -68,28 +86,38 @@ def monitor_output(self, session, params=None):
...
+                    if self.psu:  # Check to make sure there weren't any timeout errors so we don't publish incomplete data and make block errors.
+                        # self.log.info(str(data))
+                        self.agent.publish_to_feed('psu_output', data)
+
+                        # Allow this process to be queried to return current data
+                        session.data = data
+
+                    else:
+                        if not self._initialize_module(session, params):
+                            time.sleep(5)  # wait 5 sec before trying to re-initialize
```
BrianJKoopman, line 120: This reconnect logic looks good and can be kept. Just to offer a suggestion for a less nested approach (since this is already a heavily nested block) I would have done something like this pseudo-code:

```
while monitor:
    if not psu:
        # reconnect to PSU
        # check connection, if not connected continue

    try:
        # take data
    except:
        psu = None
        continue

    # publish data
```

Essentially ensure the connection is established right off the bat, else continue to the next loop iteration where we try to re-establish connection. After which you can assume the connection is good when you try to take data, if that fails, continue to the 're-establish connection/take data' in the next loop, then publish data.

socs/agents/scpi_psu/agent.py
```diff
@@ -48,9 +53,22 @@ def init(self, session, params=None):
...
-
         return True, 'Initialized PSU.'
 
+    def _initialize_module(self, session, params=None):
+        """Initialize the ScpiPsu module."""
+        try:
+            self.psu = PsuInterface(self.ip_address, self.gpib_slot)  # this is only necessary if the ip_address or gpib_slot has changed. I could declare a self.connected flag if this is uneeded.
+            self.idn = self.psu.identify()
+        except Exception as e:  # socket.tiout OR OSError 113 No route to host.
+            self.log.info("Failed to reconnect, trying again")
```
BrianJKoopman, line 64: Two things here -- I'd change `self.log.info` to `self.log.warn`, and also log the error encountered, either in a separate line or like:
```
self.log.warn(f"Error establishing connection: {e}")
```

BrianJKoopman (changes requested):
This is great, thanks @agthomas-uc! And thanks for including the results of your thorough testing. A few suggestions below.

As far as coordinating with #715, currently that adds a separate class for the direct Ethernet connection, so I think a merge would be quite minimal, though we'd have to re-implement this reconnection logic in that class as well. I pointed out this PR in a recent re-review there, so maybe we can see what @CAWNoodle says?

socs/agents/scpi_psu/agent.py
agthomas-uc: I slightly modified the structure to the following. The if psu check before publishing is needed before publishing data to prevent block errors.
```
while monitor:
    if not acquired:
        # warn

    elif not psu:
        # reconnect to PSU
        # check connection, if not connected continue
    
    else:
        for chan in channels:
            try:
                # take data
            except:
                psu = None
                break
        
        if psu:
            # publish data
```

agthomas-uc: I found a better solution using `continue ` to reduce nesting. I moved the sleep command to the beginning since using continue means the rest of while loop is skipped and placing 4 identical sleep command for the same purpose does not seem ideal. Psuedo-code of the new structure is below for ease of understanding:
```
while monitor:
    # sleep
    if not acquired:
        # warn
        continue

    if not psu:
        # reconnect to PSU
        continue
    
    try:
        for chan in channels:
            # take data
    except:
        psu = None
        continue

    # publish data
```

BrianJKoopman: This looks great, thanks for the updates!

Issue #727, opened 2024-08-19
Error reading direction in HWP PID Agent


BrianJKoopman:
<div type='discussions-op-text'>

<sup>[Originally posted](https://github.com/simonsobs/chwp-discussions/discussions/26) by **@ykyohei** August 15, 2024</sup>
satp3 hwp-pid agent crashed by the error in reading direction.
This might have been introduced by the recent update of hwp-pid agent. https://github.com/simonsobs/socs/pull/686

Based on manual https://assets.omega.com/manuals/M3397.pdf
R24 indicates "miscellaneous".
```
2024-08-15T19:15:24+0000 main:0 CRASH: [Failure instance: Traceback: <class 'ValueError'>: Could not get direction
2024-08-15T19:15:24+0000 Error reading direction: Unrecognized Read
2024-08-15T19:15:24+0000 DecodedResponse(msg_type='error', msg='Unrecognized Read', measure=None)
2024-08-15T19:15:24+0000 ['R2400000']
2024-08-15T19:15:24+0000 Error reading direction: Unrecognized Read
2024-08-15T19:15:24+0000 DecodedResponse(msg_type='error', msg='Unrecognized Read', measure=None)
2024-08-15T19:15:24+0000 ['R2400000']
2024-08-15T19:15:24+0000 Error reading direction: Unrecognized Read
2024-08-15T19:15:24+0000 DecodedResponse(msg_type='error', msg='Unrecognized Read', measure=None)
2024-08-15T19:15:24+0000 ['R2400000']
2024-08-15T19:15:24+0000 Finding CHWP Direction
```
Usually we read
```
2024-08-15T19:20:19+0000 DecodedResponse(msg_type='read', msg='Direction = Forward', measure=0)
2024-08-15T19:20:19+0000 ['R02400000']
2024-08-15T19:20:18+0000 Finding CHWP Direction
```
</div>

ykyohei:
Similar error due to the drop of '0' is observed.
```
2024-08-22T23:00:13+0000 main:0 CRASH: [Failure instance: Traceback: <class 'ValueError'>: Could not get target frequency
2024-08-22T23:00:13+0000 Error reading target: Unrecognized Read
2024-08-22T23:00:13+0000 DecodedResponse(msg_type='error', msg='Unrecognized Read', measure=None)
2024-08-22T23:00:13+0000 ['R14007D0']
2024-08-22T23:00:13+0000 Error reading target: Unrecognized Read
[...]

PR #686, opened 2024-06-03
added handling of pid error response

bbixler500:
Attempts to fix hwp pid error handling issues

## Description

Occasionally the hwp pid controller will read an error value which it does not know how to handle, which will crash the monitoring process. This PR adds handling to this specific error code.

## Motivation and Context

Issue described here ([link](https://github.com/simonsobs/chwp-discussions/discussions/20))

## How Has This Been Tested?

Briefly ran changes on satp1 hardware. Error rarely occurs, so was unable to verify correct handling explicitly.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #728, opened 2024-08-19
hwp-pid direction read bugfix

bbixler500:
Initial attempt to fix occasional issues in reading the hwp-pid direction

## Description

Adds additional logic to handle read strings that are formatted in a non standard way.

## Motivation and Context

Recently changes were pushed to the hwp-pid agent #686 which changed how response messages were processed. As an unintended consequence, a new bug surfaced when reading the rotation direction #727. From these error logs it appears as though the response message is mostly correct, but it is missing an initial 0 digit. This PR is a simple fix which will remedy this specific issue. Feedback is appreciated if the issue is more complicated than described above.

## How Has This Been Tested?

This fix has not been tested and needs to be before it is merged into main

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_pid/drivers/pid_controller.py
+++ b/socs/agents/hwp_pid/drivers/pid_controller.py
@@ -524,7 +524,12 @@ class PID:
 
         """
         end_string = string.split('\r')[-1]
-        read_type = end_string[1:3]
+        if len(end_string) == 9:
+            read_type = end_string[1:3]
+        elif len(end_string) == 8:
+            read_type = '0' + end_string[1:2]
+        else:
+            return DecodedResponse(msg_type='error', msg='Unrecognized Read Length')
         # Decode target
         if read_type == '01':
             target = float(int(end_string[4:], 16) / 1000.)
@@ -536,7 +541,7 @@ class PID:
             else:
                 return DecodedResponse(msg_type='read', msg='Direction = Forward', measure=0)
         else:
-            return DecodedResponse(msg_type='error', msg='Unrecognized Read')
+            return DecodedResponse(msg_type='error', msg='Unrecognized Read Type')
 
     @staticmethod
     def _decode_write(string):
--- a/socs/agents/hwp_pid/drivers/pid_controller.py
+++ b/socs/agents/hwp_pid/drivers/pid_controller.py
@@ -257,9 +257,11 @@ class PID:
         if decoded_resp.msg_type == 'measure':
             return decoded_resp.measure
         elif decoded_resp.msg_type == 'error':
-            raise ValueError(f"Error reading freq: {decoded_resp.msg}")
+            print(f"Error reading freq: {decoded_resp.msg}")
+            raise ValueError
         else:
-            raise ValueError("Unknown freq response")
+            print("Unknown freq response")
+            raise ValueError
 
     @retry_multiple_times(loops=3)
     def get_target(self):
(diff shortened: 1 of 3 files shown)


ykyohei:
Yeah, I was not sure if 'R2400000' means miscellaneous or it actually means direction but zero is missing.
I'm fine with merging this and see if resolves this type of error.
But I request to test this for a while to make sure that this does not introduce other errors.

ykyohei:
Actually, I got IndexError from pid agent probably for the first time. 
I think this means that there is a problem in the current string decoding for edge cases..
https://github.com/simonsobs/chwp-discussions/discussions/28

jlashner (changes requested):
How sure are we that there is actually a missing 0 in the return string, and not that this is some other error?

Instead of assuming this is the case, I'd much rather add proper error handling to the string parsing such that unexpected responses do no crash the main process. I think string parsing like this must be wrapped in a try/catch, where if there is an error, we log the unexpected response and return an "error" response so that the main process can handle it properly.

bbixler500:
I've added a decorator to the read functions which will cause those functions to try multiple times before failing. This should hopefully fix some of the crashes this agent has been getting. I've also reverted some of the response handling changes that were previously implemented. The manner in which the read statements fail doesn't seem to be as consistent as I had believed, so those changes did not cover all failure cases. The most recent additions have not been tested yet, but I've pushed what they are currently in case anyone has comments.

socs/agents/hwp_pid/drivers/pid_controller.py
```diff
@@ -236,17 +251,17 @@ def get_freq(self):
...
-            elif decoded_resp.msg_type == 'error':
-                print(f"Error reading freq: {decoded_resp.msg}")
-        raise ValueError('Could not get current frequency')
+        if self.verb:
+            print(responses)
+            print(decoded_resp)
+        if decoded_resp.msg_type == 'measure':
+            return decoded_resp.measure
+        elif decoded_resp.msg_type == 'error':
+            raise ValueError(f"Error reading freq: {decoded_resp.msg}")
```
jlashner, line 260: This will be swallowed by the `try/catch` in the `retry_multiple_times` decorator, so it won't be logged to stdout. Make sure to log an error here with the problematic response.

jlashner (changes requested):
Hi Bryce, thanks for the changes! Looks good to me for the most part but there are a couple of important points in inline comments that I think need to be addressed.

bbixler500:
I've added handling of the `None` response to the agent and changed how the errors are logged. Let me know if there is any other issue or if this PR looks good to merge

jlashner:
I'm sorry I don't think I agree with this method of handling the readout errors. I think that if any of the parameters fail to read-out, we should throw an error in `get_pid_state` rather than returning a valid dict but with specific fields removed.

Agents calling this function expect to get a complete response, and if that is not possible then they need decide how to handle that case. Removing fields like you do here kind of just hides the problem, and moves the responsibility of checking whether fields exist to the caller, which they're not currently doing.

We'll also need to add error handling to the functions in this agent that call `get_pid_state`. I think the behavior we want is:
 - If `get_pid_state` fails in the main process, we should mark the process as "degraded", and continue onto the next iteration after a short sleep
 - If the function fails in the `get_state` task, then I think we want that task to fail.

bbixler500:
Okay, I've changed the handling of readout errors to follow Jack's suggestions. Read errors in the `main` process will print a warning message but will continue on to the next iteration, while read errors in the `get_state` will raise a `ValueError`

jlashner:
Ok cool, I think this is starting to look better. One quick thing and then I think this is good to merge... to set the process to be degraded, we want to run:
```
session.degraded = True
```
and then we want to set this back to `False` after it has recovered.

This  will make it so the operation state will appear as "degraded" in grafana, and we can set alarms based on that.

jlashner (approved):
Hey @bbixler500, looks like there is a test failing, but as soon as that's fixed and this is tested I think we can merge this.

bbixler500:
Okay, all tests are passing now

BrianJKoopman:
> Okay, all tests are passing now

Great, has the most recent version been tested against hardware?

bbixler500:
Yeah, every step of the way I was testing on satp2 before pushing my updates

PR #729, opened 2024-08-20
Added sample heater display mode check in custom_pid function.

### LINKED ISSUE #494: 372 custom pid -- sample heater setting
For the sample heater range setting, custom pid process should set the display value appropriately, rather than assume it's in current mode.

mjrand:
## Description

I added in a simple check for the display mode of the sample heater in the custom PID function.
If the sample heater is not in "current" mode, then the display mode will be changed to "current" mode.

## Motivation and Context

The custom PID function assumes that the sample heater display mode is set to current mode but does not check.
This leads to situations where the sample heater is in power mode and the custom PID will attempt to set the sample heater to a very high number in Watts instead of % of max current. This usually causes the PID function to break because the power in W is over the maximum range of the heater output. This confuses people when the custom PID will seemingly not work when started.

Resolves #494.

## How Has This Been Tested?

Simple function addition shouldn't cause issues.

## Types of changes

## Checklist:


BrianJKoopman (approved):
This looks good to me. I linked it to the relevant issue, #494. Also just merged in `main`, since the branch is 8 months old and docs weren't building properly because of it.

PR #731, opened 2024-08-23
HTTPCameraAgent Handle token lease and usage

davidvng:
## Description

Add logic to handle Reolink's API token. Grabs token if doesn't already have one or lease is close to expired.

## Motivation and Context

The agent grabbed a token in each loop before, reaching the limit of 20 max tokens per account.

## How Has This Been Tested?

Tested on `daq-dev`

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/http_camera/agent.py
+++ b/socs/agents/http_camera/agent.py
@@ -40,7 +40,7 @@ class HTTPCameraAgent:
         txaio logger object, created by the OCSAgent
     """
 
-    def __init__(self, agent, config_file):
+    def __init__(self, agent, config_file, renew_token=2700):
         self.agent = agent
         self.is_streaming = False
         self.log = self.agent.log
@@ -49,6 +49,8 @@ class HTTPCameraAgent:
         file_path = os.path.join(os.environ['OCS_CONFIG_DIR'], config_file)
         with open(file_path, 'r') as f:
             self.config = yaml.safe_load(f)
+
+        self.renew_token = renew_token
 
         agg_params = {
             'frame_length': 10 * 60  # [sec]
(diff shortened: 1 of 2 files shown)


socs/agents/http_camera/agent.py
```diff
@@ -200,6 +224,8 @@ def add_agent_args(parser=None):
     pgroup = parser.add_argument_group("Agent Options")
     pgroup.add_argument("--config-file", type=str, help="Config file path relative to OCS_CONFIG_DIR")
     pgroup.add_argument("--mode", choices=['acq', 'test'])
+    pgroup.add_argument("--renew-token", type=int, default=2700,
+                        help="Renew API token after this amount of seconds. Used for Reolink cameras.")
```
BrianJKoopman, line 228: It seems to me like this should go into the `cameras.yaml` config file.

It's unique to the Reolink cameras. You can still default to 2700, but that also lets you set it per camera if needed for whatever reason. Existing configs would still work since it'll just grab the default if missing.

Maybe there's another reason to have it global like this?

BrianJKoopman (commented):
Looks good to me, just one suggestion about input of the `--renew-token` duration. See below.

socs/agents/http_camera/agent.py
davidvng: Right, I decided to go this route to future-proof in case we add other brands that also require tokens, which I imagine wouldn't be uncommon.

BrianJKoopman: Alright, I don't think having it in the config file is any less future proof, but we can roll with this if you'd like. I'll go ahead and merge now.

BrianJKoopman (approved):

PR #738, opened 2024-08-27
HWP Encoder Update (Restart process)

bbixler500:
## Description

Adds a `restart` task to the agent, similarly to what is done for the hwp-gripper agent in #737. This task, when called, will remotely restart the process running separately on a beaglebone microcontroller.

## Motivation and Context

Currently, any time there is a power cycle to the hwp enclosure, a remote user needs to manually ssh into the encoder beaglebone microcontrollers and restart a process on each. This is both a hassle and requires an expert's knowledge. The restart task condenses this process into a single button on OCS-Web which can be run by anyone.

## How Has This Been Tested?

This task has been tested to work on the satp2 hwp. The encoder processing portion of the code is completely unaffected.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_encoder/agent.py
+++ b/socs/agents/hwp_encoder/agent.py
@@ -726,10 +726,10 @@ def make_parser(parser=None):
     # Add options specific to this agent.
     pgroup = parser.add_argument_group('Agent Options')
     pgroup.add_argument('--port', type=int, default=8080,
-                        help='IP of bbb running the corresponding encoder process')
-    pgroup.add_argument('--ip', type=str, default='None',
                         help='Listening port of Agent for receiving UDP encoder packets. '
                              'This should match what is defined in the bbb encoder process configs')
+    pgroup.add_argument('--ip', type=str, default='None',
+                        help='IP of bbb running the corresponding encoder process')
 
     return parser
 
--- a/socs/agents/hwp_encoder/agent.py
+++ b/socs/agents/hwp_encoder/agent.py
@@ -488,12 +488,13 @@ class HWPBBBAgent:
 
     """
 
-    def __init__(self, agent_obj, port=8080):
+    def __init__(self, agent_obj, port=8080, ip='None'):
         self.active = True
         self.agent = agent_obj
         self.log = agent_obj.log
         self.lock = TimeoutLock()
         self.port = port
+        self.ip = ip
         self.take_data = False
         self.initialized = False
         # For clock count to time conversion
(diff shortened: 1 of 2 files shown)


ykyohei (commented):
This is great, thank you very much!
I have one request, if there is a setup requirement for restart, I would like to have a brief document for it in the README of sobonelib or other appropriate places. What's the setup method? Do we just need to use latest sobonelib? I assume that the sobonelib have to be placed in specific path or something, is it correct?

bbixler500:
Agreed that I need to add additional information to the README of sobonelib of how these changes effect things. To give a brief summary, the processes have changed from a script which is running in some terminal session into a service located in the /etc/systemd/system/ folder (similar to how the hostmanager functions). These services are setup to restart automatically whenever there is a powercycle/crash and everything required for the setup is included into the Makefile (so it doesn't matter where the sobonelib directory is located). To implement the change all you need to do is run the command `make` in the `hwp_encoder` directory. Once that's done, the process will be running in the background without the need for a terminal session. For debugging purposes the old method will still work, so long as you disable the service first.

ykyohei (approved):
I confirmed there is no packet drops after start using this.

socs/agents/hwp_encoder/agent.py
```diff
@@ -695,6 +726,7 @@ def make_parser(parser=None):
     # Add options specific to this agent.
     pgroup = parser.add_argument_group('Agent Options')
     pgroup.add_argument('--port', type=int, default=8080)
+    pgroup.add_argument('--ip', type=str, default='None')
```
BrianJKoopman, line 729: At minimum we need some description here. With the only options being `--port` and `--ip` one could easily assume this is the IP and port on the beaglebone, when really it's a listening port in the Agent for receiving UDP packets, and the IP of the beaglebone.

socs/agents/hwp_encoder/agent.py
```diff
@@ -511,6 +512,36 @@ def __init__(self, agent_obj, port=8080):
...
+        _restart_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        _restart_socket.connect((self.ip, 5656))
+        _restart_socket.sendall(('reset\n').encode())
+        time.sleep(0.5)
+
+        resp = _restart_socket.recv(4096).decode().strip()
+        log = f'Restart command response: {resp}'
+        result = True if resp == "Success" else False
+        _restart_socket.close()
+        time.sleep(10)
```
BrianJKoopman, line 540: Is this sleep just to give the beaglebone time to restart?

BrianJKoopman (commented):
Just a couple of comments/questions from me.

socs/agents/hwp_encoder/agent.py
bbixler500: Yeah. I'm not sure how long the restart process on the beaglebone takes (It's likely only a couple seconds) so I just put a conservative sleep time just to be sure that the agent doesn't try to restart when a restart is already running

socs/agents/hwp_encoder/agent.py
```diff
@@ -720,6 +720,12 @@ def _stop_acq(self, session, params=None):
 
 
 def make_parser(parser=None):
+    """
+    Arguments:
+        port (int): Listening port of Agent for receiving UDP encoder packets. This 
+                    should match what is defined in the bbb encoder process configs
+        ip (str): IP of bbb running the corresponding encoder process
+    """
```
BrianJKoopman: Thanks for adding these. Can you move them to the `description` argument within the `pgroup.add_argument()` calls below. See [argparse docs](https://docs.python.org/3/library/argparse.html#description) for an example.

socs/agents/hwp_encoder/agent.py
```diff
@@ -694,7 +725,11 @@ def make_parser(parser=None):
 
     # Add options specific to this agent.
     pgroup = parser.add_argument_group('Agent Options')
-    pgroup.add_argument('--port', type=int, default=8080)
+    pgroup.add_argument('--port', type=int, default=8080,
+                        help='IP of bbb running the corresponding encoder process')
+    pgroup.add_argument('--ip', type=str, default='None',
+                        help='Listening port of Agent for receiving UDP encoder packets. '
+                             'This should match what is defined in the bbb encoder process configs')
```
BrianJKoopman, line 732: Thanks for moving these, but they got flipped in the move.

BrianJKoopman (approved):
Thanks, looks good to me!

@jlashner, did you want to take a look? Else we can squash and merge.

jlashner (approved):
Looks good to me! Excited to have this functionality.

PR #740, opened 2024-08-29
HTTPCameraAgent write file change

davidvng:
## Description

- Move the write to file into a `try-except` block and write in chunks
- Slight change to setting `connected` for each camera

## Motivation and Context

To avoid `ReadTimeoutError` during writing to file, since the response data is streamed

## How Has This Been Tested?

Tested on `daq-dev`

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/http_camera/agent.py
+++ b/socs/agents/http_camera/agent.py
@@ -173,7 +173,7 @@ class HTTPCameraAgent:
                                 out_file.write(chunk)
                         out_file.flush()
                         os.fsync(out_file.fileno())
-                        self.log.debug(f"Wrote {ctime}.jpg to /{camera['location']}/{ctime_dir}.")
+                    self.log.debug(f"Wrote {ctime}.jpg to /{camera['location']}/{ctime_dir}.")
                 except BaseException as e:
                     self.log.error(f'{e}')
                     self.log.info("An error occurred while writing to file.")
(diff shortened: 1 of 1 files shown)


socs/agents/http_camera/agent.py
```diff
@@ -159,29 +157,36 @@ def acq(self, session, params=None):
...
-                self.log.debug(f"Wrote {ctime}.jpg to /{camera['location']}/{ctime_dir}.")
+                try:
+                    with open(filename, 'wb') as out_file:
+                        for chunk in response.iter_content(chunk_size=8192):
+                            if chunk:
+                                out_file.write(chunk)
+                        out_file.flush()
+                        os.fsync(out_file.fileno())
+                    self.log.debug(f"Wrote {ctime}.jpg to /{camera['location']}/{ctime_dir}.")
+                except BaseException as e:
```
BrianJKoopman, line 177: Catching `BaseException` like this isn't ideal. That catches exceptions that indicate the program should exit like `SystemExit` and `KeyboardInterrupt`, which if raised at the right time could make the agent difficult to stop.

With this code block:
```
shutil.copyfileobj(response.raw, out_file)
# Ensure all data is written to the disk before copying to latest
out_file.flush()
os.fsync(out_file.fileno())
```

We know the error we were running into is another `urllib3.exceptions.ReadTimeoutError` on the `shutil.copyfileobj` line. I don't know that you're going to see that with the new chunk iteration code.

Good practice is to catch and handle specific errors, so my preference here is to `except ReadTimeoutError as e`, like you do above, with the log indicating a timeout occurred. If needed, catching `Exception` is better than the broader `BaseException`, as that doesn't catch those exit exceptions.

BrianJKoopman (changes requested):
Looks good, though we should be more specific in the error handling.

PR #743, opened 2024-08-30
Added Agent for srs cg635m timing clock.

### LINKED ISSUE #350: SRS CG635 Clock Generator Agent
We need an Agent for the [SRS CG635 Clock Generator](https://www.thinksrs.com/products/CG635.htm). This has GPIB and RS-232 serial outputs, and there exists a python package for communication already: https://pypi.org/project/srs-cg635/

This is for the 122.88 MHz source needed for the timing system.

mjrand:
Added an agent for the srs cg635m clock used for timing in the highbay office.

## Description

This agent tracks the lock registers of the srs clock so we can monitor if the 10MHZ lock ever died on the srs clock.
The agent also tracks the frequency, CMOS output, and the running state as well.

## Motivation and Context

Previously there was no tracking of the 10MHZ lock for the srs clock.
We need to be able to remotely track this lock and srs clock output.

This agent does not add any ability to actually adjust any clock outputs or output states!

Resolves #350.

## How Has This Been Tested?

This agent was tested on daq-dev. All queries worked without error.
Grafana showed all lock statuses as well as the frequency, CMOS setting, and running state with no issues.

## Types of changes

## Checklist:


socs/agents/srs_cg635m/drivers.py
```diff
@@ -0,0 +1,129 @@
...
+
+            if not 0 <= lckr <= 255:
+                # If the lckr register is outside of an 8 bit range
+                raise ValueError
+
+            # Decode the lckr int by performing successive int division and subtractionof 2**(5-i)
+            for i, register in enumerate(list(registers)[::-1]):
+                register_bit = int(lckr / (2**(5 - i)))
+                registers[register] = int(register_bit)
+                lckr -= register_bit * (2**(5 - i))
```
msilvafe, line 124: I think you can do this as is but also could maybe do it a bit more compactly: `register_list = [bool(msg & 1 << (8-i)) for i in range(1,9)]` I tested this as follows:

```
In [64]: msg=int('01110000',2);print(msg)
112

In [65]: [bool(msg & 1 << (8-i)) for i in range(1,9)]
Out[65]: [False, True, True, True, False, False, False, False]

In [66]: msg=int('10000000',2);print(msg)
128

In [67]: [bool(msg & 1 << (8-i)) for i in range(1,9)]
Out[67]: [True, False, False, False, False, False, False, False]

In [68]: msg=int('00000001',2);print(msg)
1

In [69]: [bool(msg & 1 << (8-i)) for i in range(1,9)]
Out[69]: [False, False, False, False, False, False, False, True]
```

Then you can define `registers = {rn: rl for r, rl in zip(reg_names, register_list)}` where `reg_names` is a list of the register key names that you've defined in the dict in line 106.

msilvafe (approved):
Thanks Michael, I think we should test unplugging and plugging back in the 10 MHz clock input and turning the output off and back on (with the run key) and double check that this agent catches the state changes. That can happen during a compute maintenance window otherwise we need to wait for all telescopes to have a break for data taking. Once tested I'm happy to merge, will give us more peace of mind knowing we have alarms on more failure points of the timing system.

socs/agents/srs_cg635m/drivers.py
mjrand: I knew there was probably a much cleaner way to do this that one of the friendly daq reviewers would surely suggest to me. :)

 I can switch this decoding to this way for clarity sake. Though I'd like to keep them as ints so they're easily displayable on grafana, much like the leak detectors.

socs/agents/srs_cg635m/drivers.py
```diff
@@ -0,0 +1,129 @@
+# This device uses the Prologix GPIB interface
+from socs.common.prologix_interface import PrologixInterface
+
+
+class SRS_CG635m_Interface(PrologixInterface):
```
BrianJKoopman, line 5: By convention (and following PEP08) we use CapWords for classnames, so `SRSCG635mInterface`.

socs/agents/srs_cg635m/drivers.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    This device driver is written for the SRS CG635m clock used for the timing system in the SO Office.
```
BrianJKoopman, line 7: I'd drop the "in the SO office", that's overly specific.

socs/agents/srs_cg635m/agent.py
```diff
@@ -0,0 +1,198 @@
...
+
+        return True, 'Initialized Clock.'
+
+    @ocs_agent.param('test_mode', default=False, type=bool)
+    @ocs_agent.param('wait', default=1, type=float)
+    def acq(self, session, params):
+        """acq(wait=1, test_mode=False)
+
+        **Process** - Continuously monitor srs clock lock registers
+        and send info to aggregator.
```
BrianJKoopman, line 73: Replace "send info to aggregator" with "publish to a feed." The current sentence doesn't really capture that it's also going to other subscribers, like the InfluxDB publisher agent, i.e. to Grafana.

socs/agents/srs_cg635m/agent.py
```diff
@@ -0,0 +1,198 @@
...
+from os import environ
+
+import txaio
+from ocs import ocs_agent, site_config
+from ocs.ocs_twisted import TimeoutLock
+
+from socs.agents.srs_cg635m.drivers import SRS_CG635m_Interface
+
+
+class SRSCG635mAgent:
```
BrianJKoopman, line 13: Needs a docstring.

Also, more generally, we need a docs page for this agent in `socs/docs/agents/`. Plenty of examples there, but also see https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html.

This is especially important for deciphering those returned integers.

BrianJKoopman (changes requested):
This looks good to me, I don't have much to add beyond what @msilvafe pointed out, mostly documentation/style suggestions. I do agree that those state change tests that @msilvafe suggested should be done.

docs/agents/srs_cg635.rst
```diff
@@ -0,0 +1,80 @@
...
+
+      {'agent-class': 'SRSCG635Agent',
+       'instance-id': 'srs-cg635',
+       'arguments': [['--ip-address', '10.10.10.10'],
+                     ['--gpib-slot', 23],
+                     ['--mode', 'acq']]},
+
+.. note::
+    The ``--ip-address`` argument should be the IP address of the Prologix GPIB interface.
+    The ``--gpib-slot`` argument should be the GPIB address set on the SRS CG635.
```
msilvafe, line 38: For future reference I think it would be pretty useful to put here how to change the IP address of the prologix and the gpib address of the srs-cg635 since that was mildly annoying and required a little googling and random button smashing.

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+        self.lock = TimeoutLock()
+
+        self.job = None
+        self.ip_address = ip_address
+        self.gpib_slot = gpib_slot
+        self.monitor = False
+
+        self.clock = None
+
+        # Registers Temperature and Voltage feeds
```
msilvafe, line 38: Fix inline comment.

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+                self.clock = SRSCG635Interface(self.ip_address, self.gpib_slot)
+                self.idn = self.clock.identify()
+
+            except (socket.timeout, OSError) as e:
+                self.log.error(f"Clock timed out during connect: {e}")
+                return False, "Timeout"
+            self.log.info("Connected to Clock: {}".format(self.idn))
+
+        # Start data acquisition if requested in site-config
+        auto_acquire = params.get('auto_acquire', False)
```
msilvafe, line 73: Since this is decorated you can just call `params['auto_acquire']`. Can probably just delete this line and call `if params['auto_acquire']:`

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+                }
+            }
+
+        Refer to drivers for interpretation of outputs.
+
+        Parameters
+        ----------
+        wait: float, optional
+            time to wait between measurements [seconds]. Default=1s.
+
```
msilvafe, line 128: describe test_mode too

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+                    'Frequency': 122880000.0
+                    'Standard_CMOS_Output': 3,
+                    'Running_State': 1,
+                    'Timebase': 3,
+                    'PLL_RF_UNLOCKED': 1,
+                    'PLL_19MHZ_UNLOCKED': 1,
+                    'PLL_10MHz_UNLOCKED': 0,
+                    'PLL_RB_UNLOCKED': 1,
+                    'PLL_OUT_UNLOCKED': 0,
+                    'PLL_Phase_Shift': 0,
```
msilvafe, line 118: Is this the stuff that just measures if the state has changed since it was last queried? I find these fields kinda confusing to parse perhaps we just remove them.

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+    """Build the argument parser for the Agent. Allows sphinx to automatically
+    build documentation based on this function.
+
+    """
+    if parser is None:
+        parser = argparse.ArgumentParser()
+
+    # Add options specific to this agent.
+    pgroup = parser.add_argument_group('Agent Options')
+    pgroup.add_argument('--ip-address', type=str, help="Internal GPIB IP Address")
```
msilvafe, line 199: Maybe help should say "prologix IP address"

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+    return parser
+
+
+def main(args=None):
+    # Start logging
+    txaio.start_logging(level=environ.get("LOGLEVEL", "info"))
+
+    parser = site_config.add_arguments()
+
+    # Get the default ocs agrument parser
```
msilvafe, line 214: typo but also don't think you need these inline comments.

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+                                  args=args)
+
+    if args.mode == 'init':
+        init_params = {'auto_acquire': False}
+    if args.mode == 'acq':
+        init_params = {'auto_acquire': True}
+
+    agent, runner = ocs_agent.init_site_agent(args)
+
+    p = SRSCG635Agent(agent, args.ip_address, int(args.gpib_slot))
```
msilvafe, line 228: `args.gpib_slot` always be an int since its specified in the argparse so you can get rid of the int() casting here.

msilvafe (changes requested):
In general I think this is good to go. Most of my inline comments are pretty simple/minimal. Only reason I marked as "request changes" is because I would like you to test out sending multiple scpi queries at once before we close this one out. But thanks for helping test this out and finish it up @davidvng!

socs/agents/srs_cg635/agent.py
```diff
@@ -0,0 +1,237 @@
...
+                        # Loop through the items to add each to the data
+                        lock_statuses = self.clock.get_lock_statuses()
+                        for register, status in lock_statuses.items():
+                            # Not adding PLL causes a naming error
+                            # Two of the registers start with an number
+                            data['data']["PLL_" + register] = status
+
+                    except socket.timeout as e:
+                        self.log.error(f"Timeout in retrieving clock data: {e}")
+                        self.clock = None
```
BrianJKoopman, line 162: Add `session.degraded = True` here too:
```suggestion
                        self.clock = None
                        session.degraded = True
```

msilvafe (approved):
Only comment I have which is unaddressed from my first review is to add instructions on how to set the prologix ip and srs gpib addresses into the `/docs/agents/srs_cg635.rst` file.

BrianJKoopman (approved):
Thanks for adding that new info to the docs. I made one small change, ready to merge when checks pass.

PR #687, opened 2024-06-05
Adds pacemaker to LS372 custom PID loop

jlashner:
Adds pacemaker to regulate temp reading rate to LS372 custom PID loop.

## Description

Adds pacemaker, to LS372 custom pid acquisition loop. 

## Motivation and Context

Nick noticed variable readout rates for temp data with the custom PID. Part of the reason is the lack of timing regulation in the PID control loop. This PR adds a OCS Pacemaker to the body of the pid loop to limit sampling to a specified rate, defaulting to 10 Hz. This will hopefully also speed up and improve PID processing by reducing the number of repeated samples.

## How Has This Been Tested?

Not yet tested.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #753, opened 2024-09-12
Ls372 custom pid update

mjrand:
Fixed the slow response time the custom PID had to going under the setpoint after it was above the setpoint for a long time.

## Description

Updated the I term in the custom_PID function so it could no longer go negative.
Being above the setpoint for a long time would cause the I value to become very negative.
It would take a long time for the custom_PID to overcome that large negative value so the PID would be effectively off for that time.

## Motivation and Context

This issue was prominent after IV curves when the IV curves would heat the whole focal plane above the setpoint
The PID could take upwards of 30 mins to come back online after IV curves because of this.
The custom PID should respond in <5 minutes now, depending on the I value it is set to.
It is not perfect but this relatively simple solution doesn't affect normal PID operations and should improve functionality a lot.

## How Has This Been Tested?

I tested this code on both satp1 and satp2.

For satp2 I simple heated the FPA and then dropped the setpoint to something very low.
The FP was above the setpoint for a few minutes while the FPA cooled. The PID then responded nearly instantly after the FPA dipped below the setpoint and turned the heater on immediately (slowly because the I value was untuned and low).

On satp1 I ran multiple IVs with the updated code. The new custom_PID function turned back on immediately after the IVs were finished and resumed a normal PID operating temperature in < 5minutes after they were done (instead of the ~30 minutes without the changes). I also ran a test scan for 15 minutes with the new custom_PID function to make sure the changes did not affect the operation. The custom_PID operated normally as intended during the scans.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/lakeshore372/agent.py
+++ b/socs/agents/lakeshore372/agent.py
@@ -897,12 +897,12 @@ class LS372_Agent:
     @ocs_agent.param('P', type=float)
     @ocs_agent.param('I', type=float)
     @ocs_agent.param('update_time', type=float)
-    @ocs_agent.param('sample_heater_range', type=float, default=10e-3)
     @ocs_agent.param('query_rate', type=float, default=10.)
+    @ocs_agent.param('sample_heater_range', type=float, default=10e-3)
     @ocs_agent.param('test_mode', type=bool, default=False)
     def custom_pid(self, session, params):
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
Thanks for adding and testing this. I merged in #687, which had one docstring update that was missing here. Will merge shortly.

PR #763, opened 2024-09-29
fixing minor issues in scpi_psu and bluefors agents

CAWNoodle:
log result of scpi_psu's get_output task always returned 'disabled' even when it was actually enabled; bluefors agent not recognizing log file types based on capitalization (e.g. Heaters vs. heaters)

## Description
For SCPI_PSU: set the `enabled` variable based on the outcome of the get_output() driver function. This was the original intent, but the check was lost during rearranging of the session data structure.

For bluefors:
Added the `flags=re.I` option to each of the regex checks to make them case-insensitive.

## Motivation and Context
For SCPI_PSU: The change is mostly for debugging. The correct state of the output was already being reported in the session data, but the log output wasn't properly reflecting this.

For bluefors: The new Bluefors Temperature Controller saves the Heaters log files with a capital 'H' whereas Lakeshore is lower-case. The contents and other formatting is identical however. This change should now treat both cases the same.

## How Has This Been Tested?

Attempted to run `python3 -m pytest --cov` but execution seem to hang at the HWP Supervisor.

Otherwise, the only testing has been with our lab's setup:
Influxdb Agent
Aggregator Agent
Bluefors Agent
SCPI PSU Agent
Fakedata Agent
Labjack Agent
Cryomech CPA Agent

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/bluefors/agent.py
+++ b/socs/agents/bluefors/agent.py
@@ -342,10 +342,10 @@ class LogParser:
                       'heater': '(heaters)'}
 
         for k, v in file_types.items():
-            if re.search(v, filename):
+            if re.search(v, filename, flags=re.I):
                 _type = k
 
-                m = re.search(file_types[_type], filename)
+                m = re.search(file_types[_type], filename, flags=re.I)
                 if _type == 'lakeshore':
                     return _type, "{}_{}".format(_type, m.group(0).lower().replace(' ', '_'))
                 else:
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
This looks good, thanks for the PR @CAWNoodle! Since this contains two unrelated changes though I'd like to split this out to two separate PRs, just for easier tracking when making a new release. Just to avoid further delay with another back and forth, I went ahead and cherry-picked the changes out to #814 and #815. I'm going to close this. Thanks again!

PR #778, opened 2024-10-21
ACU: acuconfigs.yaml setting for forcing sequential moves

mhasself:
## Description

This supports platforms where one generally doesn't want to move az and el axes simultaneously.  When setting "axes_sequential" is enabled in acu-configs.yaml, it will perform all moves as multi-legged, separating az and el.  This applies whether Sun Avoidance is on, or not.  But special handling is applied, for case of Sun Avoidance.

## Motivation and Context

Relates  to https://github.com/simonsobs/scheduler/issues/122, https://github.com/simonsobs/scheduler/pull/100.

## How Has This Been Tested?

Test case; tested on SATP1 today in sequential and simultaneous modes.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #791, opened 2024-11-04
Enable more verbose codecov reporting

BrianJKoopman:
## Description

This re-enables comments, patch statuses, the verbose flag in uploads, and does an ls in the pytest workflow to verify coveage reports are generated properly.

I'll plan to leave this as is for a while until I'm convinced I understand how codecov works. I think the main problem is the stale PRs don't have base reports, and rebasing sometimes doesn't actually work to update the reporting.

## Motivation and Context

PRs aren't showing their coverage status when reports are uploaded. This seems to mostly relate to the fact that the head commits don't have reports, so codecov can't report a difference in coverage. But I tried rebasing #778 to a commit that has a report and it still didn't work.

I'd like to not have codecov create comments, but having that on for now will be helpful, as those seem to report before the status checks.

## How Has This Been Tested?

I've been testing a bunch of configuration changes and scenarios in a [test repo](https://github.com/BrianJKoopman/test). It's sort of convinced me that the configuration we have should be fine, but I still think these changes will be helpful until things are working smoothly.

## Types of changes

## Checklist:

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,6 +1,8 @@
 ---
 codecov:
   branch: main
+  # https://docs.codecov.com/docs/common-recipe-list#see-coverage-on-ci-test-failures
+  require_ci_to_pass: false
 
 # https://docs.codecov.com/docs/pull-request-comments#disable-comment
 comment: true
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
I'm going to admit defeat here, I'm not sure why coverage reports aren't working on PRs, but I'm done trying to get this to work. Closing and will revert back to using coveralls.

PR #794, opened 2024-11-08
Wiregrid Actuator Agent: Add 'position' data in `session.data`

d-hoshino2626:
## Description

I added 'position' data, which means the wiregrid is in the view of the telescope (`'inside'`) or out of the view (`'outside'`), in `session.data`.

## Motivation and Context

This change enables to monitor the location of the wiregrid easily.

## How Has This Been Tested?

no test

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/wiregrid_actuator/agent.py
+++ b/socs/agents/wiregrid_actuator/agent.py
@@ -758,7 +758,9 @@ class WiregridActuatorAgent:
                          'STR2': 0 or 1, (0: OFF, 1:ON)
                          .
                          .
-                         }
+                         },
+                     'position':
+                        'inside' or 'outside' or 'unknown'
                     },
                  'timestamp':1601925677.6914878
                 }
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
This looks good to me, thanks for moving this over from runlib. Any plans to test before we squash and merge?

d-hoshino2626:
Actually, I have no plans to test it because there is no hardware available in Kyoto. However, it might be possible to test it with software, for example, by streaming fake data and observing its response.

BrianJKoopman:
> Actually, I have no plans to test it because there is no hardware available in Kyoto.

Can we coordinate a test at the site during the sun avoidance window? I'd be happy to help with that.

d-hoshino2626:
What kind of test will you do? This PR just add new information to the session.data so I think we have no nessecities to test mechanically..

BrianJKoopman:
> What kind of test will you do?

I was thinking just running the updated agent and inserting/ejecting the wiregrid while checking the reported position makes sense. But maybe this is as simple as you say if one is familiar with which limit switch is which.

d-hoshino2626:
Thanks for explanation! As you say, it would be safer to test. Can you coordinate it? It's just inserting and ejecting, so it takes about 10mins, I think.

BrianJKoopman:
I just tested this, at least partially, on satp1. 'inside' and 'outside' position reporting worked well. All I did during the test was insert then eject the wiregrid.

I did note, however, that during motion the `insert()` and `eject()` tasks hold the lock throughout motion, so position information is not updated during motion. This means it'll be important to check the timestamp is recent enough to be confident of the position when checking the position in sorunlib.

I wasn't able to put it in an 'unknown' position state, as I didn't know how to move the wiregrid into a position that wasn't fully insert or fully ejected.

Let me know what you think of the test and if you think I should test anything else.

d-hoshino2626:
Great thanks! There is a way to move the wiregrid into a half way position, but this is a little tricky and I don't think we necessarily need to test 'unknown' position status. In my opinion, your test is sufficient!

BrianJKoopman (approved):

PR #800, opened 2024-12-09
RTSP Camera Agent: fix docker dependency and make copy for latest file

davidvng:
## Description

- Added `opencv-python-headless` dependency to Dockerfile
- Make a copy of latest file instead of creating symlink

## Motivation and Context

- `opencv-python` only works for local installations, need headless version for Docker
- Symlinks not valid outside of docker container

## How Has This Been Tested?

Tested on `daq-dev`

## Types of changes

## Checklist:


BrianJKoopman (commented):
Looks good, some thoughts on possible configs that wouldn't require changes to the symlink section below.

BrianJKoopman (approved):
Thanks for the updates!

PR #801, opened 2024-12-11
Add check tasks to hwp-pmx

ykyohei:
Add several check tasks to hwp-pmx agent

## Description
Add task to check current/voltage setting and protection limit.
This is fixed once it's set, so I didn't include in the process.
This also includes other minor changes. Changed default protection limits, removed "check" in decollator because it's unnecessary.

## Motivation and Context
New check tasks are necessary for making sure the pmx settings.

## How Has This Been Tested?
Need to test -> Tested by daq-dev

## Types of changes

## Checklist:


socs/agents/hwp_pmx/agent.py
```diff
@@ -184,7 +208,27 @@ def ign_ext(self, session, params):
         return True, 'Set PMX Kikusui to direct control'
 
     @defer.inlineCallbacks
-    @ocs_agent.param('curr', default=0, type=float, check=lambda x: 0 <= x <= 3)
```
BrianJKoopman, line 187: Just checking, this is unnecessary (on `set_i`) because the current limit is set? Still seems reasonable to me to have the check in the event the current limit is set wrong.

socs/agents/hwp_pmx/agent.py
```diff
@@ -212,7 +256,27 @@ def set_v(self, session, params):
...
+        """check_v_lim()
+        **Task** - Check the voltage protection limit.
+        """
+        action = Actions.CheckVLim(**params)
+        self.action_queue.put(action)
+        session.data = yield action.deferred
+        return True, 'Check voltage protection limit is done'
+
+    @defer.inlineCallbacks
+    @ocs_agent.param('curr', default=1.3, type=float)
```
BrianJKoopman, line 279: Default here differs from the docstring of `curr=1`. Assuming 1.3 is the correct value, please update the docstring on line 281.

socs/agents/hwp_pmx/agent.py
```diff
@@ -226,7 +290,7 @@ def set_i_lim(self, session, params):
         return True, 'Set voltage limit is done'
 
     @defer.inlineCallbacks
-    @ocs_agent.param('volt', default=32., type=float, check=lambda x: 0. <= x <= 35.)
+    @ocs_agent.param('volt', default=37., type=float)
```
BrianJKoopman, line 293: Default differs from docstring, please update docstring on line 295.

BrianJKoopman (commented):
Apologies this took so long to get to. A couple of spots that docstrings need updating. And I'm wondering about the removal of checks. I agree the checks on parameters passed for the limits maybe are needed, but why drop the checks on actual current/voltage settings?

socs/agents/hwp_pmx/agent.py
ykyohei: I would like to remove these because it turned out we would like to set larger voltage limit (37V) than the maximum value of 'check' (35V), to avoid the false triggering of the protection.
Also because this device doesn't allow to set the current or voltage larger than its maximum value anyway.

BrianJKoopman (approved):
Looks good, thanks!

PR #816, opened 2025-01-15
RTSPCameraAgent: connection robustness

### LINKED ISSUE #809: RTSP Agent: Error while decoding stream
The RTSP agents all seem to eventually encounter this error and stop collecting images.

```
2024-12-19T20:03:54+0000 startup-op: launching acq
2024-12-19T20:03:54+0000 start called for acq
2024-12-19T20:03:54+0000 acq:0 Status is now "starting".
2024-12-19T20:03:54+0000 acq:0 Status is now "running".
[h264 @ 0x7f84741781c0] error while decoding MB 3 22, bytestream -5
2024-12-25T04:25:29+0000 Failed to retrieve snapshot image from stream
2024-12-25T04:25:29+0000 acq:0 Broken stream
2024-12-25T04:25:29+0000 acq:0 Status is now "done".
```

**socs image tag:** `simonsobs/socs:v0.5.2-15-g7631d0e-dev`

davidvng:
## Description

- Move connection to camera stream to its own function
- Continue loop instead of stopping `acq` process when snapshot fails
- Add `connected` to `session.data` and OCS feed

## Motivation and Context

Resolves #809 

## How Has This Been Tested?

Tested on `daq-dev`

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/rtsp_camera/agent.py
+++ b/socs/agents/rtsp_camera/agent.py
@@ -177,6 +177,18 @@ class RTSPCameraAgent:
         else:
             return False
 
+    def _init_stream(self):
+        """Connect to camera and initialize video stream."""
+        if self.fake:
+            cap = FakeCamera()
+        else:
+            cap = cv2.VideoCapture(self.connection)
+        if not cap.isOpened():
+            self.log.error(f"Cannot open RTSP stream at {self.connection}")
+            return False, "Could not open RTSP stream"
+
+        return cap
+
     @ocs_agent.param("test_mode", default=False, type=bool)
     def acq(self, session, params=None):
         """acq(test_mode=False)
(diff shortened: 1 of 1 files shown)


socs/agents/rtsp_camera/agent.py
```diff
@@ -177,6 +177,18 @@ def _dt_convert(timestr):
...
 
+    def _init_stream(self):
+        """Connect to camera and initialize video stream."""
+        if self.fake:
+            cap = FakeCamera()
+        else:
+            cap = cv2.VideoCapture(self.connection)
+        if not cap.isOpened():
+            self.log.error(f"Cannot open RTSP stream at {self.connection}")
+            return False, "Could not open RTSP stream"
```
BrianJKoopman, line 188: This is returning a tuple if `cap` doesn't open, which results in `not cap:` evaluating to `False` down on line 226, so this will always result in `connected = True`. It made sense to return this tuple within the process, since that's what ocs needs, but now that it's moved best to just return `None`, that will evaluate the connected check correctly.

```python
>>> cap = False, "Could not open RTSP stream"
>>> connected = True
>>> if not cap:
...     connected = False
...
>>> print(connected)
True
```

socs/agents/rtsp_camera/agent.py
```diff
@@ -338,13 +355,9 @@ def record(self, session, params=None):
...
-            if self.fake:
-                cap = FakeCamera()
-            else:
-                cap = cv2.VideoCapture(self.connection)
-            if not cap.isOpened():
-                self.log.error(f"Cannot open RTSP stream at {self.connection}")
-                return False, "Cannot connect to camera"
+            cap = self._init_stream()
+            if not cap:
+                return False
```
BrianJKoopman, line 360: `return False` -> `return False, "Cannot connect to camera"`

BrianJKoopman (changes requested):
Thanks for this! There's one bug with the connected status, but otherwise looks good.

PR #824, opened 2025-02-05
Adds checks to grip_state before attempting HWP spinup

jlashner:
## Description

This adds a check to the HWP supervisor to make sure the HWP is not gripped before attempting to spin up.
This also adds tests of this behavior, to make sure that the HWP cannot spin up while gripped or grip while spinning.

## Motivation and Context

This check is clearly needed, unfortunately it was missing before. From the slack messages I believe this was not attempted on real hardware.

## How Has This Been Tested?

Added tests to validate this behavior.

## Types of changes

## Checklist:


BrianJKoopman (approved):
Looks good, thanks for the new tests!

PR #826, opened 2025-02-07
suprsync: add support to chmod files on transfer

jlashner:
## Description

This PR adds a `chmod` option to the SuprsyncFileHandler that can be used to modify permissions of transferred files.
By default, this sets the default of the suprsync agent so that it adds group-write permissions if they are not there already.

## Motivation and Context

We have seen several instances of problems where level2 file deletion is not able to work properly because sometimes copied files do not give the suprsync group write permissions. This should change that, enforcing that copied files are group-writeable.

## How Has This Been Tested?

Added a test to make sure permissions are adjusted properly.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/suprsync/agent.py
+++ b/socs/agents/suprsync/agent.py
@@ -68,6 +68,7 @@ class SupRsync:
         self.db_echo: bool = args.db_echo
         self.db_pool_size: int = args.db_pool_size
         self.db_pool_max_overflow: int = args.db_pool_max_overflow
+        self.chmod = args.chmod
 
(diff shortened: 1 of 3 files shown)


mhasself:
It's actually g+wx, on the _directories_, that we need, for deleting the files... does this achieve that?

jlashner:
I think the latest commit should do this, by making the default chmod arg `g+rwX`.

mhasself (approved):
This seems like it will do the trick.  Thanks.

PR #837, opened 2025-02-20
hwp_supervisor: add wait in pmx_off, add max_duration for spin up.

ykyohei:
Add wait in pmx_off, add max_duration for spin up.

## Description
- In pmx_off added an option to wait and block other commands until hwp stops.
- Added max_duration for spin up. If the spin up time exceeds max_duration, supervisor calls pmx_off.

## Motivation and Context
- wait of pmx_off is motivated from the scheduler. satp3 has been using pmx_off and time.sleep for safe spin down, but we would like to replace it with this. I plan to update sorunlib to use this by default.
- max_duration is for hwp instrument safety. This will avoid applying large current to motor for a long time when spin up is not successful. Typical spin up time is 7-10 mins.

## How Has This Been Tested?
Tested by daq-dev at satp3

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -803,13 +803,19 @@ class ControlState:
 
         Attributes
         -----------
+        wait_stop : bool
+            Whether to wait until hwp stops
+        freq_tol : float
+            Tolerance of frequency to consider hwp is stopped.
+        freq_tol_duration : float
+            Duration in seconds that the frequency must be within the tolerance
         success : bool
             Whether the last state was completed successfully
-        wait_stop : Optional[bool] default is False
-            Whether to wait until hwp stops
         """
+        wait_stop: bool
+        freq_tol: float
+        freq_tol_duration: float
         success: bool = True
-        wait_stop: Optional[bool] = False
 
     @dataclass
     class GripHWP:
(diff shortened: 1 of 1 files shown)


bbixler500 (approved):
Looks good to me

socs/agents/hwp_supervisor/agent.py
```diff
@@ -719,13 +719,17 @@ class WaitForTargetFreq:
...
+            Maximum duration of time to wait in seconds
         start_time : float
             Time that the state was entered
         """
         target_freq: float
         freq_tol: float
         freq_tol_duration: float
         freq_within_tol_start: Optional[float] = None
+        max_duration: Optional[float] = None
+        start_time: Optional[float] = None
```
jlashner, line 732: Instead of passing this in, you can set this automatically like:
```
start_time: float = field(default_factory=time.time)
```

socs/agents/hwp_supervisor/agent.py
```diff
@@ -1154,6 +1167,7 @@ def check_gripper_ok_for_spinup():
                         target_freq=state.target_freq,
                         freq_tol=state.freq_tol,
                         freq_tol_duration=state.freq_tol_duration,
+                        max_duration=1800,
```
jlashner, line 1170: Instead of hardcoding 1800 multiple places, we should define a global constant at the top like:
```
MAX_SPIN_UP_DURATION: float = 1800.
```
which we can check against inside the update function. 

If we're hardcoding this and not expecting it to change, then I don't think we have to include it as a state field.

jlashner (changes requested):
Thanks Kyohei! This is looking good to me. The one thing I think we should change before merging is how we're hard-coding the max spin-up duration.

The state variable is written as if it is expects to be user-configurable, however its always hardcoded to 1800. If this is a good hard-stop (which I think it is), then I think we should make this a global constant, and not have the "max_duration" as a modifiable state variable.

jlashner (approved):
Thanks Kyohei!

PR #852, opened 2025-03-31
MeinbergSyncboxAgent updates

### LINKED ISSUE #844: Convert string output from SNMP to float in Meinberg Syncbox agent
Some of the output from the Syncbox SNMP interface is provided as strings, for instance the mean path delay, which is reported as "5.111 us". This is inconvenient as it's not easily plotted in Grafana. It'd be better to convert this within the agent before publishing to the OCS feed so that we get floats within InfluxDB/Grafana.

We can't mix types in InfluxDB, so we won't be able to use the same field name we're using without re-writing all of what's currently in InfluxDB. Doing that might be the cleanest solution for users who want to plot this, since if we change the field name, both will still persist within the database.

I suspect the units might change in this string output depending on the magnitude of the offset, so some care will be needed in the conversion.

davidvng:
## Description

Update the syncbox agent to handle conversion of strings into floats for various OIDs, standardizing units to microseconds.
Also adds `--sample-period` as a parameter.

## Motivation and Context

Resolves #844 

## How Has This Been Tested?

Tested on daq-dev

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/meinberg_syncbox/agent.py
+++ b/socs/agents/meinberg_syncbox/agent.py
@@ -17,6 +17,7 @@ CONVERT_OIDS = ['mbgSyncboxN2XPtpUTCOffset',
                 'mbgSyncboxN2XPtpOffsetToGrandmaster',
                 'mbgSyncboxN2XPtpMeanPathDelay']
 
+
 def _extract_oid_field_and_value(get_result):
     """Extract field names and OID values from SNMP GET results.
 
@@ -67,18 +68,19 @@ def _extract_oid_field_and_value(get_result):
     if (field_name.split('_')[0] in CONVERT_OIDS) and (oid_value is not None):
         unit_multipliers = {
             "s": 1000000,   # seconds to microseconds
+            "sec": 1000000,  # seconds to microseconds
             "ms": 1000,     # milliseconds to microseconds
             "us": 1,        # microseconds
             "ns": 0.001     # nanoseconds to microseconds
         }
-        
+
         value, unit = oid_value.split()
         value = float(value)
 
         # Normalize to microseconds
         if unit in unit_multipliers:
             oid_value = value * unit_multipliers[unit]
-        
+
         # Change the field name to differentiate from raw output
         field_name = field_name + "_value"
 
@@ -229,8 +231,7 @@ class MeinbergSyncboxAgent:
                 'mbgSyncboxN2XPtpGrandmasterClockAccuracy',
                 'mbgSyncboxN2XPtpGrandmasterClockVariance',
                 'mbgSyncboxN2XPtpOffsetToGrandmaster',
-                'mbgSyncboxN2XPtpMeanPathDelay',
-                'mbgSyncboxN2XNtpRefSourceOffset']
+                'mbgSyncboxN2XPtpMeanPathDelay']
         output_oids = ['mbgSyncboxN2XOutputMode']
         mib = 'MBG-SYNCBOX-N2X-MIB'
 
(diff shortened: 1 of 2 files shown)


socs/agents/meinberg_syncbox/agent.py
```diff
@@ -280,7 +303,7 @@ def acq(self, session, params=None):
                 {'status': 0,
                  'description': 'tai'},
             'mbgSyncboxN2XPtpUTCOffset_0':
-                {'status': '37 sec',
+                {'status': '37000000',
```
BrianJKoopman, line 306: These all won't be strings anymore, right?
```suggestion
                {'status': 37000000,
```

Also, the `field_name` modification would make this `'mbgSyncboxN2XPtpUTCOffset_0_value'` now too, right?

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -304,10 +327,10 @@ def acq(self, session, params=None):
                 {'status': 13563,
                  'description': '13563'},
             'mbgSyncboxN2XPtpOffsetToGrandmaster_0':
-                {'status': '10 ns',
+                {'status': '0.01',
```
BrianJKoopman, line 330:
```suggestion
                {'status': 0.01,
```

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -304,10 +327,10 @@ def acq(self, session, params=None):
...
                  'description': '10 ns'},
             'mbgSyncboxN2XPtpMeanPathDelay_0':
-                {'status': '875 ns',
+                {'status': '0.875',
```
BrianJKoopman, line 333:
```suggestion
                {'status': 0.875,
```

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -61,6 +65,25 @@ def _extract_oid_field_and_value(get_result):
     if not isinstance(oid_value, (int, bytes, str)):
         oid_value = None
 
+    if (field_name.split('_')[0] in CONVERT_OIDS) and (oid_value is not None):
+        unit_multipliers = {
+            "s": 1000000,   # seconds to microseconds
+            "sec": 1000000,  # seconds to microseconds
+            "ms": 1000,     # milliseconds to microseconds
+            "us": 1,        # microseconds
+            "ns": 0.001     # nanoseconds to microseconds
```
BrianJKoopman, line 74: Small nit, but since they're mostly aligned already...

```suggestion
            "s": 1000000,    # seconds to microseconds
            "sec": 1000000,  # seconds to microseconds
            "ms": 1000,      # milliseconds to microseconds
            "us": 1,         # microseconds
            "ns": 0.001      # nanoseconds to microseconds
```

socs/agents/meinberg_syncbox/agent.py
```diff
@@ -61,6 +65,25 @@ def _extract_oid_field_and_value(get_result):
...
+
+        value, unit = oid_value.split()
+        value = float(value)
+
+        # Normalize to microseconds
+        if unit in unit_multipliers:
+            oid_value = value * unit_multipliers[unit]
+
+        # Change the field name to differentiate from raw output
+        field_name = field_name + "_value"
```
BrianJKoopman, line 85: This is good, and necessary for the InfluxDB type collision issue. However it might be better to add three new fields that are published/kept in `session.data`, rather than outright replace them.

I'm thinking mostly for compatibility with any existing code moving forward. For instance, we now have some code that parses the existing strings for the plots that @msilvafe made. As we continue to compare these old instances of timing issues to new ones that same code will likely be used still, given the old instances will always need it.

Otherwise this should be listed as a breaking change (I checked that box on your PR just now), since these fields will just disappear in the data files once this is deployed.

BrianJKoopman (commented):
Thanks for this! Couple of small comments below.

BrianJKoopman (approved):
Looks good, thanks for the updated changes!

PR #856, opened 2025-04-06
Fix wiregrid_encoder irig

ykyohei:
Satp3 configured irig for wiregrid encoder, irig was not configured for wiregrids at site before. This PR fixes several bugs we found.

## Description
- Fix typo in irig time conversion
- Add missing timestamp in rough irig data.
- Fix irig_fdata. 'quadrature', 'pru_clock', 'reference_degree', 'error' are encoder data not irig data.

## Motivation and Context
These fixes are necessary to properly acquire irig info.

## How Has This Been Tested?
tested at daq-dev for satp3

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/wiregrid_encoder/agent.py
+++ b/socs/agents/wiregrid_encoder/agent.py
@@ -134,12 +134,7 @@ class WiregridEncoderAgent:
             irig_last_updated = 0
             irig_fdata = {'timestamps': [],
                           'block_name': 'wgencoder_irig_raw',
-                          'data': {
-                              'quadrature': [],
-                              'pru_clock': [],
-                              'reference_degree': [],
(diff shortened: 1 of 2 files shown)


sadachi5 (approved):
Thanks for your modification. 
Just for my understanding, could you let me know what was the reason of the memory leak?

ykyohei:
I actually do not understand what is leading to memory leak and not sure wh

ykyohei:
I actually do not understand the cause of memory leak and not sure if this resolves it, but I guess the irig_fdata part might be the cause.

test is failed but I think it's not because of this PR..

BrianJKoopman (approved):
Looks good, thanks!

PR #859, opened 2025-04-16
add new feeds of the number of good/total detectors from tracking result during relock

yoshinori-0778:
## Description

This PR is for adding new feeds of the number of good/total detectors based on tracking results during relock.

## Motivation and Context

Those numbers are good measure for ckecking relock quality.

## How Has This Been Tested?

No. I would like to consult with Max and Brian.

## Types of changes

## Checklist:


socs/agents/pysmurf_controller/agent.py
```diff
@@ -584,16 +585,9 @@ def uxm_relock(self, session, params):
...
-                   'amp_50k_Id': 15.0,
-                   'amp_hemt_Id': 8.0,
-                   'amp_50k_Vg': -0.52,
-                   'amp_hemt_Vg': -0.829,
-                },
-                'noise': {
-                   'band_medians': List of median white noise for each band
+                'filepath': Filepath of saved TrackingResults object
+                'all_det_num': Total number of detectors for each band
+                'good_det_num': Total number of good tracking detectors for each band
```
BrianJKoopman, line 590: The exact structure here will most easily be determined via a test. It's clear to me from looking at a recently run `uxm_relock` at site, that the removed structure here isn't correct. I'm currently seeing:
```
{
  "result": {
    "success": true,
    "return_val": null,
    "traceback": null
  }
}
```

Do you have the ability to test this in the lab?

Also, does `all_det_num` ever change?

socs/agents/pysmurf_controller/smurf_subprocess_util.py
```diff
@@ -157,8 +157,14 @@ def run_uxm_relock(bands=None, kwargs=None):
...
         kwargs = {}
     S, cfg = get_smurf_control()
-    uxm_relock.uxm_relock(S, cfg, bands=bands, **kwargs)
-    return None
+    _, summary = uxm_relock.uxm_relock(S, cfg, bands=bands, **kwargs)
+    tr = summary['tracking_setup_results']
+    data = {
+        'filepath': tr.filepath,
+        'all_det_num': np.array([len(tr.is_good[tr.bands == iband]) for iband in range(0, 8)], dtype=int),
+        'good_det_num': np.array([len(np.where(tr.is_good[tr.bands == iband])[0]) for iband in range(0, 8)], dtype=int),
```
BrianJKoopman, line 165: numpy array's won't work here. ocs won't be able to send them over crossbar as they can't be serialized. Use built in python types, like a `list`.

BrianJKoopman (commented):
Thanks for the PR, couple of small comments/questions inline below.

socs/agents/pysmurf_controller/agent.py
msilvafe: I think some minimal testing could be done at Princeton or slac but they don't have all of the ocs stuff setup the same as the site so I think probably best that could be done would be to subscribe to a feed from the controller agent and confirm that the feeds output in this new branch look right. Possibly it will be faster/easier to test at site if satp1 has more downtime for motor replacement or lat is still holding off on daytime observing in which case we would probably want some help to configure the test to publish to the dev instance for influx instead of the main one.

socs/agents/pysmurf_controller/smurf_subprocess_util.py
yoshinori-0778: Thank you for the comment. I will fix it.

socs/agents/pysmurf_controller/agent.py
yoshinori-0778: `all_det_num` could be changed although it looks stable.

yoshinori-0778:
```
{
  "result": {
    "success": true,
    "return_val": null,
    "traceback": null
  }
}
```
Is this because the current relock in smurf_subprocess_util.py return `None`? 
I modified that part.
[socs/agents/pysmurf_controller/smurf_subprocess_util.py](https://github.com/simonsobs/socs/pull/859/commits/23a695df1b7d0db148c1145bbd34a22bf8aa48e3#diff-f1c5b9c55b13683d02738a9ed4c02fc47df5d7c62fb8e7118fa18a611e9cb422)

yoshinori-0778: @BrianJKoopman What do you think about what Max suggested? I am not familiar with this, so I would like to follow your opnion.

BrianJKoopman: Let's talk next week once the drone calibrations are complete, maybe we can find some time to do the test then.

yoshinori-0778: Sounds good. Thanks

yoshinori-0778: Hi, what do you think how we should test it. SATp1/3 are running normaly and observation gap due to the Sun is not large.  We might be able to use SATp1 when they work jackscrew stuff, but I am not sure when they will do that.

BrianJKoopman: Thanks for the ping, I'll try early next week to figure out when a good time for this test could be.

yoshinori-0778: Thank you!

BrianJKoopman: Okay, so we managed to test this, and the resulting `session.data` structure was:
```
{
  "filepath": "/data/smurf_data/20250519/ufm_mv22/1747687323/outputs/1747687648_tracking_results.npy",
  "all_det_num": [
    226,
    227,
    224,
    234,
    225,
    225,
    208,
    232
  ],
  "good_det_num": [
    184,
    213,
    201,
    172,
    203,
    197,
    190,
    207
  ],
  "result": {
    "success": true,
    "return_val": {
      "filepath": "/data/smurf_data/20250519/ufm_mv22/1747687323/outputs/1747687648_tracking_results.npy",
      "all_det_num": [
        226,
        227,
        224,
        234,
        225,
        225,
        208,
        232
      ],
      "good_det_num": [
        184,
        213,
        201,
        172,
        203,
        197,
        190,
        207
      ]
    },
    "traceback": null
  }
}
```

The strange thing to me here is the duplication of data. The "filepath", "all_det_num", "good_det_num" show up as their own keys/value pairs, but then again under `session.data['result']['return_val']`.

I see this is due to this:
https://github.com/simonsobs/socs/blob/4038404a2269a51a87d1dfbb6ca85684f5fe0e22/socs/agents/pysmurf_controller/agent.py#L82-L87

It's setting `session.data` to the returned value from `run_uxm_setup`, then also inserting the result into the 'result' key, which duplicates the output. I'd need to think about why it's doing this, but it'd be nice to remove this duplication.

BrianJKoopman: I'm looking at other tasks and this duplication is present in their output too. Seems like a general bug, I won't let it hold up this PR. Going to mark this as resolved.

BrianJKoopman (approved):
This looks good to me after testing on site today. @msilvafe any input?

PR #860, opened 2025-04-16
New agent for Dataprobe iBootBar IBB-N20

ykyohei:
This PR introduces an agent for Dataprobe iBootBar IBB-N20

## Description
Make new agent that monitor the state of IBB-N20 and send commands to it via Telnet

## Motivation and Context
This different type of iBootBar is used in satp3, we tried to swap the iBootBar with agent compatible device, iBoot PDU, but it turned out that it's physically not easy to do. Therefore, we make a new agent.

## How Has This Been Tested?
Tested by daq-dev for satp3

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #861, opened 2025-04-22
Make ibootbarAgent compatible with legacy ibootbar

ykyohei:
Make ibootbarAgent compatible with legecy ibootbar

## Description
Make ibootbarAgent compatible with legacy ibootbar. There are differences in SNMP version, MIB. This PR supersedes https://github.com/simonsobs/socs/pull/860.

## Motivation and Context
This legacy iBootBar (iBB-N20) is used in satp3, we tried to swap the iBootBar with agent compatible device, iBoot PDU, but it turned out that it's physically not easy to do. Therefore, we modify the ibootbarAgent to be compatible with legacy ibootbar.

## How Has This Been Tested?
Tested by daq-dev. Tested this works for legecy ibootbar, also confirmed this works for iboot-pdu.
One weird point is that the agent tasks works on legacy ibootbar but I see the following timeout failure in log.

> 2025-04-22T20:19:59+0000 start called for acq
> 2025-04-22T20:19:59+0000 acq:0 Status is now "starting".
> 2025-04-22T20:19:59+0000 acq:0 Status is now "running".
> 2025-04-22T20:20:00+0000 UdpTwistedTransport starting on 53749
> 2025-04-22T20:20:00+0000 Starting protocol <pysnmp.carrier.twisted.dgram.udp.UdpTwistedTransport object at 0x7fa5fd2057e0>
> 2025-04-22T20:20:25+0000 start called for set_outlet
> 2025-04-22T20:20:25+0000 set_outlet:1 Status is now "starting".
> 2025-04-22T20:20:25+0000 set_outlet:1 Status is now "running".
> 2025-04-22T20:20:31+0000 192.168.13.24 failure: [Failure instance: Traceback (failure with no frames): <class 'pysnmp.proto.errind.RequestTimedOut'>: No SNMP response received before timeout
> ]
> 2025-04-22T20:20:31+0000 None
> 2025-04-22T20:20:31+0000 set_outlet:1 Set outlet 2 to off
> 2025-04-22T20:20:31+0000 set_outlet:1 Status is now "done".

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/ibootbar/agent.py
+++ b/socs/agents/ibootbar/agent.py
@@ -164,6 +164,9 @@ class ibootbarAgent:
         Address of the ibootbar.
     port : int
         SNMP port to issue GETs to, default to 161.
+    ibootbar_type : str
+        Type of dataprobe ibootbar (IBOOTPDU or IBOOTBAR), defaults to
+        IBOOTPDU.
     version : int
         SNMP version for communication (1, 2, or 3), defaults to 2.
     lock_outlet : list of ints
@@ -352,7 +355,10 @@ class ibootbarAgent:
                 state = 0
 
             # Issue SNMP SET command to given outlet
-            outlet = [('IBOOTPDU-MIB', 'outletControl', outlet_id)]
+            if self.ibootbar_type == 'IBOOTPDU':
+                outlet = [('IBOOTPDU-MIB', 'outletControl', outlet_id)]
+            elif self.ibootbar_type == 'IBOOTBAR':
+                outlet = [('IBOOTBAR-MIB', 'outletCommand', outlet_id)]
             setcmd = yield self.snmp.set(outlet, self.version, state)
             self.log.info('{}'.format(setcmd))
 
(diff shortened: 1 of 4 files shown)


socs/agents/ibootbar/agent.py
```diff
@@ -414,6 +428,9 @@ def set_initial_state(self, session, params=None):
         Performs a software reboot. The outlets are then set to their
         respective initial states. This takes about 30 seconds.
         """
+        if self.ibootbar_type == 'IBOOTBAR':
+            return True, 'Skipped because this is not supported for IBOOTBAR.'
```
BrianJKoopman, line 432: I'd probably opt to fail out here by returning `False`. I don't think this is a commonly used task, but if someone is using it expecting it to work and checking the return status, they might miss that this is really a no-op.

```suggestion
            return False, 'Software reboot is not supported for IBOOTBAR type devices.'
```

BrianJKoopman (commented):
This looks good to me. One small comment on one of the tasks below. Thanks for working through the MIB/SNMP complications to get this working on this agent!

On SNMP version -- I was having a hard time finding it in the manual, but some devices will have a version selection within the UI on the device. Is the IBOOTBAR just so old it doesn't support v2? Or is it just set to v1 by default? (Which is fairly common in my experience.)

I'm not sure what to make of the timeout you posted. Does it show up *every* time you run with the legacy device. @davidvng might have some insights, as he's worked with many SNMP devices.

ykyohei:
> On SNMP version -- I was having a hard time finding it in the manual, but some devices will have a version selection within the UI on the device. Is the IBOOTBAR just so old it doesn't support v2? Or is it just set to v1 by default? (Which is fairly common in my experience.)

SNMP version is written in manual v1.5, top of p7.
https://www.42u.com/pdf/Dataprobe%20iBootBar%20Manual.pdf
Also I contacted dataprobe to get the MIB for legacy ibootbar and they say it supports v1 only.

ykyohei:
Thanks, I checked it via telnet and the setting seems to match with agent.

```
$ telnet 192.168.13.24 23
Trying 192.168.13.24...
Connected to 192.168.13.24.
Escape character is '^]'.


iBootBar Rev 1.5d.275

User Name:  admin
Password:  *****

iBootBar > get snmp
Write Community: private
Read Community:  public
Manager 1:
  IP Address: 192.168.13.24
  Enabled:    yes
```

BrianJKoopman (approved):
Thanks for the secondary review @davidvng! Going to merge this. (Ignoring the lack of tests being run, I suspect some strange interaction with the commit via suggestion. And that change was small, and tests ran before that fine.)

PR #862, opened 2025-04-24
ACU Agent: generate_scan waits for settling, before exiting

mhasself:
## Description

After it stops uploading points, generate_scan waits for the az to settle at (a) the last commanded position and (b) at velocity zero.  Once those conditions are met, the process exits.  Previously there was just a flat 10 second wait.  There were cases observed where this wasn't long enough.

This does break fromfile_scan.  Sorry.  That's not used in observations but it's useful so I will get it working again, soon.

## Motivation and Context

Resolves a crash seen on SAT where a scan ends, and immediately a new Preset position is commanded (while ProgramTrack is still settling down); the ACU seems to get confused and approach the new target very slowly, which times out the go_to state progress checker.

## How Has This Been Tested?

Tested extensively on simulator and also on SATP3 just now.  Behaves as desired.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -1996,24 +1996,30 @@ class ACUAgent:
 
             # Values for mode are:
             # - 'go' -- keep uploading points (unless there are no more to upload).
-            # - 'stop' -- do not request more points from generator; finish the ones you have.
-            # - 'abort' -- do not upload more points; exit loop and clear stack.
+            # - 'stop' -- do not request more points from generator;
+            #   finish the ones that are already in "points", let the stack empty,
+            #   and wait for settling condition.
+            # - 'abort' -- do not upload more points; exit loop with error; wait
+            #   a few seconds and clear the stack.
             mode = 'go'
 
             lines = []
             last_mode = None
-            was_graceful_exit = True
+            last_upload_az = None
             start_time = time.time()
             got_progtrack = False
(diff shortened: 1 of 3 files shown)


socs/agents/acu/agent.py
```diff
@@ -2067,22 +2072,24 @@ def _run_track(self, session, point_gen, step_time, azonly=False,
...
                     # If the last line has a "group" flag, keep transferring lines.
-                    while len(lines) and len(upload_lines) and upload_lines[-1][0] != 0:
+                    while len(lines) and len(upload_lines) and upload_lines[-1].group_flag != 0:
                         upload_lines.append(lines.pop(0))
 
                     if len(upload_lines):
                         # Discard the group flag and upload all.
-                        text = ''.join([line for _flag, line in upload_lines])
+                        text = sh.get_track_points_text(
+                            upload_lines, timestamp_offset=3, text_block=True)
```
BrianJKoopman, line 2092: Just trying to make sure I follow. This is what's breaking `fromfile_scan`? Because `_run_track` now expects a list of `PointTrack` objects, rather than the already formatted list of text?

When we merge this, would you mind opening an issue about the broken `fromfile_scan` with some details?

BrianJKoopman (approved):
Looks good to me, just one small question and a request to open an issue for the broken task. Going to go ahead and merge.

PR #865, opened 2025-04-29
ACU: motions can be constrained by the HWP state.

mhasself:
## Description

Using the ``hwp_interlocks`` configuration block, the Agent can be asked to monitor state in an HWPSupervisor agent, and use that information to restrict motion in specific axes.

When configured, the Agent's `monitor_hwp` process queries a particular HWPSupervisor instance, to learn the grip and spin state of the HWP.  It combines that information with the current elevation (or range of elevations required for some move) to determine whether a requested motion should be permitted.

## Motivation and Context

These changes complement the HWPSupervisor's interlocks that prevent HWP grip and spin state changes, based on platform position. 

## How Has This Been Tested?

This has been tested extensively with ACU simulator and a dummy HWPSupervisor agent.  Two development versions were tested briefly on real platform, along the way.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -20,7 +20,7 @@ from twisted.internet.defer import DeferredList, inlineCallbacks
 
 from socs.agents.acu import avoidance
 from socs.agents.acu import drivers as sh
-from socs.agents.acu import exercisor
+from socs.agents.acu import exercisor, hwp_iface
 
 #: The number of free ProgramTrack positions, when stack is empty.
 FULL_STACK = 10000
@@ -92,6 +92,8 @@ class ACUAgent:
             If True, don't auto-start idle_reset process for LAT.
         disable_sun_avoidance (bool): If set, start up with Sun
             Avoidance completely disabled.
+        disable_hwp_interlocks (bool): If set, start up with HWP
+            Interlocks disabled.
         min_el (float): If not None, override the default configured
             elevation lower limit.
         max_el (float): If not None, override the default configured
@@ -103,6 +105,7 @@ class ACUAgent:
                  startup=False, ignore_axes=None,
                  disable_idle_reset=False,
                  disable_sun_avoidance=False,
+                 disable_hwp_interlocks=False,
                  min_el=None, max_el=None,
                  ):
 
@@ -188,6 +191,12 @@ class ACUAgent:
             except Exception:
                 agent.log.error('Failed to parse named position "{k}"', k=k)
 
+        # HWP interlocks.
+        self.hwp_rules = hwp_iface.HWPInterlocks.from_dict(
+            self.acu_config.get('hwp_interlocks'))
+        if disable_hwp_interlocks:
+            self.hwp_rules.enabled = False
+
         # Exercise plan.
         self.exercise_plan = self.acu_config.get('exercise_plan')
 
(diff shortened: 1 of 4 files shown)


socs/agents/acu/agent.py
```diff
@@ -2702,15 +2751,228 @@ def _get_sunsafe_moves(self, target_az, target_el, zero_legs_ok=True):
...
+            "third": [false, null, []],
+          }
+        }
+
+        """
+        if not self.hwp_rules.configured:
+            session.data = {
+                'interlocks_config': self.hwp_rules.encoded(basic=True),
+            }
+            return True, "HWP Interlocks not configured."
```
BrianJKoopman, line 2816: Should this not return `False`?

tests/agents/test_acu_agent.py
```diff
@@ -72,3 +119,71 @@ def test_tracks():
...
+    mo_rules = hwp_iface.HWPInterlocks.from_dict(cfg)
+
+    def ex(d, item=0):
+        return tuple([d[k][item] for k in ['el', 'az', 'third']])
+
+    mo_rules = hwp_iface.HWPInterlocks.from_dict(_gen_rules(
+        (40, 70, 'gripped', 'not_spinning', True, False, False),
+        (50, 60, 'ungripped', 'not_spinning', False, True, True),
+        (55, 62, 'ungripped', 'spinning', True, True, False),
+    ))
```
BrianJKoopman, line 168: I'm not quite following this, why define `HWP_IFACE_TEST_CONFIG` and generate the `HWPInterlocks` from that config if we just redefine them on line 164 from this list?

BrianJKoopman (changes requested):
This looks pretty good, a few small comments/questions below.

The one place that I see potential motion commands that isn't checking for HWP safety is in `escape_sun_now`. Can you comment on any potential interaction that should/shouldn't be there?

socs/agents/acu/agent.py
```diff
@@ -1427,16 +1457,21 @@ def go_to(self, session, params):
...
+                self.log.info(f'Executing move via {len(legs) - 1} separate legs (sun optimized)')
+
+            # Check HWP safety
+            hwp_safe, msg = yield self._check_hwpsafe_legs(legs)
+            if not hwp_safe:
+                self.log.info('{msg}', msg=msg)
+                return False, msg
 
-            for leg_az, leg_el in legs:
+            for leg_az, leg_el in legs[1:]:
```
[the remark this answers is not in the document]
mhasself: Good point -- this is an artifact of earlier leg generation code.  I made the two places more similar now -- since len(legs) >= 2 now, always, we can safely walk through legs[1:].

socs/agents/acu/hwp_iface.py
```diff
@@ -0,0 +1,248 @@
...
+        spin_state, to obtain the ranges of els over which moves in
+        each axis is permitted.  Then the el_range, if provided is
+        tested against those ranges.
+
+        The returned structure looks like this::
+
+          {'el': (True, (30, 70), [(30, 70)]),
+           'az': (True, (30, 70), [(30, 70)]),
+           'third': (False, None, [(30, 50), (60, 70)]),
+          }
```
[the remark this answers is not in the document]
mhasself: Functional call added, though the output is harder to interpret in the absence of a full statement of the "rules". This docstring was intended to remind the developer of the returned value's structure.

socs/agents/acu/hwp_iface.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def _simplify_ranges(*a):
+    """Combine and order a set of intervals, to form an ordered,
+    non-abutting list of intevals.
+
+    """
+    def _merge(a, b):
+        if a[0] > b[0]:
+            return _merge(b, a)
```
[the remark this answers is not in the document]
mhasself: Thanks; simplified.

socs/agents/acu/agent.py
mhasself: Yes, totally.  Changed (but had to mess with "startup" a little bit too).

tests/agents/test_acu_agent.py
mhasself: Added comments about this ... I think the idea was to test parsing a "full config" and then test some specific combination of tricky rules.

mhasself:
Thanks for the thorough review!

> The one place that I see potential motion commands that isn't checking for HWP safety is in `escape_sun_now`. Can you comment on any potential interaction that should/shouldn't be there?

The `monitor_hwp` process actively updates the allowed elevation range for use by `escape_sun`. This is guided by the "limit_sun_avoidance" parameter, and described in the docs.

socs/agents/acu/hwp_iface.py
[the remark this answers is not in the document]
BrianJKoopman: Yeah, point taken. My initial read of the description at the bottom of the docstring made me think the second value was the subset of the range that overlaps with `el_range`, i.e. `(40, 50)` if `el_range` was `(40, 50)` and valid ranges were `[(30, 70)]`.

I see now from the example call that the second value is just the full range from the list of ranges that `el_range` is contained in.

tests/agents/test_acu_agent.py
BrianJKoopman: Ah, that makes sense. I'd probably split it into two tests personally -- one that makes sure `mo_rules` has the form you expect when loading from `cfg`, and the other that tests the range combination rules. Your call though.

BrianJKoopman (approved):
Thanks for the updates!

I left one comment unresolved with my thoughts. If you want to split that test, great, if not go ahead with the squash and merge.

PR #868, opened 2025-04-30
Add restart_rssi action to pysmurf controller

tristpinsm:
Using `caput` to directly access the register here is not ideal and it would probably be better to wrap this in a function in `pysmurf`. What do you think?

Pending simonsobs/smurf_dockers#4


tristpinsm:
We should probably also tag and upgrade `sodetlib` as long as we're at it so that fixed tones and other features are included.

tristpinsm:
I wasn't able to test the new action on the SLAC system without disrupting the running OCS agents.

msilvafe:
Looks good to me.

socs/agents/pysmurf_controller/agent.py
```diff
@@ -1138,6 +1138,20 @@ def all_off(self, session, params):
 
             return True, "Everything off"
 
+    def restart_rssi(self, session):
```
BrianJKoopman, line 1141: Could decorate with `@ocs_agent.param('_')`, which is for operations that don't accept any parameters.

BrianJKoopman (approved):
Agreed that a wrapped function in `pysmurf` for this would be better than the `S._caput()` call.

Any plans to test this somewhere, even at site, before we merge?

One comment below, but not super critical to add.

tristpinsm:
> Agreed that a wrapped function in `pysmurf` for this would be better than the `S._caput()` call.
> 
> Any plans to test this somewhere, even at site, before we merge?
> 
> One comment below, but not super critical to add.

re testing: I haven't worked with OCS before and I was trying to figure out how to test it. I think you would have to update the code checked out by the crossbar and restart everything (?) This seemed like it could be disruptive

BrianJKoopman:
> re testing: I haven't worked with OCS before and I was trying to figure out how to test it. I think you would have to update the code checked out by the crossbar and restart everything (?) This seemed like it could be disruptive

It'll sort of depend on your lab setup. You could bring down a pysmurf-controller agent there, and start up this modified agent in its place, then run the new task and verify functionality. This certainly is disruptive if you're actively using the pysmurf-controller. Rolling back once testing is complete is generally straight forward though.

For starting up the modified agent you can either build the image with the modified code in it, or run it directly on the host via something like:
```
$ cd $OCS_CONFIG_DIR
$ OCS_CONFIG_DIR=$(pwd) ocs-agent-cli --site-file=default.yaml --instance-id=pysmurf-controller-id --site-host=hostname
```

You'll need to change `pysmurf-controller-id` and `hostname` to suite your setup.

Building the docker image, you'd need to update the `docker-compose.yml` file with the new image tag and run a `docker compose up -d` (or down/up if you're using Host Manager.) This is the more thorough option, as it would catch any issues with how the software within the image runs, useful especially if the underlying image is going to change with those potential sodetlib changes.

tristpinsm:
OK, thanks for the suggestions. Starting the `pysmurf-controller` docker with the updated code is not an issue. I just don't know how to trigger the new command. The way things are set up there is a "crossbar" server running on another host -- which does not have updated software. So if I spawn a client there it doesn't know about the new `restart_rssi` command.

PR #872, opened 2025-05-12
fix(pysmurf_controller): Bypass smurf instance for RSSI restart.

tristpinsm:
## Description
We found that the `restart_rssi` action fails on a slot that is actually locked-up because this prevents it from instantiating a pysmurf instance, which we had overlooked. This fix bypasses that interface by sending the command directly via EPICS.

## How Has This Been Tested?
Tested on a locked up slot on the LAT.

## Types of changes

## Checklist:


BrianJKoopman (changes requested):
Good catch, just one small request.

BrianJKoopman (approved):

PR #873, opened 2025-05-12
suprsync: Change default chmod to set world-readable

### LINKED ISSUE #871: SupRsync `--chmod` restricts 'others' from being able to read files
#826 introduced the `--chmod` argument to the SupRsync agent. By default this specifies 'g+rwX' as the permissions to pass to rsync's `--chmod` flag, which specifies the permissions on the remote end. rsync will modify existing permissions to include the specified permissions, i.e. it won't change 'user' or 'other' permissions in this default case.

The way SupRsync readies files for transfer on the local (SMuRF server) end of things is it creates a temporary directory and populates it full of symlinks to files it needs to copy. `rsync` is then passed the `-L` argument, in order to copy the files that are linked to the remote. The temporary directory is created using Python's [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), which creates a directory with the same rules as `tempfile.mkdtemp`:
> Creates a temporary directory in the most secure manner possible. There are no race conditions in the directory’s creation. The directory is readable, writable, and searchable only by the creating user ID.

This results in a top level directory that has `drwx------` permissions. With the current default this results in a top level directory on the remote (i.e. `/so/level2-daq/lat/smurf/`) with `drwxrwx---` permissions, when what we really want is `drwxrwxr-x` so that other users can read the files.

BrianJKoopman:
## Description

This PR changes the default `--chmod` argument passed to rsync within the suprsync agent to 'g+rwX,o+rX', which ensures files are group writable and world readable.

## Motivation and Context

Resolves #871.

## How Has This Been Tested?

I tested setting this argument via the SCF on the LAT SMuRF servers. It resulted in changing the top level directory permissions from `drwxrwx---` to `drwxrwxr-x`, which is what we want.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/suprsync/agent.py
+++ b/socs/agents/suprsync/agent.py
@@ -270,7 +270,7 @@ def make_parser(parser=None):
     )
     pgroup.add_argument('--chmod', type=str, default="g+rwX,o+rX",
                         help="Comma-separated chmod strings to apply to file permissions "
-                             "on transfer. Defauls to making sure files are group-writeable "
+                             "on transfer. Defaults to making sure files are group-writeable "
                              "and world-readable.")
     return parser
 
--- a/socs/agents/suprsync/agent.py
+++ b/socs/agents/suprsync/agent.py
@@ -268,9 +268,10 @@ def make_parser(parser=None):
         '--db-pool-max-overflow', type=int, default=10,
         help="Number of connections to allow in the overflow pool."
     )
-    pgroup.add_argument('--chmod', type=str, default="g+rwX",
+    pgroup.add_argument('--chmod', type=str, default="g+rwX,o+rX",
                         help="Comma-separated chmod strings to apply to file permissions "
-                             "on transfer. Defauls to making sure files are group-writeable.")
+                             "on transfer. Defauls to making sure files are group-writeable "
+                             "and world-readable.")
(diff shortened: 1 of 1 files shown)


socs/agents/suprsync/agent.py
```diff
@@ -268,9 +268,10 @@ def make_parser(parser=None):
         '--db-pool-max-overflow', type=int, default=10,
         help="Number of connections to allow in the overflow pool."
     )
-    pgroup.add_argument('--chmod', type=str, default="g+rwX",
+    pgroup.add_argument('--chmod', type=str, default="g+rwX,o+rX",
                         help="Comma-separated chmod strings to apply to file permissions "
-                             "on transfer. Defauls to making sure files are group-writeable.")
+                             "on transfer. Defauls to making sure files are group-writeable "
```
mhasself, line 273: *Defaults

Yup, just doing my part to keep the code up to the highest standards possible.

BrianJKoopman: Heh, thanks. Resolved in https://github.com/simonsobs/socs/pull/873/commits/750f1e1b103a36ad3b34a719f0f2acf0a705eed7.

Issue #864, opened 2025-04-28
ACU: fromfile_scan is broken


mhasself:
It has been broken for a while but it is now very broken because of #862 .  Specifically the `line_batcher` needs to be updated to return TrackPoint objects.

The way it has worked historically is that an operator writes (az, el, etc) vectors to a numpy.save file, and that gets loaded and uploaded as programtrack points.  For the future we should support a few additional capabilities:

- track file should include metadata that specify whether it is to be run in linear or spline mode
- possibly specify whether absolute or relative timestamps are used
- specify what to do with el / corot axes (stop / preset / following)
- track file and agent code should support a "loop" mode, where the loaded data represent a template that can be repeated ad infinitum.

It's a good time to tweak the interface.

PR #876, opened 2025-05-14
 ACU: restore, and then enhance, "fromfile" functionality

mhasself:
(Note this PR includes #865 -- can be rebased once that is merged.)

## Description

- fromfile_scan is now a process rather than a task, because it can run ad infinitum (like generate_scan).
- New file format is supported -- pickle-based, so user can pass a dict with the "free_form" and "loopable" options.
- _run_track now explicitly sets the ACU profiler and interpolation modes. This will ensure that standard generate_scans are always run with those in the classic way.

There are some interface changes here, in fromfile_scan, but that function has not been broadly used and is not implicated in regular observations.

## Motivation and Context

Addresses #864 .

## How Has This Been Tested?

Tested with simulator; tested on LAT to confirm that generate_scan works as before, and also tested a sine-like trajectory for the free_form, fromfile mode.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -20,7 +20,7 @@ from twisted.internet.defer import DeferredList, inlineCallbacks
 
 from socs.agents.acu import avoidance
 from socs.agents.acu import drivers as sh
-from socs.agents.acu import exercisor
+from socs.agents.acu import exercisor, hwp_iface
 
 #: The number of free ProgramTrack positions, when stack is empty.
 FULL_STACK = 10000
@@ -92,6 +92,8 @@ class ACUAgent:
             If True, don't auto-start idle_reset process for LAT.
         disable_sun_avoidance (bool): If set, start up with Sun
             Avoidance completely disabled.
+        disable_hwp_interlocks (bool): If set, start up with HWP
+            Interlocks disabled.
         min_el (float): If not None, override the default configured
             elevation lower limit.
         max_el (float): If not None, override the default configured
@@ -103,6 +105,7 @@ class ACUAgent:
                  startup=False, ignore_axes=None,
                  disable_idle_reset=False,
                  disable_sun_avoidance=False,
+                 disable_hwp_interlocks=False,
                  min_el=None, max_el=None,
                  ):
 
@@ -188,6 +191,12 @@ class ACUAgent:
             except Exception:
                 agent.log.error('Failed to parse named position "{k}"', k=k)
 
+        # HWP interlocks.
+        self.hwp_rules = hwp_iface.HWPInterlocks.from_dict(
+            self.acu_config.get('hwp_interlocks'))
+        if disable_hwp_interlocks:
+            self.hwp_rules.enabled = False
+
         # Exercise plan.
         self.exercise_plan = self.acu_config.get('exercise_plan')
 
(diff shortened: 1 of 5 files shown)


socs/agents/acu/drivers.py
```diff
@@ -80,123 +95,140 @@ def get_track_points_text(tpl, timestamp_offset=None, with_group_flag=False,
...
+            vects['az_flag'] = np.zeros(n, int)
+        if vects['el_flag'] is None:
+            vects['el_flag'] = np.zeros(n, int)
+        if vects['group_flag'] is None:
+            vects['group_flag'] = np.zeros(n, int)
+
+        output.preamble_count = data.get('preamble_count', 0)
+        if data.get('loopable'):
+            # Measure repeat time.
+            output.loop_time = np.dot([-1, 1], vects['timestamp'][[output.preamble_count, -1]])
```
BrianJKoopman, line 216: This syntax seems to assume `vects['timestamp']` is a `np.array`. Based on the docs I would have thought a plain `list` would work here.

socs/agents/acu/drivers.py
```diff
@@ -80,123 +95,140 @@ def get_track_points_text(tpl, timestamp_offset=None, with_group_flag=False,
...
+            vects['el_vel'] = np.gradient(vects['el']) / dtv
+        if vects['az_flag'] is None:
+            vects['az_flag'] = np.zeros(n, int)
+        if vects['el_flag'] is None:
+            vects['el_flag'] = np.zeros(n, int)
+        if vects['group_flag'] is None:
+            vects['group_flag'] = np.zeros(n, int)
+
+        output.preamble_count = data.get('preamble_count', 0)
+        if data.get('loopable'):
```
BrianJKoopman, line 214: When looping, what happens if we end up with a large discontinuity in the loop? Like if the loop is over:
```
az = [0, 1, 2, ..., 99, 100]
```

And not something nice like:
```
az = [0, 1, ..., 99, 100, 99, ..., 1, 0]
```

socs/agents/acu/agent.py
```diff
@@ -1687,68 +1688,67 @@ def _read_modes():
         return True, 'Job completed'
 
     @ocs_agent.param('filename', type=str)
-    @ocs_agent.param('adjust_times', type=bool, default=True)
+    @ocs_agent.param('absolute_times', type=bool, default=False)
     @ocs_agent.param('azonly', type=bool, default=True)
     @inlineCallbacks
     def fromfile_scan(self, session, params=None):
-        """fromfile_scan(filename, adjust_times=True, azonly=True)
+        """fromfile_scan(filename, absolute_times=True, azonly=True)
```
BrianJKoopman, line 1695:
```suggestion
        """fromfile_scan(filename, absolute_times=False, azonly=True)
```

socs/agents/acu/agent.py
```diff
@@ -1687,68 +1688,67 @@ def _read_modes():
...
+                will be a few seconds in the future from when this
+                function is called.
             azonly (bool): If True, the elevation part of the track
                 will be uploaded but the el axis won't be put in
                 ProgramTrack mode.  It might be put in Stop mode
                 though.
 
         Notes:
-            The columns in the numpy array are:
+            See :func:`drivers.from_file` for discussion of the file
```
BrianJKoopman, line 1712: In order for this to link it has to be the full:
```suggestion
            See :func:`socs.agents.acu.drivers.from_file` for discussion of the file
```

If that's too long, you can use:
```suggestion
            See :func:`drivers.from_file <socs.agents.acu.drivers.from_file>` for discussion of the file
```

socs/agents/acu/drivers.py
mhasself: User error. The ACU might give up and raise a fault. Or it might do its best to follow.

mhasself:
> Sorry, one other general comment related to resolving #864. Does this now "specify what to do with el / corot axes (stop / preset / following)"?

No, but it's a step in right direction.  I propose to close that issue and re-open if those features become critical.

BrianJKoopman (approved):

PR #880, opened 2025-05-20
ACU: support passive axes in go_to

mhasself:
## Description

`go_to` will now accept `az=None` and/or `el=None`, in which case the mode for the axis will not be changed.  This is an interface change, but the behavior is backwards compatible with current common usage.

## Motivation and Context

This is required for star camera surveys, where it is desirable to command azimuth moves while either holding elevation in Stop, or in Preset.

## How Has This Been Tested?

Tested with ACU simulator from soaculib.  Interface tested with current ocs-web (which does not propagate az=null, el=null) and an [updated ocs-web](https://github.com/simonsobs/ocs-web/pull/72) that can activate the feature.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -1039,6 +1039,10 @@ class ACUAgent:
         tuple of limits (lower, upper).
 
         """
+        if axis == 'az':
+            axis = 'azimuth'
+        elif axis == 'el':
+            axis = 'elevation'
         limits = self.motion_limits[axis.lower()]
         limits = limits['lower'], limits['upper']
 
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
Looks good to me.

It looks like with az and el both optional on `go_to` that if both are omitted it would be equivalent to a `go_to` to the current location and wouldn't change modes, is that right?

mhasself:
> It looks like with az and el both optional on `go_to` that if both are omitted it would be equivalent to a `go_to` to the current location and wouldn't change modes, is that right?

It is a no-op in that case -- the code will basically check that the current position is Sun Safe.  So it might return an error (not sun safe), but regardless it will return without doing anything to either axis.

PR #881, opened 2025-05-20
Add ability to deactivate slot in smurf crate

msilvafe:
Add a task to disable a smurf slot.

## Description
This adds a task to the `smurf_crate_monitor` agent called `deactivate_slot` which allows turning off a specific slot within the smurf crate.

## Motivation and Context
SATp3 weather stow procedure involves closing their air hatch to their ATCA crates. Current the signals are shutoff but the boards are not powered off before the hatch is closed. To prevent thermal issues its better to power off the slots before closing the hatch which this task allows.

## How Has This Been Tested?
I haven't tested this yet, I would like to test it on one of the site crates.

## Types of changes

## Checklist:

Tagging @samdayweiss, and @tpsatt to let them know I'm adding this in.

### DIFF
--- a/socs/agents/smurf_crate_monitor/agent.py
+++ b/socs/agents/smurf_crate_monitor/agent.py
@@ -271,7 +271,7 @@ class SmurfCrateMonitor:
         else:
             return False, 'acq is not currently running'
 
-    @ocs_agent.param('slot', type=int, choices=[1, 2, 3, 4, 5, 6, 7], default=None)
+    @ocs_agent.param('slot', type=int, choices=[1, 2, 3, 4, 5, 6, 7])
     def deactivate_slot(self, session, params=None):
         """
         deactivate_slot()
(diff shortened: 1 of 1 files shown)


msilvafe:
Ok I tested this on SATp1 so3 in the dev environment at site. I tested starting up the agent, and then connect through client in a notebook, calling the deactivate function, inspecting the session data and then repeating for the activate function.

socs/agents/smurf_crate_monitor/agent.py
```diff
@@ -271,6 +271,78 @@ def _stop_acq(self, session, params=None):
...
+
+        **Task** - Deactivates a specific slot in the crate.
+        This is used to ensure that the boards are not active
+        when the boards are being worked on or air cooling is blocked.
+
+        Args:
+            slot (int):
+                Slot number to deactivate. Allowed values are 1-7.
+        """
+        slot = params.get('slot')
```
mhasself, line 287: Once you've decorated with param, you don't have to use .get anymore.  `params['slot']` is guaranteed to be None or one of those integers.

If you remove the "default=None" line, from the decorator, then None will be forbidden -- so then you don't have to handle that as an error, yourself -- caller will get an automatic informative error.

Issue #875, opened 2025-05-13
Have RSSI Reset Task Return its success


msilvafe:
Currently the `restart_rssi` task in the `pysmurf_controller` agents returns as soon as it has written the `RestartRSSI` rogue register. However this does not ensure that rssi lane was properly restarted. @swh76 has suggested there's another register (which hopefully he can reply with here) we can poll which will tell us if it succeeded. I think we should modify the agent with the following logic:

```
epics.caput(f"smurf_server_s{self.slot}:AMCc:RestartRssi", 1)

state = epics.caget(<our_register_to_check>)
session_data = {"fields": {"rssi_state": state},
                          "last_updated": time.time()}
session.data.update(session_data)
if state == <good_state>:
     return True, "RSSI restarted"
else:
     return False, "RSSI restart failed"
```

Then an ROC will see the task fail in ocs-web in the little output under the button and the return from the caget will get published to session.data which is viewable from the "browser" section of ocs-web. This will help with directions for ROCs. Tagging @BrianJKoopman @tpsatt @swh76 and @tristpinsm for discussion.

PR #896, opened 2025-06-13
pysmurf-controller: Return status from RssiRestart in session.data

tristpinsm:
Check that the system is responsive after the RSSI restart.

## Description
According to the instructions [here](https://simonsobs.atlassian.net/wiki/spaces/PRO/pages/774209554/Restarting+the+SMuRF+RSSI+connection), the RSSI lane being locked up can be identified by querying a given register. This PR performs this check after triggering the RSSI restart to confirm that it was successful and returns the result.

## Motivation and Context
Suggested in #875. It will alert the user if the system is still unresponsive after the restart, or may be useful to rule out an RSSI lock-up when trying to identify an issue by confirming that it is responsive.

## How Has This Been Tested?
Tested at SLAC to confirm that the register is read successfully and the return value is as expected. Not tested on a locked-up system.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/pysmurf_controller/agent.py
+++ b/socs/agents/pysmurf_controller/agent.py
@@ -1148,15 +1148,48 @@ class PysmurfController:
         """restart_rssi()
 
         **Task** - Restarts the RSSI. Use to recover from lock-up.
+
+        Notes
+        ------
+        The following data will be written to the session.data object::
+
+            >> response.session['data']
+            {
+                'rssi_responsive': Boolean indicating whether queries over RSSI succeed,
+                'last_updated': The time of the update,
+            }
         """
         with self.lock.acquire_timeout(0, job='restart_rssi') as acquired:
             if not acquired:
                 return False, f"Operation failed: {self.lock.job} is running."
 
-            # if the system is locked up, spawning a SmurfControl instance will hang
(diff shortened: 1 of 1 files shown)


socs/agents/pysmurf_controller/agent.py
```diff
@@ -1156,7 +1156,18 @@ def restart_rssi(self, session, params):
...
 
-            return True, "RestartRssi sent"
+            # check the system has recovered
+            slot = epics.caget(
+                f"smurf_server_s{self.slot}:AMCc:FpgaTopLevel:AmcCarrierCore:AmcCarrierBsi:SlotNumber"
+            )
+            session.data.update({
+                "fields": {"rssi_responsive": slot is not None},
+                "last_updated": time.time()
+            })
```
BrianJKoopman, line 1166: Great, can you add an example to the docstring for this structure? Here's an example from another operation:
https://github.com/simonsobs/socs/blob/269330b275017407254134366c3c6b1b1daa0e51/socs/agents/pysmurf_controller/agent.py#L423-L432

BrianJKoopman (commented):
Looks good to me, just need to document the `session.data` structure in the docstring. Thanks!

msilvafe (approved):
Looks good to me, thanks @tristpinsm

PR #899, opened 2025-06-17
ACU: fix Sun Avoidance for extended el range

mhasself:
## Description

Extends sun avoidance code and tests to cover situations where a scope is pointed below horizon level, or in the el > 90 range.

## Motivation and Context

The ACU Sun Avoidance code was developed for SAT, considering elevation range (0, 90).  The LAT can move through elevations (-90, 180), which presents new challenges.

The most pressing concern was to recognize that the Earth blocks the Sun, so it's always safe to point at el=-90, even at local midnight.

A full examination yielded lots of other edge cases though, such as how to handle cases where the Sun has just set or is just about to rise.

## How Has This Been Tested?

I tested this extensively using ACU simulator by tweaking the Sun position into key areas (zenith, sunset, sunrise) and testing status variables and escape behavior.  These manual tests led to many, and more pointed unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/avoidance.py
+++ b/socs/agents/acu/avoidance.py
@@ -8,21 +8,33 @@ exclusion zone around the Sun with a user-specified radius. This is
 called the ``exclusion_radius`` or the field-of-view radius.
 
 The Safety of the instrument at any given moment is parametrized
-by two numbers:
+by two numbers and a boolean:
 
   ``sun_dist``
     The separation between the Sun and the boresight (az, el), in
-    degrees.
+    degrees.  This is defined without regards for the horizon --
+    i.e. even if the Sun and the boresight are both below "the
+    horizon", the sun distance can be small.
 
   ``sun_time``
     The minimum time, in seconds, which must elapse before the current
     (az, el) pointing of the boresight will lie within the exclusion
-    radius of the Sun.
+    radius of the Sun.  Although some positions will "never" see the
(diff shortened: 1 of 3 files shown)


socs/agents/acu/avoidance.py
```diff
@@ -332,26 +394,57 @@ def get_sun_pos(self, az=None, el=None, t=None):
         """
         if t is None:
             t = self._now()
-        v = self._sun(t)
-        qsun = quat.rotation_lonlat(v.ra, v.dec)
+        eff_t = t + self.sun_time_shift
 
-        qzen = coords.CelestialSightLine.naive_az_el(t, 0, np.pi / 2).Q
-        neg_zen_az, zen_el, _ = quat.decompose_lonlat(~qzen * qsun)
+        v = self._sun(eff_t)  # This applies sun_time_shift, internally
```
BrianJKoopman, line 399: This comment seems inaccurate now, given `+ self.sun_time_shift` was dropped from within `self._sun()` above. It's now added in `eff_t` on line 397.

BrianJKoopman (commented):
Looks good to me. Good to see all the added tests! Just a couple of small comments.

mhasself:
> Looks good to me. Good to see all the added tests! Just a couple of small comments.

Thanks for the careful read!  All comments addressed now, I believe.

PR #902, opened 2025-06-23
ACU Agent: new LAT datasets

mhasself:
## Description

- Redoes logic for mapping ACU field names into OCS feed block and field names.  This is clearer and includes some helpful log messages.
- Adds two new status blocks with some missing LAT info (pointing correction status and ACU Sun Avoidance)

## Motivation and Context

The current ACU agent won't run because of a missing entry in MONITOR_STRUCTURE.

Debugging that led to clearer code for navigating the map from ACU field name to OCS block / field name.

While debugging that, realized some missing fields from LAT ACU that we should probably be storing.

## How Has This Been Tested?

This is running at site now -- these changes are necessary since the current published ACU agent has a bug.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -94,8 +94,10 @@ OK_RESPONSES = [
 # The sample_period is the minimum spacing between samples, even if
 # values have changed.
 #
-# To drop a block of data (as defined by fields_key) from the output
-# data, set the block_name to None.
+# Note that *all groups found in soaculib.status_keys* must be listed
+# here -- or an error will be raised on startup.  To drop a block of
+# data (as defined by fields_key) from the output data, set the
+# block_name to None.
 
 #: Block names and update policy for status fields in monitor process.
 MONITOR_STRUCTURE = [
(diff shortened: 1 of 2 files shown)


socs/agents/acu/agent.py
```diff
@@ -111,6 +116,9 @@
     ('ACU_corotator', 'corotator', None, None),
     ('ACU_shutter', 'shutter', None, None),
     ('ACU_tilt', 'tilt_slow', 'changed', 0.5),
+    (None, 'tilt_fast', None, None),
```
BrianJKoopman, line 119: This (having `None` for the `block_name`) just means the `tilt_fast` fields aren't being published to a feed, right?

socs/agents/acu/agent.py
```diff
@@ -191,7 +201,18 @@ def __init__(self, agent, acu_config='guess',
...
+        # Create a map from each status key (read through the
+        # self.datasets) to the output block and field name.
+        self.status_field_map = {}
+        for group, group_fields in \
+                status_keys.status_fields[self.platform_type]['status_fields'].items():
+            for block_name, block_group, _, _ in MONITOR_STRUCTURE:
+                if block_group == group:
+                    break
+            else:
+                raise ValueError(f"status_key block '{group}' not configured for monitoring.")
```
BrianJKoopman, line 213: This looks like it'll crash the agent on start up. Is that the desired behavior? And if a `status_key` isn't configured, that means we just need to add a line for it in `MONITOR_STRUCTURE`?

BrianJKoopman (commented):
Just a couple of questions, otherwise looks good to me. Thanks for the bug fixes!

socs/agents/acu/agent.py
mhasself: Yes... these numbers turned out to be useless.

BrianJKoopman (approved):

PR #905, opened 2025-07-24
ACU: set_shutter will not fail if already in req state

mhasself:
## Description

set_shutter(target) will quickly succeed, rather than fail, if shutter is already state target.

## Motivation and Context

This is a more convenient functionality in practice; otherwise caller has to check state before requesting desired state.

## How Has This Been Tested?

Tested on LAT today.

## Types of changes

## Checklist:


BrianJKoopman (approved):
Looks good to me. I hadn't noticed it would crash if commanded to the current position.

PR #914, opened 2025-08-06
ACU: add broadcast_ext process to sample the PositionBroadcast_Ext 200 Hz streams

mhasself:
## Description

Adds a new process and feeds, which will only be available if the 'ext' stream is defined in ACU config.  Refactors UDP sampling code so it can be re-used.

This should not change feed/field names for the usual broadcast 200 Hz / influx data.

## Motivation and Context

For fast sampling of the internal servo computations, a bunch of 200 Hz fields have been added to the 200 Hz UDP stream called PositionBroadcast_Ext.  The idea is that operators would enable this stream, when desired, to record those fields at full speed.  But otherwise it remains off.

## How Has This Been Tested?

Tested with simulator, and I will soon test it also on LAT in a dev env.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -181,10 +181,6 @@ class ACUAgent:
         self.acu_config = aculib.guess_config(acu_config)
         self.platform_type = self.acu_config['platform']  # ccat, satp.
 
-        self.udp = self.acu_config['streams']['main']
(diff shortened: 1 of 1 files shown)


mhasself:
This has now been tested on daq-dev.

BrianJKoopman (approved):
This looks good to me. Thanks for testing on `daq-dev` to confirm fields don't change.

PR #916, opened 2025-08-11
hwp-pmx: Make PMX a subclass of TCPInterface

BrianJKoopman:
## Description

This PR primarily is to make the `PMX` class a subclass of `TCPInterface`, so we can take advantage of the exception handling/re-connection logic in there.

A couple of extra things I did while I was here:
* Made the `PMX.wait()` method private -- I debated removing it entirely, it's adding a 10 ms sleep after a 500 ms sleep (built into `send_message()`, so honestly probably isn't really doing much.
* Moved adding terminating characters and encoding to the message strings to within `send_message()`, rather than handling it individually within each method that builds a message to send.

When I started this, I hadn't realized that this agent used the newer "no timeout lock" structure, and already had quite a bit of error handling for robustness in it. I didn't touch the errback handling in `_process_actions()`, only the handling for the main `_get_and_publish_data` method.

## Motivation and Context

We're hoping this will help with some recent issues that appear to be network related. It's also a step towards standardizing TCP based agents on this base class.

## How Has This Been Tested?

I've run the integration tests locally, but have not tested with any actual hardware.

## Types of changes

- [x] Code cleanup/organization

## Checklist:

### DIFF
--- a/socs/agents/hwp_pmx/agent.py
+++ b/socs/agents/hwp_pmx/agent.py
@@ -343,28 +343,29 @@ class HWPPMXAgent:
                  'last_updated': 1649085992.719602}
 
         """
-        PMX = None
+        PMX = pmx.PMX(ip_address=self.ip, port=self.port)
 
         threads.blockingCallFromThread(reactor, self._clear_queue)
 
         sleep_time = 1 / self.f_sample - 0.01
         last_daq = 0
         while session.status in ['starting', 'running']:
-            if PMX is None:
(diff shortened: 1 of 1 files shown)


ykyohei (approved):
This looks good to me, and I have tested it with actual hardware.
Sorry for the delay; I didn't want to make many changes until we resolved the PMX connection issues. I think we now understand that the cause was mDNS, so I'm happy to merge this.

PR #919, opened 2025-08-26
add new feed for relock alarm

yoshinori-0778:
This PR is to add new feed for relock.

## Description

Based on the feeds (number of good/all detectors) we have, just add new feed being calculated as 1 if bad(all - good) > good, else 0.

## Motivation and Context

This is for new alarm that check relock quality.
In terms of making new alarm of this, calculating this with the feeds (number of good/all detectors) we have after publishing is computational intense. Thus, I would like to make those values here and publish them in addition to the current feeds

## How Has This Been Tested?

Not yet on the site computer but I hope we do not need to test because this is simple change.
Script itself has been tested as below, which is expected.

```
alls = np.array([100,100,100,100,100,100])
goods = np.array([30,40,49,50,51,80])
block_data = {}
for iband, (iall, igood) in enumerate(zip(alls, goods)):
    block_data[f'alldet_band{iband}'] = iall
    block_data[f'gooddet_band{iband}'] = igood
    block_data[f'alarm_band{iband}'] = 1 if (iall - igood) > igood else 0
    #print(iall - igood, igood)

block_data -->
{'alldet_band0': 100,
 'gooddet_band0': 30,
 'alarm_band0': 1,
 'alldet_band1': 100,
 'gooddet_band1': 40,
 'alarm_band1': 1,
 'alldet_band2': 100,
 'gooddet_band2': 49,
 'alarm_band2': 1,
 'alldet_band3': 100,
 'gooddet_band3': 50,
 'alarm_band3': 0,
 'alldet_band4': 100,
 'gooddet_band4': 51,
 'alarm_band4': 0,
 'alldet_band5': 100,
 'gooddet_band5': 80,
 'alarm_band5': 0}
```

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/pysmurf_controller/agent.py
+++ b/socs/agents/pysmurf_controller/agent.py
@@ -129,6 +129,8 @@ class PysmurfController:
                 'observatory.{}.feeds.pysmurf_session_data'.format(args.monitor_id),
             )
 
+        self.good_threshold_relock = args.good_threshold
+
         self.agent.register_feed('tracking_results', record=True)
         self.agent.register_feed('bias_step_results', record=True)
         self.agent.register_feed('noise_results', record=True)
@@ -623,7 +625,7 @@ class PysmurfController:
                 for iband, (iall, igood) in enumerate(zip(result.return_val['all_det_num'], result.return_val['good_det_num'])):
                     block_data[f'alldet_band{iband}'] = iall
                     block_data[f'gooddet_band{iband}'] = igood
-                    block_data[f'alarm_band{iband}'] = 1 if (iall - igood) > igood else 0
+                    block_data[f'alarm_band{iband}'] = 1 if igood/iall < self.good_threshold_relock else 0
                 data = {
                     'timestamp': time.time(),
                     'block_name': 'tracking_results',
@@ -1220,6 +1222,8 @@ def make_parser(parser=None):
     pgroup.add_argument('--slot', type=int,
                         help="Smurf slot that this agent will be controlling")
     pgroup.add_argument('--poll-interval', type=float,
+                        help="Time between check-state polls")
+    pgroup.add_argument('--good_threshold', type=float, default=0.5,
                         help="Time between check-state polls")
     return parser
 
--- a/socs/agents/pysmurf_controller/agent.py
+++ b/socs/agents/pysmurf_controller/agent.py
@@ -1223,7 +1223,7 @@ def make_parser(parser=None):
                         help="Smurf slot that this agent will be controlling")
     pgroup.add_argument('--poll-interval', type=float,
                         help="Time between check-state polls")
-    pgroup.add_argument('--good_threshold_relock', type=float, default=0.5,
+    pgroup.add_argument('--good-threshold-relock', type=float, default=0.5,
                         help="Alarm system threshold for relock quality check")
     return parser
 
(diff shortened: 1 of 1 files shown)


yoshinori-0778:
Thanks for checking, Brian.
This alarm would be one alarm for each slot, and each alarm is checking all 8 bands for that slot and would fire if any of the 8 bands pass the threshold, so 7 alarms for one SAT. @davidvng can comment as well.

50/50 threshold is based on our experience. If Relock looks bad, it was completely bad (i.e., number of bad dets >> number of good dets), so this threshold should be good to detect it as a first step. Of course we can change it later as you commentd.

BrianJKoopman:
> Do you mean I should add number of 50% (e.g. `int(f'alldet_band{iband}/2'` )) instead of adding `alarm_band{iband}` that I added?

No, more like:
```
block_data[f'alarm_band{iband}'] = 1 if igood/iall < good_threshold else 0
```

where `good_threshold` gets passed in via the config file, so we have some control over how sensitive the alarm is -- though it sounds like it's either they're mostly bad or they are mostly good.

yoshinori-0778:
I see, that makes sense

Where is the config file? https://github.com/simonsobs/sodetlib/blob/master/sodetlib/det_config.py? or something new?
Once I can get it, I will modify the script and config file.

BrianJKoopman:
> Where is the config file? https://github.com/simonsobs/sodetlib/blob/master/sodetlib/det_config.py? or something new? Once I can get it, I will modify the script and config file.

I was originally thinking in the ocs site config file (SCF), but that is an awkward mechanism for single task items like this, since it requires making the option a command line argument.

I'm not against putting it in the `sys_config.yml` as you suggest, that sounds like it would be simplest. I would say try to load it from there and otherwise default to 50%. That way we don't have to go update all the sys_config files unless we need to change this initial value.

BrianJKoopman:
On second thought, I'm not sure it really belongs there either, that would require updates to the [sodetlib documentation for the structure of that configuration file](https://sodetlib.readthedocs.io/en/latest/configs.html#docker-environment).

Really what we need is a more general config file format for ocs agents, but that isn't going to be ready for this application.

For now, the best thing would be to put it in the SCF, even if it is awkward. That involves adding an argument to the parser here: https://github.com/simonsobs/socs/blob/f40b3355d807b1094ef64dadc402cbd864282e5d/socs/agents/pysmurf_controller/agent.py#L1208-L1223

The whole arguments list gets passed to the `PysmurfController` object when it's created, you can then use it from there.

yoshinori-0778:
Thank you for your comments, Brian!
I modified the script based on you comments.
Is this what you expected?

socs/agents/pysmurf_controller/agent.py
```diff
@@ -1220,6 +1223,8 @@ def make_parser(parser=None):
                         help="Smurf slot that this agent will be controlling")
     pgroup.add_argument('--poll-interval', type=float,
                         help="Time between check-state polls")
+    pgroup.add_argument('--good_threshold', type=float, default=0.5,
+                        help="Time between check-state polls")
```
BrianJKoopman, line 1227: Update the 'help' here to describe the new argument.

yoshinori-0778: Thank you for catching it, fixed.

yoshinori-0778:
Thanks, Brian!
Do you think we need to test on site computers before merge?
If not I would like to merger it into main once current checks are finished.

BrianJKoopman:
> Thanks, Brian! Do you think we need to test on site computers before merge? If not I would like to merger it into main once current checks are finished.

I think we're good, this is a small change and you already caught a good edge case from the tests. I'm planning on merging once the checks complete.

yoshinori-0778:
Sounds great, thanks Brian!

yoshinori-0778:
Once this is implemented on the site computer, could you let me know so I can ask David to make the alarm.
Thanks!

PR #920, opened 2025-08-28
wg-actuator: Fix limit switch issue at extreme boresight angles

sadachi5:
## Description

To avoid limit switch issue, modify the `wiregrid_actuator/agent.py` and add additional small (0.5mm) moving after the insertion and ejection.

## Motivation and Context

There is a known issue that the outside limit switch turns OFF right after ejection, probably due to the gravity force of the wiregrid rotator at boresight=-45deg.
To avoid it, I added an additional small movement (0.5 mm) at the last step of the insertion and ejection to ensure that the limit switch is ON.
The limit switch lever has a 1.1 mm margin in the ON condition. So, the 0.5 mm additional movement is no problem for the limit switch.
(The limit switch: Omron D2JW-01K31-MD)
image

## How Has This Been Tested?

Need the operation test at the site to check if the insert/eject works as expected.
Run insert/eject several times at SATp1 at boresight = 0, -45, +45, and check the limit switch turns ON as expected.
El should be the nominal elevation (60 deg).

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/wiregrid_actuator/agent.py
+++ b/socs/agents/wiregrid_actuator/agent.py
@@ -242,10 +242,11 @@ class WiregridActuatorAgent:
             self.log.error(msg)
             return False, msg
 
-        # Additional small moving to ensure the limit switch ON
-        # No limit switch check during this moving
-        # Only this function reflects pos/neg of <distance>.
-        # Please use +/- distance for insertion/ejection, respectively.
+        # Additional small move to ensure the limit switch ON
+        # No limit switch check during this move
+        # `self.actuator.move()` preserves the sign of `distance`.
+        # In contrast, `move_func()` sets the sign automatically based on `is_insert`.
+        # Pass a positive distance for insertion and a negative distance for ejection.
         ret = self.actuator.move(0.5 * (1. if is_insert else -1.), speedrate=0.2)
         if not ret:
             msg = '_insert_eject(): ERROR!:'\


socs/agents/wiregrid_actuator/agent.py
```diff
@@ -242,6 +242,17 @@ def _insert_eject(
             self.log.error(msg)
             return False, msg
 
+        # Additional small moving to ensure the limit switch ON
+        # No limit switch check during this moving
+        # Only this function reflects pos/neg of <distance>.
+        # Please use +/- distance for insertion/ejection, respectively.
+        ret = self.actuator.move(0.5 * (float)(1. if is_insert else -1.), speedrate=0.2)
```
BrianJKoopman, line 249:
```suggestion
        ret = self.actuator.move(0.5 * (1. if is_insert else -1.), speedrate=0.2)
```
The `(float)()` syntax is unusual in Python. We don't really need to be casting `1.` or `-1.` to `float` here anyway, so I'd just drop it.

socs/agents/wiregrid_actuator/agent.py
```diff
@@ -242,6 +242,17 @@ def _insert_eject(
...
+        if not ret:
+            msg = '_insert_eject(): ERROR!:'\
+                  'Failed to move in the additional small moving after the last slow moving.'
```
BrianJKoopman, line 252:
```suggestion
                  'Failed to move in the additional small move after the last slow move.'
```

Just a small grammar suggestion.

socs/agents/wiregrid_actuator/agent.py
```diff
@@ -242,6 +242,17 @@ def _insert_eject(
             self.log.error(msg)
             return False, msg
 
+        # Additional small moving to ensure the limit switch ON
+        # No limit switch check during this moving
+        # Only this function reflects pos/neg of <distance>.
+        # Please use +/- distance for insertion/ejection, respectively.
```
BrianJKoopman, line 248: Is this flipping a sign? `move_func` is `self._forward` if `is_insert`, and `self._forward` takes a positive distance always:
https://github.com/simonsobs/socs/blob/f40b3355d807b1094ef64dadc402cbd864282e5d/socs/agents/wiregrid_actuator/agent.py#L101-L105

So it seems like this is also moving in the typical expected direction to me.

BrianJKoopman (commented):
Thanks for the PR. A couple of questions/comments about the implementation here. And am I understanding this hasn't been tested on site yet? Do you have plans to do so?

socs/agents/wiregrid_actuator/agent.py
sadachi5: I mean that `move_func` (=`_forward` or `_backward`) will change the sign of `distance` automatically, but `self.actuator.move`, which is used here, will not change it.
Sorry for confusing you. I tried to modify the comments. Please check it.

sadachi5:
> Thanks for the PR. A couple of questions/comments about the implementation here. And am I understanding this hasn't been tested on site yet? Do you have plans to do so?

Yes. I want to test it. Can I test it on the site before PR? Sorry, actually, I don't know how to do that before PR.
Let me know it.

BrianJKoopman (approved):
> > Thanks for the PR. A couple of questions/comments about the implementation here. And am I understanding this hasn't been tested on site yet? Do you have plans to do so?
> 
> Yes. I want to test it. Can I test it on the site before PR? Sorry, actually, I don't know how to do that before PR. Let me know it.

Great, the changes look good, thanks for those. Approving for now. I'll message you about testing before we merge.

Issue #891, opened 2025-06-01
wiregrid_tiltsensor RecursionError


ykyohei:
satp3 wiregrid_tiltsensor agent crashed with RecursionError
acu was stable at az = 200, el = 60, roll = 0 around this error.
`2025-05-30T15:25:35+0000 acq:0 CRASH: [Failure instance: Traceback: <class 'RecursionError'>: maximum recursion depth exceeded in comparison`

ykyohei:
satp1 agent was also crashing at the same time with different error.
```
2025-05-30T15:25:35+0000 acq:0 Status is now "done".
]
/opt/venv/lib/python3.10/site-packages/socs/common/moxa_serial.py:138:readbuf
/opt/venv/lib/python3.10/site-packages/socs/common/moxa_serial.py:172:read
/opt/venv/lib/python3.10/site-packages/socs/agents/wiregrid_tiltsensor/drivers/dwl.py:48:get_angle
/opt/venv/lib/python3.10/site-packages/socs/agents/wiregrid_tiltsensor/agent.py:101:acq
/opt/venv/lib/python3.10/site-packages/ocs/ocs_agent.py:984:_running_wrapper
/opt/venv/lib/python3.10/site-packages/twisted/python/context.py:82:callWithContext
/opt/venv/lib/python3.10/site-packages/twisted/python/context.py:117:callWithContext
/opt/venv/lib/python3.10/site-packages/twisted/python/threadpool.py:285:<lambda>
/opt/venv/lib/python3.10/site-packages/twisted/python/threadpool.py:269:inContext
2025-05-30T15:25:35+0000 acq:0 CRASH: [Failure instance: Traceback: <class 'TypeError'>: can't concat str to bytes
```

BrianJKoopman:
Thanks for filing this! It sounds possibly related to https://github.com/simonsobs/socs/issues/784.

mjrand:
It looks like the error is coming from this function:

```
 def readbuf(self, n):
        """Returns whatever is currently in the buffer. Suitable for large
        buffers.

        Args:
            n: Number of bytes to read.

        """
        if n == 0:
            return ''
        try:
            msg = self.sock.recv(n)
        except BaseException:
            msg = ''
        n2 = min(n - len(msg), n / 2)
        return msg + self.readbuf(n2)
```

specficially the last two lines: 

```
n2 = min(n - len(msg), n / 2)
return msg + self.readbuf(n2)
```

[...]

PR #924, opened 2025-09-10
Fix wiregrid tiltsensor recursion error / Update tiltsensor agent to reconnect to the sensor automatically

sadachi5:
## Description

- Change the `socs/common/moxa_serial.py` to have better error handle
- Change the wiregrid tilt-sensor agent to reconnect to the sensor automatically if the read data is wrong.

## Motivation and Context

The change to the `moxa_serial.py` is related to the issue: #891 
The change to the tilt-sensor agent is useful to recover from network disconnection automatically.

## How Has This Been Tested?

I ran the updated agent in SATp1 on 10th Sept.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/wiregrid_tiltsensor/agent.py
+++ b/socs/agents/wiregrid_tiltsensor/agent.py
@@ -33,7 +33,7 @@ class WiregridTiltSensorAgent:
         self.ip = ip
         self.port = int(port)
         self.sensor_type = sensor_type
-        self.tiltsensor = connect(self.ip, self.port, self.sensor_type)
+        self._connect()
         self.pm = Pacemaker(2, quantize=True)
 
         agg_params = {'frame_length': 60}
@@ -41,6 +41,19 @@ class WiregridTiltSensorAgent:
                                  record=True,
                                  agg_params=agg_params,
                                  buffer_time=1.)
+
+
+    ######################
+    # Internal functions #
+    ######################
+
+    # Return: Nothing
+    def _connect(self):
+        self.tiltsensor = connect(self.ip, self.port, self.sensor_type)
+
+    ##################
+    # Main functions #
+    ##################
 
     def acq(self, session, params=None):
         """acq()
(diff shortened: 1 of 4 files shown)


socs/agents/wiregrid_tiltsensor/agent.py
```diff
@@ -99,8 +112,25 @@ def acq(self, session, params=None):
                 # data taking
                 current_time = time.time()
                 msg, angles = self.tiltsensor.get_angle()
-                if self.sensor_type == 'sherborne':
-                    msg, temperatures = self.tiltsensor.get_temp()
```
BrianJKoopman, line 103: This change drops the definition of `temperatures`, which is later accessed down on line 139+140. It's now undefined which is why [pre-commit is failing](https://results.pre-commit.ci/run/github/186511668/1757500927.VV05wgo-SXm3lHCPwuGEAQ). This needs to be addressed before merging.

socs/agents/wiregrid_tiltsensor/agent.py
```diff
@@ -99,8 +112,25 @@ def acq(self, session, params=None):
...
+                self.log.info(f'self.tiltsensor.get_angle(): {msg}')
+                if angles[0] == -999:
+                    msg = 'Failed to get angle. -> Reconnect to the sensor'
+                    self.log.warn(msg)
+                    del self.tiltsensor
+                    self._connect()
+                    self.pm.sleep()  # DAQ interval
+                    continue
+                # No implementation of `get_temp()` now and ignore the below lines
+                # if self.sensor_type == 'sherborne':
```
BrianJKoopman, line 124: Are we just removing support for the 'sherbone' sensor type? Can you comment more about this removal?

socs/common/moxa_serial.py
```diff
@@ -1,3 +1,5 @@
+# 9/2025 Shun
+# Change all BaseExeption --> TimeoutError
```
BrianJKoopman, line 2: Unless you feel strongly about including this changelog in the file, I'd suggest we just rely on `git blame` for this sort of detail.

This code was imported from an older project "as is", which is the only reason this old comment was left here.

socs/common/moxa_serial.py
```diff
@@ -132,9 +134,9 @@ def readbuf(self, n):
             return ''
         try:
             msg = self.sock.recv(n)
-        except BaseException:
+        except (TimeoutError, BlockingIOError):
             msg = ''
-        n2 = min(n - len(msg), n / 2)
+        n2 = min(n - len(msg), (int)(n / 2))
```
BrianJKoopman, line 139:
```suggestion
        n2 = min(n - len(msg), int(n / 2))
```

BrianJKoopman (changes requested):
Thanks for addressing this. The main thing here is we need to get pre-commit to pass before we can merge. I'd also like some details on the possible removal of the sherbone sensor support.

socs/common/moxa_serial.py
```diff
@@ -114,28 +114,33 @@ def readbuf_slow(self, n):
...
+        n_current = n
+        msg = ''
+        for i in range(max_loop):
+            if n_current <= 0:
+                return msg
+            try:
+                msg_current = self.sock.recv(n)
+            except (TimeoutError, BlockingIOError):
+                msg_current = ''
+            msg += msg_current
```
BrianJKoopman, line 140: Looking back at the error that @ykyohei posted in https://github.com/simonsobs/socs/issues/891#issuecomment-2928017025, it looks like this will fail if `msg` actually has anything in it, as it will be a bytes object:
```
>>> msg = b'response from device'
>>> msg_current = ''
>>> msg += msg_current
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can't concat str to bytes
```

Easiest fix is to make `msg_current = b''`.

BrianJKoopman (changes requested):
Thanks for the updates. This new structure should resolve the max recursion depth error reported in #891. Looking back at that issue there's also the bytes concatenation that we can fix right now too.

I'm also still curious, are we dropping support for the sherborne type sensors moving forward?

BrianJKoopman (approved):
Approving this, though I'm going to split the PR into two. This will remain for the automatic reconnection, and a new one will contain the fix for the recursion error. Thanks again for addressing this issue!

BrianJKoopman:
Alright, this'll be ready to merge when checks are complete. Likely won't merge until tomorrow.

PR #930, opened 2025-09-22
Added three_leg_turnaround generator

mjrand:
I've added in a new python file to generate three leg turnarounds and modified the `generate_constant_velocity_scan` in the ACU `drivers.py` to accept the new turnaround function.

## Description

This function is the implementation of the three part turnarounds discussed on last weeks SAT Analysis call.
This turnaround function breaks the turnarounds into three "legs":

1. The initial deceleration.
2. The middle portion with a gentle/zero velocity acceleration to allow the motor gears to change direction and come in contact with the opposite azimuth bearing face more gently.
3. The final acceleration to the scan velocity.

This should hopefully make the turnarounds more gentle for the SATs. 

I have renamed the turnarounds from "full_stop_turnarounds" to "three_leg_turnarounds". I am not married to this naming scheme and I am OK with changing any names in this PR prior to approval. 

A few other points of discussion:
1. This current implementation forces the time of the second leg of the turnaround to be 1 second. This limits the minimum turnaround time to ~2 seconds (so that there is at least 0.5 seconds for the first and third leg). This hard coded 1 second is only because that is the only time I've tested for the second leg. This is really only a problem for planet scans where the turnaround time is only (2 * 0.8 deg/s) / 1.0deg/s^2 = 1.6 seconds. I think the solution to this is testing an implementation where each leg has an equal length, but this would need to be tested.
2. The step_time for the turnaroudn is hard coded to 0.1seconds. It may not need to be this fast but I don't think there's a harm in keeping it this fast. Because I have to control the ACU very directly through the turnaround there is probably a maximum step time to keep the ACU following the three leg turnaround accurately but I haven't tested it.
3. I've implemented this with a "TrackPoint queue" to ensure the batch_size is always obeyed. Because the three_leg_turnaround function returns a list of TrackPoints at the beginning of the turnaround I had to either fully ignore the batch_size when turnarounds happened or implement this queue. Matthew has said we don't have to obey the batch_size so we can ditch the queue for a slightly simpler implementation if desired.

This implementation basically leaves 95% of the `constant_scan_velocity` generator the same and only changes the turnaround portions. I've added a simple boolean into the function call that activates the new turnaround. If it is set to False then the constant velocity scan should generate the same trajectories as before.

## Motivation and Context

This implements the new three leg turnarounds for the SATs that have shown improvements to the FPA temps during turnarounds.

## How Has This Been Tested?

These turnarounds have been tested on SATp1 multiple times. We should probably test that this specific implementation of the `constant_velocity_scan` generator function works!

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/drivers.py
+++ b/socs/agents/acu/drivers.py
@@ -6,6 +6,7 @@ import time
 from dataclasses import dataclass, replace
 
 import numpy as np
+import three_leg_turnaround as three_leg_tr
 
 #: The number of seconds in a day.
 DAY = 86400
@@ -295,7 +296,8 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
                                     batch_size=500,
                                     az_start='mid_inc',
                                     az_first_pos=None,
-                                    az_drift=None):
+                                    az_drift=None,
+                                    turnaround_method=None):
     """Python generator to produce times, azimuth and elevation positions,
     azimuth and elevation velocities, azimuth and elevation flags for
     arbitrarily long constant-velocity azimuth scans.
@@ -337,8 +339,12 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
             start at this position (but otherwise proceed in the same
             starting direction).
         az_drift (float): The rate (deg / s) at which to shift the
-            scan endpoints in time.  This can be used to better track
+            scan endpoints i n time.  This can be used to better track
             celestial sources in targeted scans.
+        turnaround_method (str): The method used for generating turnaround.
+            Default (None) generates the baseline minimal jerk trajectory.
+            'three_leg' generates a three-leg turnaround which attempts to
+            minimize the acceleration at the midpoint of the turnaround.
 
     Yields:
         points (list): a list of TrackPoint objects.  Raises
@@ -413,10 +419,15 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
     point_group_batch = 0
 
     i = 0
+    point_queue = []
     while i < stop_iter and check_num_scans():
         i += 1
         point_block = []
         for j in range(batch_size):
+            if len(point_queue):  # Pull from points in the queue first
+                point_block.append(point_queue.pop(0))
+                continue
+
             point_block.append(TrackPoint(
                 timestamp=t + t0,
                 az=az, el=el, az_vel=az_vel, el_vel=el_vel,
@@ -436,6 +447,14 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
                     el_flag = 0
                 elif az == target_az:
                     # Turn around.
+                    if turnaround_method is not None and turnaround_method == "three_leg":
+                        turnaround_track = three_leg_tr.gen_three_leg_turnaround(t0=t + t0, az0=az, el0=el, v0=az_vel,
+                                                                                 turntime=turntime,
+                                                                                 az_flag=az_flag, el_flag=el_flag,
+                                                                                 point_group_batch=point_group_batch)
+                        for track_point in turnaround_track:
+                            point_queue.append(track_point)  # Add the TrackPoints from the turnaround into the queue.
+
                     t += turntime
                     az_vel = -1 * az_speed
                     el_vel = el_speed
@@ -463,6 +482,14 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
                     el_flag = 0
                 elif az == target_az:
                     # Turn around.
+                    if turnaround_method is not None and turnaround_method == "three_leg":
+                        turnaround_track = three_leg_tr.gen_three_leg_turnaround(t0=t + t0, az0=az, el0=el, v0=az_vel,
+                                                                                 turntime=turntime,
+                                                                                 az_flag=az_flag, el_flag=el_flag,
+                                                                                 point_group_batch=point_group_batch)
+                        for track_point in turnaround_track:
+                            point_queue.append(track_point)  # Add the TrackPoints from the turnaround into the queue.
+
                     t += turntime
                     az_vel = az_speed
                     el_vel = el_speed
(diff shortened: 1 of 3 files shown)


socs/agents/acu/three_leg_turnaround.py
```diff
@@ -0,0 +1,310 @@
+import numpy as np
+
+from socs.agents.acu.drivers import TrackPoint
+
+
+def gen_three_leg_turnaround(t0, az0, el0, v0, turntime, az_flag, el_flag, point_group_batch, **kwargs):
```
mhasself, line 6: Why are kwargs not declared explicitly here?

socs/agents/acu/three_leg_turnaround.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+
+    # Solve for the polynomial components that fits our initial and final conditions
+    A = solve_fifth_polynomial_lin_eqs(t_i, t_f, x0_i, x0_f, x1_i, x1_f, x2_i, x2_f)
+
+    ts = np.arange(t_i, t_f, step_time)  # Divide our times into points with step_time spacing
+    vs = x0(ts, *A)  # Solve for the velocity at every time using our solved components
+
+    an1 = get_an1(t_i, xn1_i, *A)  # We have to solve for the first antiderivative component to get the positions
+    xs = xn1(ts, an1, *A)  # Now solve for the positions at every time
```
mhasself, line 158: I think these two can be replaced by
```
xs = xn1(ts, 0, *A)
xs = xs - xs[0] + xn1_i
```
Then you don't need the `get_an1` function.

socs/agents/acu/three_leg_turnaround.py
```diff
@@ -0,0 +1,310 @@
...
+        an1-a5 (floats): The solved parameters of the order five polynomial from `solve_fifth_polynomial_lin_eqs` and `get_an1`.
+
+    Returns:
+        xn1(t) (float): The solution of the first anti-derivative of the order five polynomial equation calculated at time t.
+    """
+
+    return an1 + a0 * t + (1 / 2) * a1 * t**2 + (1 / 3) * a2 * t**3 + (1 / 4) * a3 * t**4 + (1 / 5) * a4 * t**5 + (1 / 6) * a5 * t**6
+
+
+def x0(t, a0, a1, a2, a3, a4, a5):
```
mhasself, line 253: I believe the functions below this point aren't used?  Remove.

mhasself (changes requested):
Couple specific changes requested.

We also need this functionality controllable from the generate_scan Process --  we can't just turn this on, by default, for everyone, forever.

I'd have thought this needed to be run in "free form" mode... was that not the case?

socs/agents/acu/drivers.py
```diff
@@ -278,7 +280,8 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
                                     batch_size=500,
                                     az_start='mid_inc',
                                     az_first_pos=None,
-                                    az_drift=None):
+                                    az_drift=None,
+                                    three_leg_turnaround=True):
```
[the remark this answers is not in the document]
mjrand: Makes sense to me. I've changed the argument to `turnaround_method=None` and added in the following if statement where the turnarounds are generated:
```if turnaround_method is not None and turnaround_method == "three_leg":```

This will preserve the default behavior unless specified by `turnaround_method`.

File: socs/agents/acu/three_leg_turnaround.py [hunk 5]
```diff
@@ -0,0 +1,310 @@
...
+        t0 (float): The initial time of the turnaround.
+        az0 (float): The iniital azimuth position of the turnaround. Should be equal to the final azimuth position of the turnaround.
+        el0 (float): The initial elevation of the turnaround. El velocity is forced to 0 here so this is only used for creating TrackPoints.
+        v0 (float): The initial azimuth velocity of the turnaround.
+        turntime (float): The turnaround time given by the above equation.
+        az_flag (int): The az flag used by the ACU. Inherited from the scan generation function and not changed. Used for TrackPoints.
+        el_flag (int): The el flag used by the ACU. Inherited from the scan generation function and not changed. Used for TrackPoints.
+        point_group_batch (int): the point group batch used by the ACU. Inherited from the scan generation function and not changed. Used for TrackPoints.
+
+    kwargs:
```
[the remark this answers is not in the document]
mjrand: Is `**kwargs` usually not the correct way to use them? I've just assumed them to be these sets of optional arguments but maybe there's a better way to do this stylistically. (I could just add these three args into the function header with default values if that seems better)

socs/agents/acu/three_leg_turnaround.py
mjrand: I've changed them to be explicit since there's only three of them.

socs/agents/acu/three_leg_turnaround.py
```diff
@@ -0,0 +1,310 @@
...
+        midpoint_time (float): The time used by the second leg of the turnaround. Defaults to 1 second.
```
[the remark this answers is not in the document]
mjrand: The second leg of the turnaround here isn't necessarily "stationary" of the velocity is non-zero... how about "second_leg_time"/"second_leg_velocity"?

socs/agents/acu/three_leg_turnaround.py
mjrand: Oh yeah that does work... and its so much more elegant! Awesome, thanks!

mjrand:
> Couple specific changes requested.

Addressed all of the comments!

> We also need this functionality controllable from the generate_scan Process -- we can't just turn this on, by default, for everyone, forever.

I've added in changes to the agent as well!

> I'd have thought this needed to be run in "free form" mode... was that not the case?

I've never dealt with "free form" mode... I just ran it using `from_file` and pre-generated scripts. Should I be using free form mode? I actually don't know to be honest.

mjrand:
Closing this and opening a new PR with a new branch here : #931

The commit is identical to this commit but the merge that added in type2 and type3 scans into the acu driver diverged my branch in a way that I didn't want to learn how to fix... so I just made a new branch based off of main and recommitted.

PR #938, opened 2025-10-24
Type23 cmdturn

mhasself:
This adds turnaround method as arg to type3 scans... only partial implementation -- this will make "standard" turn-arounds not work anymore. Which is fine, just need to be clearer about that.

I have tested the function in drivers.py fully, but the agent altogether.


mjrand (approved):
Looks good!

PR #939, opened 2025-10-24
Galil Axis Stage Controller Agent for SO SAT Coupling Optics 

sanahabhimani:
## Description
This PR adds a Galil DMC Axis Controller Agent, which provides motion control and readout for the Galil DMC motor controller. This agent was intentionally written for the hardware incorporated into the Simons Observatory SAT Coupling Optics system. When implemented in this way, the agent controls four axes—two linear and two angular—that move the hardware as needed to perform detector passband measurements.

Core updates involved:
- Implementation of driver functions for motion, position, torque, gearing control and more
- Added retry logic for Galil “?” responses 
   - we occasionally received “?” responses from the motor controller in an unpredictable manner. While this issue occurred less frequently on the SAT Coupling Optics motor controller, most of the agent development was done using a spare Galil controller, where “?” responses to queries were much more common—so a retry mechanism was implemented to handle those cases.
- Added configuration file support for axis conversions (counts/mm, counts/deg)

## Motivation and Context
Introduces a dedicated OCS agent for controlling the Galil DMC stage hardware used in SAT coupling optics measurements.

## How Has This Been Tested?
Has been tested on physical hardware in and outside of a docker container. 

**Note on Integration Tests:** Integration tests are still in development, and I’d love some help troubleshooting. At first, I ran into some Mac-related issues—but from past DAQ work, I know the Lakeshore240 and Radiometer tests have run fine on my local setup. When I started writing integration tests for this agent, even those familiar tests were failing, but after stabilizing my environment, I confirmed the issue is specific to this agent. So far, have only written one test --for the `init` task---and trying to debug before adding on more tests. I suspect the `init` test failure _may_ be related to how the config file is being used, since the agent relies on it to set parameters and select which axes to publish, though I’m not entirely sure yet. There may also be additional features I need to mock up for the integration tests, which I’d like to brainstorm with the DAQ team. I haven’t pushed the test files yet since I’m still debugging locally, but can push them shortly. 

## Types of changes

## Checklist:

## Miscellaneous Notes from Sanah
- This agent was developed based on existing code that had been handed down from an engineer at UChicago, who uses the same Galil control framework across multiple labs in the department. It wasn’t until later that I learned about the `gclib` Python package for the Galil motor. I suspect it wasn’t used originally because the engineer primarily operates from a Mac when assisting labs, and `gclib` isn’t macOS-friendly.
- However, I did take a brief detour to explore the use of `gclib`, since it’s used by the HWP `socs` agents. I set up a Windows VM on my Mac to test whether `gclib` could help address the “?” response issue, but it didn’t provide additional robustness or features that justified rewriting the driver I had already implemented.
- You’ll also note that I integrated the `TCPInterface` class into the device driver
- Some naming conventions could be improved. These motors will be used exclusively for the SAT Coupling Optics, so I’m unsure what the best instance ID naming convention should be when deploying to the site. There’s also some inconsistency between the directory name (`galildmc_axis_controller`) and the agent class name (`GalilAxisController`). But I'm also overall unsure if we want to name this more specifically for the SAT Coupling Optics (I doubt that given how we like to name agents after the specific piece of hardware we're commanding...)

### DIFF
--- a/socs/agents/galil_axis/agent.py
+++ b/socs/agents/galil_axis/agent.py
@@ -91,7 +91,7 @@ class GalilAxisAgent:
                                  agg_params=agg_params,
                                  buffer_time=1)
 
-    @ocs_agent.param('input_config', type=bool, default=None)
+    @ocs_agent.param('input_config', type=bool, default=False)
     @ocs_agent.param('auto_acquire', type=bool, default=False)
     def init(self, session, params=None):
         """init(input_config=False, auto_acquire=False)


sanahabhimani:
Hardware Testing Summary (as of today):

All tasks that can be verified using our spare motor controller were tested today. This is the controller that has been used for ongoing development after the SAT Coupling Optics motor controllers were shipped to Chile a few weeks ago.

The shipped SAT controllers include additional hardware features: limit switches, brakes, and sinusoidal amplifiers that require initialization before servoing motion. The spare controller used for development does not include these features.

That said, the driver code and OCS agent tasks related to:
- releasing and engaging brakes,
- querying brake status,
- setting limit switch polarity, and
- disabling limit switches

were tested on the SAT Coupling Galil motor controllers prior to shipment. Though I don’t expect any changes, since some time has passed, it is worth re-verifying these tasks first when the controllers are back online and ready for OCS communication at the site.

For completeness, the following tasks could not be tested today (to avoid risking damage to the spare axes or connected components):
- `get_brake_status`
- `set_axis_brake`
- `set_magnetic_cycle`
- `initialize_axis`
- `enable_sin_commutation`
- `disable_limit_switch`

Can also confirm:
- Agent reconnects after unplugging and re-plugging hardware's Ethernet connection

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+            try:
+                self.stage.get_data(self.axes)
+            except ConnectionError:
+                self.log.error("Could not establish connection to galil axis motor controller")
+                return False, "Galil Axis Controller agent initialization failed"
+
+        self.initialized = True
+
+        # load configfile if provided as init argument
+        if params.get('configfile'):
```
davidvng: I think you need to differentiate this param from the `configfile` that's parsed earlier in this function. One way to do this could be to instead make this a boolean param such as `input_config` that's set in `__init__` and then if `self.input_config` is True, run the following function.

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+                # enable sin commutation
+                initstate = gal['initaxisparams']['BA']
+                if initstate == 'True':
+                    # enable sin commutation
+                    self.stage.enable_sin_commutation(axis=a)
+                    # set magnetic cycle
+                    mag = gal['initaxisparams']['BM']
+                    self.stage.set_magnetic_cycle(axis=a, val=mag)
+                    # set dwell times before initializing
+                    self.stage.set_dwell_times(t_first=self.first_dwell, t_second=self.sec_dwell)
```
davidvng: `self.first_dwell` and `self.sec_dwell` here are defined in `init`. If this function were to be run as a standalone at a later time, and the config file itself has changed, then these changed values would still be the old values, not the new ones.

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+                self.log.warn(f"Could not start Task because "
+                              f"{self.lock.job} is already running")
+                return False, "Could not acquire lock"
+
+            cfg = read_config(configfile)
+            try:
+                gal = cfg.get('galil', {})
+                self.cfg = cfg
+                self.motorsettings = gal.get('motorsettings', {})
+                self.axis_map = gal.get('motorconfigparams', {})
```
davidvng: If these keys are checked in `read_config` then you can just directly reference them instead of using `.get()`.

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+        raise ValueError("Config missing top-level 'galil' section.")
+
+    mparams = gal.get('motorconfigparams')
+    if not isinstance(mparams, dict) or not mparams:
+        raise ValueError("galil.motorconfigparams must be a non-empty dict of axes.")
+
+    return cfg
+
+
+class GalilAxisControllerAgent:
```
davidvng:
```suggestion
class GalilAxisAgent:
```

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+    elif args.mode == 'acq':
+        init_params = {'auto_acquire': True}
+
+    # Call launcher function (initiates connection to appropriate
+    # WAMP hub and realm).
+
+    agent, runner = ocs_agent.init_site_agent(args)
+
+    # create agent instance and run log creation
+    galilaxis_agent = GalilAxisControllerAgent(agent, args.ip, args.configfile)
```
davidvng:
```suggestion
    galilaxis_agent = GalilAxisAgent(agent, args.ip, args.configfile)
```

docs/agents/galilaxis_controller.rst
davidvng: Change filename to `galil_axis.rst`

docs/agents/galilaxis_controller.rst
```diff
@@ -0,0 +1,109 @@
...
+=====================
+
+The Galil DMC Axis Controller Agent provides motion control and telemetry readout for
+the Galil DMC motor controller. When used in the Simons Observatory SAT Coupling
+Optics system, the agent controls four axes—two linear and two angular—that move
+the hardware as needed to perform detector passband measurements.
+
+
+.. argparse::
+    :filename: ../socs/agents/galildmc_axis_controller/agent.py
```
davidvng:
```suggestion
    :filename: ../socs/agents/galil_axis/agent.py
```

docs/index.rst
```diff
@@ -43,6 +43,7 @@ API Reference        Full API documentation for core parts of the SOCS library.
     agents/devantech_dS378
     agents/fts_agent
     agents/generator
+    agents/galilaxis_controller
```
davidvng:
```suggestion
    agents/galil_axis
```

socs/agents/galildmc_axis_controller/drivers.py
```diff
@@ -0,0 +1,504 @@
...
+        resp = self.galil_command(command="SB", value=output_num)
+        return resp
+
+    def engage_brake(self, output_num):
+        """
+        Engage the brake for the specified axis using the Galil CB command.
+
+        """
+        resp = self.galil_command(command="CB", value=output_num)
+        return resp
```
davidvng: Same as for `release_brake`.

socs/agents/galildmc_axis_controller/drivers.py
```diff
@@ -0,0 +1,504 @@
...
+
+    def get_brake_status(self, axis, output_map):
+        """
+        Return brake status for axis via @OUT[n].
+
+        """
+        output_num = output_map[axis]
+        query_str = f"@OUT[{output_num}]"
+        val = self.galil_command(command="MG", value=query_str)
```
davidvng:
```suggestion
        val = self.galil_command(command="MG", value=query_str, expect_response=True)
```

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+        """
+        axis = params['axis']
+        speed = params['speed']
+        with self.lock.acquire_timeout(timeout=5, job='jog_axis') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start Task because "
+                              f"{self.lock.job} is already running")
+                return False, "Could not acquire lock"
+
+            self.stage.jog_axis(axis, speed)
```
davidvng:
```suggestion
            self.stage.set_jog_speed(axis, speed)
```

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+    agent.register_task('get_torque_limit', galilaxis_agent.get_torque_limit)
+    agent.register_task('set_amp_currentloop_gain', galilaxis_agent.set_amp_currentloop_gain)
+    agent.register_task('get_amp_currentloop_gain', galilaxis_agent.get_amp_currentloop_gain)
+    agent.register_task('set_motor_state', galilaxis_agent.set_motor_state)
+    agent.register_task('get_motor_state', galilaxis_agent.get_motor_state)
+    agent.register_task('set_magnetic_cycle', galilaxis_agent.set_magnetic_cycle)
+    agent.register_task('set_dwell_times', galilaxis_agent.set_dwell_times)
+    agent.register_task('initialize_axis', galilaxis_agent.initialize_axis)
+    agent.register_task('define_position', galilaxis_agent.define_position)
+    agent.register_task('jog_axis', galilaxis_agent.jog_axis)
```
davidvng:
```suggestion
    agent.register_task('set_jog_speed', galilaxis_agent.set_jog_speed)
```

socs/agents/galildmc_axis_controller/drivers.py
```diff
@@ -0,0 +1,504 @@
...
+        if state == 'enable':
+            self.galil_command("SH", axis=axis)
+        elif state == 'disable':
+            self.galil_command("MO", axis=axis)
+
+        # --- Allow controller to settle ---
+        time.sleep(1.0)
+
+        # --- Query new state ---
+        status, human_state = self.get_motor_state(axis=axis)
```
davidvng: In most/all of the other "set" functions, you don't call its related "get" function within it. I comment on this in my overall review comment, but it's best to standardize to one or the other.

docs/agents/galilaxis_controller.rst
davidvng: Also add a section to give some examples of using tasks. You don't have to provide how to use every single task, but I'd say at least one "get" task and one "set" task.

socs/agents/galildmc_axis_controller/drivers.py
```diff
@@ -0,0 +1,504 @@
...
+        except (TypeError, ValueError):
+            print(f"Could not parse brake value '{val}' for axis {axis}.")
+            return {axis: {"state": None, "status": "Unknown"}}
+
+        if state == 1:
+            status = "Brake Released"
+        if state == 0:
+            status = "Brake Engaged"
+
+        return {axis: {"state": state, "status": status}}
```
davidvng: Since this is always called for a single axis, just return `state` and `status` (you could also just return `status` only)
```suggestion
        return state, status
```

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+                self.log.warn(f"Could not start Task because "
+                              f"{self.lock.job} is already running")
+                return False, "Could not acquire lock"
+
+            self.stage.stop_motion(axis)
+
+        return True, f'Stopped motion for {axis}.'
+
+    @ocs_agent.param('order', type=str)
+    def set_gearing(self, session, params):
```
davidvng: Rename this (and its related driver function) to give more context and differentiate from `set_gearing_ratio`.

davidvng (changes requested):
Thanks for the new agent! Overall looks good.

An overall agent naming suggestion: I'm leaning towards just "GalilAxisAgent" dropping the controller part since the agent does a lot of both querying and controlling. Also drop the "dmc" part wherever that's still around. The agent directory should then just be `galil_axis`.

In general, you have "set" functions/tasks to assume that the command has successfully ran. In general, I think you should at least change any "[thing] has been set to [value]" statement to "Commanded [thing] to be set to [value]" to not assume the action was successful. Then, since every "set" command doesn't capture a response from the device, any OCS client should be expected to call its related "get" function afterwards to check that the command ran successfully. This should be documented in the agent docs.

socs/agents/galildmc_axis_controller/agent.py
sanahabhimani: oops, thanks for pointing out the typos!

socs/agents/galildmc_axis_controller/drivers.py
```diff
@@ -0,0 +1,504 @@
...
+        return self.galil_command("PA", axis=axis, value=counts)
+
+    def release_brake(self, output_num):
+        """
+        Release brake to axis by using the GalilDMC SB command which sets the digital
+        output to 1.
+
+        """
+        resp = self.galil_command(command="SB", value=output_num)
+        return resp
```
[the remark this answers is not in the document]
sanahabhimani: Thanks for pointing this out. This task predates the more recent driver and agent updates, and since the spare motor controller used to finalize the agent doesn’t have brakes to test, the older brake-status handling code wasn’t revisited thoroughly.

The brake release/engage logic is consistent with other `set_x_setting` driver methods like `set_motor_type` or `set_off_on_error`. These brake commands don’t produce a response, so if the brake doesn’t engage or release as intended, verification comes from querying the brake status directly—similar to the `get_x_setting` driver code/task.

I’ll update the agent task to remove the try/except so it aligns more closely with the `set_x_setting` pattern.

socs/agents/galildmc_axis_controller/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+
+        with self.lock.acquire_timeout(timeout=5, job='get_brake_status') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start Task because {self.lock.job} is already running")
+                return False, "Could not acquire lock"
+
+            response = self.stage.get_brake_status(axis=axis, output_map=self.brakes)
+            session.log.info(f'Brake status for {axis} is: {response}')
+
+        return True, f'Brake status for {axis} is: {response}'
```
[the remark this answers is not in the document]
sanahabhimani: Keeping both the raw state and human-readable status in the log per Shreya’s request/peace of mind— the Galil’s use of 1 and 0 isn’t consistent across commands. For example (not necessarily in this exact mapping, just off the top of my head), 1 may indicate released for brakes but off for motors, so displaying both makes it easier for users familiar with the raw hardware output to cross-check what they see in the logs.

socs/agents/galil_axis/agent.py
```diff
@@ -0,0 +1,1170 @@
...
+            try:
+                self.stage.get_data(self.axes)
+            except ConnectionError:
+                self.log.error("Could not establish connection to galil axis motor controller")
+                return False, "Galil Axis Controller agent initialization failed"
+
+        self.initialized = True
+
+        # load configfile if provided as init argument
+        if params.get('configfile'):
```
sanahabhimani: This comment helped me find that `init_params` for `if args.mode == 'init'` only included `'auto_acquire': False` and didn't define a parameter for `configfile`. Automatically forcing `init` to run `input_configfile` during agent initialization wasn’t the intention, but if `configfile` had been added to `init_params`, that’s what would have happened, which makes sense why David was highlighting the need to differentiate.

Since `init` already reads the config file to determine which axes to query for position and velocity data--- regardless of whether it runs the `input_configfile` task--- I replaced the `configfile` param  with a boolean `input_config` param.

I added it to `init_params` for both `init` and `acq` so that if `input_config` is `True`, `init` will use `self.configfile` as the arg for the `input_configfile` task. 

Will also update the docs accordingly.

sanahabhimani:
Alright, as per my usual MO, I have evaded writing tests but am ready for re-review. Holler if tests are still preferred tho---happy to delineate into a separate PR for completeness' sake.

docs/agents/galil_axis.rst
```diff
@@ -0,0 +1,257 @@
+.. highlight:: rst
+
+.. _galilaxis:
```
davidvng:
```suggestion
.. _galil_axis:
```

davidvng (approved):
Great thanks! Feel free to merge whenever.

PR #940, opened 2025-10-29
Fix to IRIG timing bug

bbixler500:
Fixes a bug in which the IRIG second value is wrong every other value by 80 seconds

## Description

There is a specific decoding function used to read the individual bits in the raw IRIG data. For the "seconds" value specifically, there is a bit shift which causes the decode function to look at a bit which is never explicitly set to a value. This bit can sometimes get stuck as a 1 when it is normally a 0, causing the second reading to be off by 80. The fix adds logic so that this bit should not contribute to the returned value in the "seconds" case only.

## Motivation and Context

When this issue occurs the HWP timing data is unusable without a correction. For some reason SATp2 has this bug occur fairly frequently, but it has been observed on the other SATs as well

## How Has This Been Tested?

I implemented this change on one of the two encoders for SATp2 and then waited for the bug to happen. On the encoder with this update the IRIG time remained at it's expected value.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_encoder/agent.py
+++ b/socs/agents/hwp_encoder/agent.py
@@ -96,12 +96,12 @@ def de_irig(val, base_shift=0):
     val : int
        raw IRIG bit info of each 100msec chunk
     base_shift : int, optional
-       number of bit shifts. This should be 0 except for seccods
+       number of bit shifts. This should be 0 except for seconds
 
     Returns
     -------
     int
-       Either of sec/min/hourds/day/year
+       Either of sec/min/hours/day/year
 
     """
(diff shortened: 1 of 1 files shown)


ykyohei (approved):
I confirmed that raw data look good.
Encoder 1 was not debugged and irig_second has offset by 80 seconds alternately, leading to invalid irig time = -1 and huge offset in irig_minus_sys.
Encoder 2 with debug has no issue in this period, irig_minus_sys is ~ -1.

This is not done yet. but it is also possible to apply corrections to existing observations that have invalid irig time with 80 seconds offset.

251029satp2irig
251029satp2irig_mag

PR #942, opened 2025-10-31
fix: fixes to prevent errors when running 3-leg turnarounds with type…

skhrg:
… 2 and 3 scans

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -2490,7 +2490,7 @@ class ACUAgent:
                         if got_progtrack:
                             self.log.warn('Unexpected exit from ProgramTrack mode!')
                             if mode == 'stop':
-                                unabort_failure = True  #close enough!
+                                unabort_failure = True  # close enough!
                             mode = 'abort'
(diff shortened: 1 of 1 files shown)


socs/agents/acu/agent.py
```diff
@@ -2157,7 +2157,11 @@ def generate_scan(self, session, params):
         if el_freq is None:
             el_freq = self.scan_params['el_freq']
         if turnaround_method is None:
-            turnaround_method = self.scan_params['turnaround_method']
+            if params['scan_type'] in [2, 3]:
```
mjrand: I'll remove this when I add the standard turnaround PointTrack generator

mjrand (approved):
Looks good, I'll add small changes on the next branch

mjrand:
This commit has some strange interaction between Matthew's and Saianeesh's fixes that caused the scan to fail from a ```RuntimeError: Operation generate_scan in Agent acu is not in the 'running' state.```

Saianeesh will open a new PR with his fix to this commit that let us scan again.

PR #954, opened 2025-11-24
fts: Update hard coded limits

ShreyaSut:
## Description
Needed to expand the hard limits for the FTS agent because the agent was previously for the ACT FTS. The current FTS is the SO FTS, which has a larger stage allowing for a wider range of motion. It uses an Aerotech controller still, allowing us to use the previous agent modified with the expanded limits. This SO FTS will stay at the site and will be used for SAT and LAT bandpass measurements.

## Motivation and Context
The previous hard limits were preventing us from testing the SO FTS mirror stage at its limits of travel. Changing the limits from +/- 74.8 mm to +/-80 mm allowed us to command and test the FTS mirror stage at its limits of travel.
## How Has This Been Tested?
Tested on the hardware at site.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/fts_aerotech/agent.py
+++ b/socs/agents/fts_aerotech/agent.py
@@ -227,7 +227,7 @@ class FTSAerotechAgent:
         **Task** - Move to absolute position relative to stage center (in mm).
 
         Parameters:
-            position (float): Position in mm, must be between -74.8 and 74.8.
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
Thanks for answering those questions, and for the docstring change!

ShreyaSut:
Thank you, Brian! And apologies for the autocorrect misspelling in the last email..
________________________________

PR #960, opened 2025-12-15
Galil3 - Added limit switch query task to agent.py and drivers.py

ShreyaSut:
Added two new tasks to query the forward and reverse limits to see if they're triggered or not.

## Description
Added two new tasks to query the forward and reverse limits to see if they're triggered or not. The argument can be any axis. (e.g. Axis "A" for motor A).

## Motivation and Context
We ran into a new issue at site with the limit switches bouncing where they were flickering. We added capacitors in the electronics to try and debounce the limit switches, but were not able to check the state of the limits without the new tasks. Now we can query those forward and reverse limits to see if they're triggered or not.

## How Has This Been Tested?
Tested the limits for the lead axes "A" and "C" in a script and they worked.

## Types of changes

## Checklist:

I think the documentation would need to be updated if a new task is added, but I'm unsure, but I can add that later if that's needed.

### DIFF
--- a/socs/agents/galil_axis/drivers.py
+++ b/socs/agents/galil_axis/drivers.py
@@ -488,7 +488,7 @@ class GalilAxis(TCPInterface):
 
         # --- Interpret raw
         human_state = "not triggered" if state == 1 else "triggered" if state == 0 else f"unknown ({state})"
-        
+
         return state, human_state
 
     def get_reverse_limitswitch(self, axis):
--- a/socs/agents/galil_axis/agent.py
+++ b/socs/agents/galil_axis/agent.py
@@ -1102,15 +1102,15 @@ class GalilAxisAgent:
                 return False, "Could not acquire lock."
 
             state, human_state = self.stage.get_forward_limitswitch(axis)
-        
+
         # store data
         session.data = {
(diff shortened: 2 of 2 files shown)


socs/agents/galil_axis/drivers.py
```diff
@@ -475,6 +475,38 @@ def get_limitswitch_polarity(self):
...
+        """
+        resp = self.galil_command("MG _LR", axis=axis, expect_response=True)
+        try:
+            state = int(float(resp))
+        except Exception:
+            print(f"Unexpected response from MG _LR{axis}: {resp}")
+            return None, "unknown"
+
+        # --- Interpret raw
+        human_state = "not triggered" if state == 1 else "triggered" if state == 0 else f"unknown ({state})"
```
davidvng:
```suggestion
        human_state = "not triggered" if state == 1 else ("triggered" if state == 0 else f"unknown ({state})")
```

davidvng (changes requested):
Thanks for the updates! Just some small changes, including a typo that should fix the failing check.

BrianJKoopman (changes requested):
A couple more small comments from me.

PR #961, opened 2025-12-15
New Agent: tauHK

merny93:
This PR aims to integrate the tauHK cryogenic housekeeping instrumentation into the ocs framework for use on the LAT.

## Description

The tauHK instrument interfaces with a host computer over ethernet and communicates with a daemon process. The code in this PR is a thin wrapper on top of the daemon.

Data and commands are sent over a UDP socket serialized using protobuf. A one time code-gen step is required to generate all the python protobuf definitions such that the agent can communicate with the daemon. Resulting protobuf definitions files are kept in the `pb2/` directory but not tracked in git.

This agent spawns the daemon process using the `subprocess` module and pushes logs to an ocs feed. The compiled binary is supplied but not tracked in git. 

## Motivation and Context

tauHK will be installed on site during the ASO upgrades. This code will be deployed on `amun` .

## How Has This Been Tested?
This agent code has been tested on a development machine at Princeton.  The hardware and software were tested end-to-end along with ocs-web integration.

## Types of changes

## Checklist:


docs/agents/tauhk.rst
BrianJKoopman: This file is largely incomplete. There's a complete example in [the ocs docs](https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors). You can also see any other agent's reference page for an example. Important things missing:
* The `argparse` block, which automatically displays arguments for the agent
* A description of dependencies for running this agent (I know we've discussed this offline, but just for the record here). This should describe, or otherwise link to, instructions for installing the required dependencies.
* Configuration file examples
* "Agent API" section

Could you also add it to the [index](https://github.com/simonsobs/socs/blob/main/docs/index.rst) (in alphabetical order)?

socs/agents/tauhk/agent.py
```diff
@@ -0,0 +1,578 @@
...
+                                value_type = type(value)
+                                data_averaged[measurement] = value_type(value / spf)
+
+                        average_message = {'block_name': 'tauhk_data_averaged', 'timestamp': message_timestamp, 'data': data_averaged}
+                        self.agent.publish_to_feed('tauhk_data_influx', average_message)
+                        self.averaged_data = dict()
+                except Exception as e:
+                    # raise e
+                    # how best to handle an error here?
+                    self.log.error(f"Error receiving data: {e}")
```
BrianJKoopman: Good to see some error handling here. How best to handle it is going to depend on the error. Typically we want to try to re-establish a connection with the device (if it was lost) and `continue` to the next loop iteration. It's good to also mark the `session` as 'degraded' via: `session.degraded = True`. (You need to unmark this when that becomes true too.) More on agent robustness is linked here: https://github.com/simonsobs/socs/issues/721

You might consider also logging a traceback so that we can track down bugs a bit easier:
```
self.log.error("{e}", e=traceback.format_exc())
```

(With an `import traceback` up top.)

socs/agents/tauhk/agent.py
```diff
@@ -0,0 +1,578 @@
...
+
+        Stops the receive_data process.
+        """
+        if self._take_data:
+            self._take_data = False
+            return True, 'Stopping data acquisition request.'
+        return True, 'Data acquisition was not running.'
+
+    def start_crate(self, session, params):
+        """Connect to tauHK crate
```
BrianJKoopman: This first line should be the method signature override:

```suggestion
        """start_crate()
```

socs/agents/tauhk/agent.py
```diff
@@ -0,0 +1,578 @@
...
+        self.process = subprocess.Popen(
+            ['./tauhk-agent'],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            env={
+                "RUST_LOG": "info",
+                "TAUHK_IP_CMD_SEND": "10.1.0.74",
+                "TAUHK_IP_CMD_RECV": "10.1.0.77",
+                "TAUHK_IP_DATA": "10.1.0.77"
```
BrianJKoopman: Similar to above, these IPs shouldn't be hardcoded here, but configurable.

socs/agents/tauhk/agent.py
```diff
@@ -0,0 +1,578 @@
...
+        """generic_send(channel=str, option=str, value=str)
+        **Task** - Send a command for a specific channel and option to tauHK.
+
+        Args:
+            channel (str): The name of the channel to which the command is sent.
+            option (str): The name of the option within the channel to which the command is
+            value (str): The value to set for the specified channel and option.
+
+        Notes:
+            Channels defintions are in the config file for tauHK.
```
BrianJKoopman:
```suggestion
            Channels definitions are in the config file for tauHK.
```

socs/agents/tauhk/agent.py
```diff
@@ -0,0 +1,578 @@
...
+                        advertised[field][option] = ["int", (0, 2**32 - 1)]
+                else:
+                    # Currently unimplemented
+                    self.log.error(f"Unknown type {data_type} for {field}.{option}")
+        session.data = {"settables": advertised}
+        return True, 'Advertised tauHK commands.'
+
+    @ocs_agent.param('config_file', type=str)
+    def load_config(self, session, params):
+        """load_config()
```
BrianJKoopman: Show the argument in the method signature:
```suggestion
        """load_config(config_file)
```

BrianJKoopman (changes requested):
Sorry this took a while to get to, but here's a first pass at a review. It'd be great to have the requested documentation for the dependencies.

socs/agents/tauhk/agent.py
merny93: These IPs are hardcoded into the firmware running on the hardware so changing them on the receiving side would require rebuilding and reflashing the hardware.

docs/agents/tauhk.rst
BrianJKoopman: Ah, sorry about that, updated in the original post. It's here: https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html

PR #963, opened 2025-12-17
hwp-supervisor: Add missing shutdown procedure

bbixler500:
This PR adds functionality to the hwp-supervisor agent to shut off bias power to the driver board during a triggered shutdown

## Description

Currently, when a HWP shutdown is triggered the only actions taken is to lock out user controls and shut off the drive power. These actions alone are not enough to fully stop the HWP as the driver board bias power needs to be disabled as well (we should be doing this anyways because the electronics shouldn't be biased if there is expected to be large swings in the cryostat temperature). This PR adds to the hwp-supervisor's monitor process additional logic which shuts off the bias power once a shutdown is triggered.

## Motivation and Context

See above

## How Has This Been Tested?

This was tested successfully on satp2 by simulating a communication failure to initiate the shutdown

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -1597,7 +1597,7 @@ class HWPSupervisor:
         """
         pm = Pacemaker(1. / self.sleep_time)
         test_mode = params.get('test_mode', False)
-        last_ok_time = time.time()
+        last_okay_time = time.time()
 
         session.data = {
             'timestamp': time.time(),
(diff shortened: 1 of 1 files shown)


socs/agents/hwp_supervisor/agent.py
```diff
@@ -1987,6 +2004,13 @@ def disable_driver_board(self, session, params):
         action.sleep_until_complete(session=session)
         return action.success, f"Completed with state: {action.cur_state_info.state}"
 
+    def cancel_shutdown(self, session, params):
+        """cancel_shutdown()
+        **Task** - Cancels shutdown mode
+        """
+        self.shutdown_mode = False
+        return True, "Cancelled shutdown mode"
+
```
BrianJKoopman: I take it this is for cancelling a shutdown before the timeout expires. It's a bit unclear in the current implementation. It looks like I could 'cancel' even after a 'stop' is called, though then the driver board will still be disabled.

You might consider adding a separate attribute to track if you've entered the `elif action == 'no_data' and self.shutdown_no_data_timeout >= 0:` check and only allow the `cancel_shutdown` task to be called successfully if we're waiting on that timeout.

BrianJKoopman: Ah, and so the user is expected to 'cancel' and then manually re-enable the driver board?

PR #964, opened 2025-12-18
ACU: Support el_mode='stop' in generate_scan

mhasself:
## Description

Main change is that the azonly argument in generate_scan is replaced by el_mode.  Default behavior is the same, which is to leave elevation axis in Preset mode for the (type 1 or type 2) scan.  But now you have the option to request el_mode="stop", which will put that axis in Stop for the scan.

The el_mode is also a "scan_param" whose default value can be set in config file and then updated using `set_scan_params`.

Note although there's an interface change here, the azonly argument does not appear to be used by anyone currently (it's not exposed through ocs-web, even).  And the functionality it advertised was not correctly implemented.

## Motivation and Context

This is a desired interim behavior for SATP2 while some issues with elevation drive are being investigated.

## How Has This Been Tested?

Tested on ACU simulator software.  Confirmed old default behavior is reproduced.

## Types of changes

## Checklist:


mhasself:
This has also been tested at site; works as designed.

PR #968, opened 2026-01-07
acu: Add two_leg turnaround method

mjrand:
I've added in a new "two_leg" turnaround method and updated many of the turnaround docstrings.

## Description

I've added in the ability to explicitly call a "two_leg" turnaround which calls a "_gen_three_leg_turnaround" with `second_leg_time=0`. This was already being done for type2 and type3 scans but can now be called explicitly by the LAT and the SATs in `generate_constant_velocity_scan`. I've updated the type2 and type3 scan generation functions to use the two_leg turnaround by default (which is the exact same functionality as was being done before). The LAT type2/3 scans should be completely unaffected by these changes.

I've also gone through and updated many of the docstrings for these generate scan functions to include all turnaround method information.

## Motivation and Context

This change lets us use `second_leg_time=0` turnarounds in the constant velocity scans for the SATs. I've shown these to be advantageous (especially for satp2) over the standard turnarounds. 

## How Has This Been Tested?

I tested two_leg turnarounds briefly on satp2 and satp1, but this exact implementation of the code has not been run. We can run this for a few minutes just to check for any syntax or spelling errors if necessary.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/turnarounds.py
+++ b/socs/agents/acu/turnarounds.py
@@ -18,7 +18,7 @@ def gen_turnaround(turnaround_method, t0, az0, el0, v0, turntime, az_flag, el_fl
             2. "three_leg": Generates a special turnaround that splits the turnaround into three "legs" that attempt to
                 minimize the acceleration at the midpoint of the turnaround. See the _gen_three_leg_turnaround() docstring
                 for further documentation.
-            3. "two_leg": Generates a "three_leg" turnaround with second_leg_time = 0. 
+            3. "two_leg": Generates a "three_leg" turnaround with second_leg_time = 0.
         t0 (float): The initial time of the turnaround.
         az0 (float): The iniital azimuth position of the turnaround. Should be equal to the final azimuth position of the turnaround.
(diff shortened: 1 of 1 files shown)


socs/agents/acu/drivers.py
```diff
@@ -350,9 +351,12 @@ def generate_constant_velocity_scan(az_endpoint1, az_endpoint2, az_speed,
             scan endpoints in time.  This can be used to better track
             celestial sources in targeted scans.
         turnaround_method (str): The method used for generating turnaround.
-            Default ('standard') generates the baseline minimal jerk trajectory.
+            (Default) 'standard' generates the baseline minimal jerk trajectory.
+            'standard_gen' genetares a list of track_points throughout the turnaround
```
mhasself: 3 instances of "genetares" that should be "generates"

mjrand: Fixed!

mhasself (approved):
Ok, looks good!

PR #971, opened 2026-01-12
hwp-supervisor: Add shutdown delay flag

bbixler500:
## Description

Adds a configurable parameter for the delay between the supervisor shutdown and the pmx/gripper shutdowns

## Motivation and Context

As part of the supervisor shutdown, the bias power to the HWP driver board is turned off. When this happens the frequency readout of the HWP drops to zero, even if the HWP is still rotating. This might be confusing for someone not familiar with the HWP so we've decided that the supervisor shutdown should be delayed relative to the other HWP agent shutdowns

## How Has This Been Tested?

This was tested with SATp1 while it was partially warmed up for the HWP rotor recentering

## Types of changes

## Checklist:


ykyohei (approved):
This looks great. thanks for making this and testing this!

PR #1034, opened 2026-04-07
Updated default step_time in gen_turnaround

mjrand:
## Description

Updated the default step_time in `gen_turnaround` in `turnarounds.py` to `step_time=0.1`

## Motivation and Context

This really should have been the default before probably. I think the ACU maxes out at 0.1s step time.
I think I had set this higher when I was testing very high accelerations.
This might be the cause of the az position errors seen on satp1, but that's uncertain until tested.
The high frequency may over constrain the turnaround or cause very strange ACU behavior if we're sending too many points.

## How Has This Been Tested?

This has been tested before. 

## Types of changes

## Checklist:


BrianJKoopman:
Just a heads up -- I made some changes to the workflows/checks that run here today. You'll need to merge the latest `main` into this branch before proceeding.

PR #1037, opened 2026-04-08
Acu agent update line timestamp check

mjrand:
## Description

Added in a check for Point timestamps to add to update_lines in `_run_track` function in ACU agent.
This should added code should make sure that the update_lines list always contains at least ~3 * step_times worth of points.

## Motivation and Context

We were seeing that the ACU agent `_run_track` function was running out of points during `generate_scan` calls when using non-standard (two_leg, standard_gen, three_leg) turnaround methods. This is because the `turnaround_gen` function generates PointTrack points at a step_time (currently 0.05s) that is much lower than the `_run_track` step_time (1.0s). The `_run_track` function tries to pull ~6 points per loop (which should be ~3 seconds worth of points). Because the `turnaround_gen` uses a much lower step_time, the `_run_track` function was struggling to keep up with the number of points needed per second. This change should make sure each loop of `_run_track` should always grab at least 3*step_time worth of points.

## How Has This Been Tested?

Needs to be tested! We should test this on one of the platforms with the two_leg or standard_gen turnaround and watch the # free_position_points during turnarounds.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -2440,7 +2440,7 @@ class ACUAgent:
         # (The cost is that stopping a scan may take a little longer.)
         MIN_STACK_ADVANCE_TIME = 3.
 
-        # Minimum nuber of points to keep in the stack.
+        # Minimum number of points to keep in the stack.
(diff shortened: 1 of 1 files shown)


mjrand:
Closing this because #1047 is a better implementation of this code.

PR #1047, opened 2026-04-27
acu: Upload ProgramTrack points based on time

mjrand:
Rewrote the `_run_track` function to upload lines by time instead of free points in the ACU.

## Description

The `_run_track` function in the ACU agent has been refactored to upload ever `MIN_STACK_ADVANCE_TIME` (or the `last_upload_length`) worth of points. The filling of `upload_lines` is also written to always fill to `2 * MIN_STACK_ADVANCE_TIME` worth of points. 

## Motivation and Context

The current ACU _run_track function uploads points when the number of free points in the ACU runs below the minimum number of points based on the step_time of the generator. This causes the minimum number of uploaded points to be too small if the step_time ever increases during point generation (like during two_leg turnarounds). Here I've rewritten the uploading to be completed via time since last upload and to always upload at least 2 times the `MINIMUM_STACK_ADVANCE_TIME` worth of points. This should hopefully keep the minimum number of uploaded points at or above `MINIMUM_STACK_ADVANCE_TIME` worth of points at all times.

## How Has This Been Tested?

Needs to be tested on a real ACU.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -2563,7 +2563,8 @@ class ACUAgent:
                 # (Meaning we have less than the minimum time worth of points uploaded).
                 # Or if the total number of free positions is higher than the MAX_ALLOWABLE_FREE_POSITIONS.
                 # (Meaning we haven't uploaded at least the minimum number of points)
-                if ((last_uploaded_timestamp - time.time()) <= MIN_STACK_ADVANCE_TIME) or (free_positions > MAX_ALLOWABLE_FREE_POSITIONS):
+                if ((last_uploaded_timestamp - time.time()) <= MIN_STACK_ADVANCE_TIME) \
+                   or (free_positions > MAX_ALLOWABLE_FREE_POSITIONS):
 
                     upload_lines = []
                     # Grab points from point_prov until our last point is at least
@@ -2580,7 +2581,7 @@ class ACUAgent:
                     while not point_prov.is_empty() and len(upload_lines) and upload_lines[-1].group_flag != 0:
                         upload_lines.append(point_prov.pop())
 
-                    if point_prov.is_empty():
+                    if point_prov.is_empty() and mode == 'go':
                         mode = 'stop'
                         stop_message = 'Stop due to end of the planned track.'
 
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -2565,24 +2565,20 @@ class ACUAgent:
                 if mode == 'abort':
                     lines = []
 
-                last_upload_size = 0
+                upload_lines = []
                 # Is it time to upload more lines?
-                # STACK_REFILL_THRESHOLD is the absolute minimum number of points that can be in the stack
-                # based on the constant_velocity step_time.
-                # FULL_STACK - int(last_upload_size/2) is the minimum number of points that we should
-                # try to reupload based on the number of points in the last upload in case we uploaded
-                # points with a lower dt. We always add at least 2xMIN_STACK_ADVANCE_TIME worth of points,
-                # so this should attempt to always keep MIN_STACK_ADVANCE_TIME worth of points in the
-                # upload.
-                if free_positions >= min(STACK_REFILL_THRESHOLD, FULL_STACK - int(last_upload_size / 2)):
-                    new_line_target = max(int(free_positions - STACK_TARGET), 1)
+                # This happens when we initiate (len(upload_lines) == 0) or when our last uploaded points
+                # is less than MIN_STACK_ADVANCE_TIME worth of time away.
+                if len(upload_lines) == 0 or (time.time() + MIN_STACK_ADVANCE_TIME >= upload_lines[-1].timestamp()):
 
-                    # Make sure that you get one more line than you
-                    # need (len(lines) > new_line_target), so that
-                    # after "grabbing the minimum batch", below, there
-                    # is still >= 1 line left.  The lines-is-empty
+                    # Make sure that we have at least 1 MIN_STACK_ADVANCE_TIME
+                    # worth of lines more than we need so that 
+                    # after "grabbing 2 * MIN_STACK_ADVANCE_TIME worth of lines", below, there
+                    # is still >= 1 MIN_STACK_ADVANCE_TIME worth of lines left. The lines-is-empty
                     # check is used to decide we're done.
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Just a quick comment to point out that [per pre-commit](https://results.pre-commit.ci/run/github/186511668/1777331473.kR4hUf9AQrCxzC-yqACz3A), both `STACK_REFILL_THRESHOLD` and `STACK_TARGET` are unused and should be removed. Doing so will fix that failing check.

mjrand:
I tested this current commit on satp1 and got the following success with the standard turnaround
image

and this with the two_leg turnaround
image

I'm not entirely sure why the # of uploaded points are increasing at a linear rate of ~1 point / 3 seconds. There is some small interplay in the tracking of the current uploaded amount and the actual timing of the upload that gives it a 1 point per upload error. Its _probably_ fine except for egregiously long scans (like multiple multiple hours). But we could add a check to "relax" the points down to reasonable numbers if we ever hit like 500+ points of something. (Or find the problem I guess).

mjrand:
Ah! I figured out the bug that was causing me to increase the uploaded points by 1 every upload!

I was misestimating the `last_upload_length` by a single point because `last_upload_length = upload_lines[-1].timestamp - upload_lines[0].timestamp` is actually NOT correct. This equation doesn't actually count the time that `upload_lines[0].timestamp` adds to the total uploaded time, so I was always off by 1 point during my count. This made the upload check trigger one point worth of time too early every time.

I'll test this change tomorrow!

socs/agents/acu/agent.py
```diff
@@ -2501,14 +2496,18 @@ def _run_track(self, session, point_gen, step_time, track_axes=['az'],
...
             last_upload_az = None
             start_time = time.time()
             got_progtrack = False
             faults = {}
             got_points_in = False
             first_upload_time = None
+            t0 = time.time()
+            last_upload_length = 0
+            last_uploaded_timestamp = None
+            current_uploaded = 0
```
mhasself: I think all of these can be replaced with just "last_uploaded_timestamp = 0".

You use current_uploaded to track the time-into-the-future of the last uploaded point.  But that is easily (and much more safely) computed as `last_uploaded_timestamp - time.time()`.  That will greatly simplify things.

socs/agents/acu/agent.py
```diff
@@ -2440,16 +2440,11 @@ def _run_track(self, session, point_gen, step_time, track_axes=['az'],
         # (The cost is that stopping a scan may take a little longer.)
         MIN_STACK_ADVANCE_TIME = 3.
 
-        # Minimum nuber of points to keep in the stack.
+        # Minimum number of points to keep in the stack.
         _pbc = max(MIN_STACK_POP, MIN_STACK_ADVANCE_TIME / step_time)
         if point_batch_count is None or _pbc > point_batch_count:
             point_batch_count = _pbc
```
mhasself: point_batch_count is not used anymore, so actually this is dead code.

This is also the only place, now, where MIN_STACK_POP is used.  I think we need to preserve that, though.  All calculations of "is it time to upload points; how many points should I upload?" should account for never dropping below MIN_STACK_POP.

socs/agents/acu/agent.py
```diff
@@ -2560,44 +2559,46 @@ def _run_track(self, session, point_gen, step_time, track_axes=['az'],
...
-                            # processed.
-                            break
-
-                    # Grab the minimum batch
-                    upload_lines, lines = lines[:new_line_target], lines[new_line_target:]
+                # This happens when the current time of uploaded points is less
+                # than the MIN_STACK_ADVANCE_TIME.
+                # (Meaning we have less than the minimum amount of points uploaded).
+                if (current_uploaded <= MIN_STACK_ADVANCE_TIME):
+
```
mhasself: Replace this with test that last uploaded point is at least MIN_..._TIME in the future.

socs/agents/acu/agent.py
```diff
@@ -2560,44 +2559,46 @@ def _run_track(self, session, point_gen, step_time, track_axes=['az'],
...
+                    upload_lines = []
+                    # Grab points to upload until we'll have a total of 2 * MIN_STACK_ADVANCE_TIME uploaded.
+                    # We subtract the current_uploaded here to make sure we upload UP TO 2 * MIN_STACK_ADVANCE_TIME.
+                    while not point_prov.is_empty() and (
+                            len(upload_lines) == 0
+                            or (upload_lines[-1].timestamp - upload_lines[0].timestamp) < (2 * MIN_STACK_ADVANCE_TIME) - current_uploaded):
```
mhasself: Replace using same "in the future" check; helpfully that does not require computing a difference of upload_lines points.

mhasself (changes requested):
Thanks for work and testing here -- I propose some simplifications to the logic.

socs/agents/acu/agent.py
mjrand: Done. Removed these and set `last_uploaded_tiemstamp = 0`.

socs/agents/acu/agent.py
mjrand: Removed _pbc and this block of code as it was unused. Updated the "time to upload check" to also check that we're not below the MIN_STACK_POP worth of uploaded points.

socs/agents/acu/agent.py
mjrand: replaced with `if ((last_uploaded_timestamp - time.time()) <= MIN_STACK_ADVANCE_TIME) or (free_positions > MAX_ALLOWABLE_FREE_POSITIONS):`

socs/agents/acu/agent.py
mjrand: Updated with 

```
while not point_prov.is_empty() and (len(upload_lines) == 0
                                or upload_lines[-1].timestamp - time.time() < (2 * MIN_STACK_ADVANCE_TIME)
                                or (free_positions - len(upload_lines) > MAX_ALLOWABLE_FREE_POSITIONS)):
```

mjrand:
Updated this code with Matthew's simplification suggestsions. These suggestions have produced the behavior we want. The free points during a `two_leg` turnaround scan look like so:

image
image


I think this code has been proven to work as we want now.

mhasself (approved):
Ok -- latest version looks great.  I added one more commit to catch an edge case in "mode" transitions.

Ready for maintainer-level review, IMO.

BrianJKoopman (approved):
Looks good to me!

PR #1051, opened 2026-05-04
Create new Frequency-selectable Laser Source (FLS) agent

laurensaunders:
## Description
This is a new agent for the Frequency-selectable Laser Source (FLS), a detector bandpass calibrator
that will be used at the SO site with the LAT in May 2026.

## Motivation and Context
This agent provides an interface for the FLS in OCS.

## How Has This Been Tested?
The agent has been tested in the lab with the instrument. All operations work normally.
It has not been tested in a docker container due to issues that are currently under discussion.

## Types of changes

## Checklist:


laurensaunders:
Agent operations have now been tested in the docker container as well.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
+import argparse
+import time
+
+import numpy as np
+from ocs import ocs_agent, site_config
+from ocs.ocs_twisted import Pacemaker, TimeoutLock
+
+from socs.agents.fls.drivers import DLCSmart
+
+# TODO: put these in a .yaml file
```
sanahabhimani:
```suggestion

```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+    scan_min_freq = fls.scan_min_freq
+    scan_max_freq = fls.scan_max_freq
+    scan_step_size = fls.scan_step
+    scan_direction = fls.scan_direction
+
+    if scan_min_freq != min_freq:
+        fls.log.warn(f"Minimum frequency set to {scan_min_freq}, not {min_freq}!")
+        return False, "Scan parameter validation failed: minimum frequency."
+    if scan_max_freq != max_freq:
+        fls.log.warn(f"Minimum frequency set to {scan_max_freq}, not {max_freq}!")
```
sanahabhimani:
```suggestion
        fls.log.warn(f"Maximum frequency set to {scan_max_freq}, not {max_freq}!")
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+        return False, "Scan parameter validation failed: maximum frequency."
+    if scan_step_size != freq_step:
+        fls.log.warn(f"Scan step size set to {scan_step_size}, not {freq_step}!")
+        return False, "Scan parameter validation failed: step size."
+    if scan_direction != start_dir:
+        fls.log.warn(f"Start direction set to {scan_direction}, not {start_dir}!")
+        return False, "Scan parameter validation failed: scan direction."
+
+    fls.log.info(f"Scan parameters set: {min_freq} GHz to {max_freq} GHz "
+                 f"with step size {start_dir * freq_step}.")
```
sanahabhimani:
```suggestion   
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+        self.set_freq = None
+        self.actual_freq = None
+        self.scan_mode = None
+        self.scan_min_freq = None
+        self.scan_max_freq = None
+        self.scan_step = None
+        self.scan_direction = None
+        self.integration_time = None
+
+        agg_params = {'frame_length': 60}  # is this correct?
```
sanahabhimani:
```suggestion
        agg_params = {'frame_length': 60} 
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                                 record=True,
+                                 agg_params=agg_params,
+                                 buffer_time=1)
+
+    @ocs_agent.param('auto_acquire', type=bool, default=False)
+    def initialize(self, session, params=None):
+        """
+        initialize()
+
+        ***Task*** - Initialize the connection to the DLC Smart
```
sanahabhimani:
```suggestion
        ***Task*** - Initialize the connection to the DLC Smart
        
        Parameters:
           auto_acquire (bool): If True, start acquisition immediately after 
               initialization. Defaults to False.
```

Update docstring for `initialize` task to include info about `auto_acquire` parameter

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                self.log.warn(f"Could not start initialize because {self.lock.job}"
+                              "is already running")
+                return False, "Could not acquire lock."
+
+            # Make the connection and read out the welcome message
+            try:
+                self.dlcsmart = DLCSmart(ip_addr=self.ip, port=self.port)
+                time.sleep(1)
+                welcome = self.dlcsmart.drain_buffer()
+                print('welcome=' + str(welcome))
```
sanahabhimani: I vaguely recall from when we were debugging this together in the lab that the device sends a welcome message every time a connection is established —I can't recall if we _need_ the print statement here for that though--will you remind me?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                self.log.info('Lasers are off.')
+            else:
+                print(lasers_on)
+                self.log.warn("Could not determine if lasers are on!")
+
+            # Read the actual frequency
+            actual_freq = self.dlcsmart.get_actual_frequency()
+            try:
+                self.actual_freq = float(actual_freq)
+            except ValueError:
```
sanahabhimani:
```suggestion
            except ValueError as e:
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                self.log.warn(f'Could not convert {actual_freq} to float!')
```
sanahabhimani:
```suggestion
                self.log.warn(f'Could not convert {actual_freq} to float: {e}')
```

In case a very weird anomalous thing happens once in a blue moon, we at least know what the exact ValueError message was

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                self.actual_freq = actual_freq
```
sanahabhimani: Just to clarify the intended behavior here: in the `ValueError` case, `self.actual_freq` becomes the original non-float response object/string rather than remaining numeric. I see this in a similar vein in other try/except examples and I assume  this is intentional for downstream debugging/logging? (lines 160 and 167 for ex)

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            scan_step = scan_params[3]
+            try:
+                self.scan_step = abs(float(scan_step))
+                self.scan_direction = np.sign(scan_step)
+            except ValueError as e:
+                self.log.warn(e)
+                self.scan_step = scan_step
+
+        self.initialized = True
+        print('auto_acquire:', params['auto_acquire'])
```
sanahabhimani:
```suggestion

```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            self.log.info(f'Response from acq.start(): {resp[1]}')
+
+        return True, "FLS agent initialized"
+
+    @ocs_agent.param('test_mode', default=False, type=bool)
+    def acq(self, session, params):
+        """
+        acq()
+
+        ***Process*** - Starts the 'sampling' data acquisiton from the DLC Smart.
```
sanahabhimani: need to add parameters section for the `test_mode` bool parameter

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            last_time = time.time()
+
+            self.take_data = True
+
+            pm = Pacemaker(1 / 3, quantize=False)
+            while self.take_data:
+                pm.sleep()
+                if time.time() - last_time > 1:
+                    last_time = time.time()
+                if not self.lock.release_and_acquire(timeout=12):
```
sanahabhimani: I noticed that in other agents this line is actually nested under the previous timing condition, so I think it may need to be indented here as well. (I was looking at [Galil Axis](https://github.com/simonsobs/socs/blob/d4da326901d6c8b71a0b23e3f2431ac9be064ffa/socs/agents/galil_axis/agent.py#L199) and [LS372 agent](https://github.com/simonsobs/socs/blob/d4da326901d6c8b71a0b23e3f2431ac9be064ffa/socs/agents/lakeshore372/agent.py#L220) acq processes)

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                session.data = {"sampling_data": sampling_data,
+                                "timestamp": time.time()}
+
+                pub_data = {'timestamp': time.time(),
+                            'block_name': 'sampling_data',
+                            'data': sampling_data}
+
+                self.agent.publish_to_feed('sampling_data', pub_data)
+
+                print(pub_data)
```
sanahabhimani:
```suggestion

```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            return False, 'acq is not currently running.'
+
+    @ocs_agent.param('state', type=str, choices=['on', 'off'])
+    def toggle_laser_power(self, session, params):
+        """
+        toggle_laser_power(state)
+
+        ***Task*** - Enable or disable emission from both lasers
+
+        Parameters
```
sanahabhimani:
```suggestion
        Parameters:
```

for consistency sake in the readthedocs for the Parameters portion of each task/process

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+        ----------
```
sanahabhimani:
```suggestion

```

for consistency sake in the readthedocs for the Parameters portion of each task/process

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+        """
+        set_bias(bias)
+
+        ***Task*** - Set the bias amplitude and offset of the lasers to a preset
+                     condition.
+
+        Parameters:
+            bias (str): Preset condition to set the bias for the lasers. Options are
+                        'zero' to set the bias to zero, or 'default' to set the bias to
+                        default.
```
sanahabhimani: For longevity/readability, I think it may be worth explicitly stating in the docstring what bias values are associated with the default state, since those are the conditions being checked later in the code.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            check_bias_offset = self.tx_bias_offset
+            if bias_to_set == 'zero' and (check_bias_amp, check_bias_offset) == (0., 0.):
+                self.log.info('Bias successfully set to zero.')
+            elif bias_to_set == 'default' and round(check_bias_amp, 1) == 1.0 and round(check_bias_offset, 1) == -0.5:
+                self.log.info('Bias successfully set to default.')
+            else:
+                bias_amp, bias_off = self.dlcsmart.check_bias()
+                self.tx_bias_amp = bias_amp
+                self.tx_bias_offset = bias_off
+                check_bias_amp = self.tx_bias_amp
```
sanahabhimani:
```suggestion

```
probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                check_bias_offset = self.tx_bias_offset
```
sanahabhimani:
```suggestion

```
probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                if bias_to_set == 'zero' and (check_bias_amp, check_bias_offset) == (0., 0.):
```
sanahabhimani:
```suggestion
                if bias_to_set == 'zero' and (bias_amp, bias_off) == (0., 0.):
```
probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                    self.log.info('Bias successfully set to zero.')
+                elif bias_to_set == 'default' and round(check_bias_amp, 1) == 1.0 and round(check_bias_offset, 1) == -0.5:
```
sanahabhimani:
```suggestion
                elif bias_to_set == 'default' and round(bias_amp, 1) == 1.0 and round(bias_off, 1) == -0.5:
```

probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                    self.log.info('Bias successfully set to default.')
+                else:
+                    self.log.info(f"Bias amp is {check_bias_amp} and bias offset is {check_bias_offset}.")
```
sanahabhimani, line 342:
```suggestion
                    self.log.info(f"Bias amp is {bias_amp} and bias offset is {bias_off}.")
```
probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            time.sleep(3)
+            check_bias_amp = self.tx_bias_amp
+            check_bias_offset = self.tx_bias_offset
+            if bias_to_set == 'zero' and (check_bias_amp, check_bias_offset) == (0., 0.):
+                self.log.info('Bias successfully set to zero.')
+            elif bias_to_set == 'default' and round(check_bias_amp, 1) == 1.0 and round(check_bias_offset, 1) == -0.5:
+                self.log.info('Bias successfully set to default.')
+            else:
+                bias_amp, bias_off = self.dlcsmart.check_bias()
+                self.tx_bias_amp = bias_amp
```
sanahabhimani:
```suggestion

```

probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                self.tx_bias_offset = bias_off
```
sanahabhimani:
```suggestion

```
probably don't need the extra variable assignments surrounding `bias_off` and `bias_amp`?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+
+        with self.lock.acquire_timeout(timeout=12, job='set_frequency') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start Task because "
+                              f"{self.lock.job} is already running")
+                return False, "Could not acquire lock"
+
+            # Set the new frequency
+            set_the_freq = self.dlcsmart.set_frequency(set_frequency)
+            if set_the_freq == '0':
```
sanahabhimani: This confused me a little bit until I got on a quick slack huddle with you (Lauren) and realized that `set_the_freq` ends up returning a `0` if the frequency value was successfully sent (or set? can you confirm which one?).

So, for longevity/readability beyond Lauren/Sara/Elisa's expertise, I suggest changing `set_the_freq` to something like `freq_bool_state` or `response` or something along those lines

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+# TODO: put these in a .yaml file
+MAX_FREQ = 880.
+MIN_FREQ = 20.
+
+
+def _within(val, target, tolerance=1e-2):
+    return abs(val - target) <= tolerance
+
+
+def _check_scan_params(fls, min_freq, max_freq, freq_step, start_dir):
```
sanahabhimani: we actually don't need this helper function since a `check_scan_params` exists in the driver code so can remove this and any usage of it

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+
+            if start_dir == -1 and not _within(self.actual_freq, max_freq):
+                self.log.warn('run_frequency_sweeps called with decreasing frequency, '
+                              'but laser is not at max_freq.')
+                if max_freq != self.set_freq:
+                    self.log.warn(f'Set frequency is {self.set_freq} and max_freq is {max_freq}.')
+
+            self.dlcsmart.set_scan_params(min_freq, max_freq, freq_step, start_dir, int_time)
+            time.sleep(0.1)
+            csp = _check_scan_params(fls, min_freq, max_freq, freq_step, start_dir)
```
sanahabhimani:
```suggestion

```

You might want to instead replace this with the driver code function? And if so, then don't need to remove the logic around the `csp` condition

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+
+        assert min_freq < max_freq, "max_freq must be greater than min_freq!"
+        assert min_freq >= MIN_FREQ, f"min_freq must be at least {MIN_FREQ} GHz."
+        assert min_freq < MAX_FREQ, f"min_freq must be less than {MAX_FREQ} GHz."
+        assert max_freq >= MIN_FREQ, f"max_freq must be at least {MIN_FREQ} GHz."
+        assert max_freq < MAX_FREQ, f"max_freq must be less than {MAX_FREQ} GHz."
+        assert freq_step >= 0.01, "minimum step size is 0.01 GHz."
+        assert start_dir in (-1, 1), "Choose start_dir=1 (increasing) or -1 (decreasing)"
+
+        fls = self
```
sanahabhimani:
```suggestion

```

don't need this as it's an artifact from interacting with the helper function at the top of the agent `_check_scan_params()` that is no longer necessary

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+            if start_dir == -1 and not _within(self.actual_freq, max_freq):
+                self.log.warn('run_frequency_sweeps called with decreasing frequency, '
+                              'but laser is not at max_freq.')
+                if max_freq != self.set_freq:
+                    self.log.warn(f'Set frequency is {self.set_freq} and max_freq is {max_freq}.')
+
+            self.dlcsmart.set_scan_params(min_freq, max_freq, freq_step, start_dir, int_time)
+            time.sleep(0.1)
+            csp = _check_scan_params(fls, min_freq, max_freq, freq_step, start_dir)
+            if not csp:
```
sanahabhimani:
```suggestion

```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                return False, "Could not correctly set scan params."
```
sanahabhimani:
```suggestion

```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+        ***Task*** - Set the integration time of the laser system. Time is in
+                     milliseconds.
+
+        Parameters:
+            integration_time (float): The integration time in milliseconds.
+
+        """
+        int_time = params['integration_time']
+        self.dlcsmart.param_set("lockin:integration-time", int_time)
+        return True, f"Set integration time to {int_time}."
```
sanahabhimani: My impression (from when David was reviewing the galil axis agent many moons ago) is that in cases like this — where the task sends a command but does not explicitly verify the resulting hardware state afterward (unlike, for example, `set_bias()`, where the bias values are queried and checked) — it is better to phrase the return statement as something like:
`f"Commanded integration time to be set to {int_time}."`  It's worth doing that for all tasks where this applies.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,580 @@
...
+                self.log.warn(f'Bias amplitude is {bias_amp} and bias offset '
+                              f'is {bias_offset}. Setting bias to zero, then '
+                              f'turning lasers {state}.')
+                self.dlcsmart.set_bias_to_zero()
+                time.sleep(0.3)
+                bias_amp, bias_offset = self.dlcsmart.check_bias()
+                if bias_amp != 0.0 or bias_offset != 0.0:
+                    return False, "Bias could not be set to zero so did not toggle laser power."
+
+            countdown = 10
```
sanahabhimani, line 276: I briefly discussed the countdown logic (lines 333–339) with Brian, see thoughts below:

Right now, my understanding is that if you're running from the terminal outside of a Docker container as you had been when developing the agent on the lab computer, the practical way to stop the task mid-countdown would just be to stop the agent entirely via keyboard interrupt. Similarly, once this is running through OCS-web / inside a Docker container, there also wouldn’t be a clean way to cancel the countdown and stop the task without bringing the agent down in its current form. This is non-ideal.

You need an aborter. One example comes from the[ webcam agent.](https://github.com/simonsobs/socs/blob/d4da326901d6c8b71a0b23e3f2431ac9be064ffa/socs/agents/rtsp_camera/agent.py#L395-L397), another example comes from the [acu agent](https://github.com/simonsobs/socs/blob/d4da326901d6c8b71a0b23e3f2431ac9be064ffa/socs/agents/acu/agent.py#L472).


Basically, you don't need to remove or restructure the existing countdown `while` loop itself, but for the countdown to be interruptible/useful, you need something like the following added to your agent:
```
    def _abort_laser_power(self, session, params):
        if session.status == "running":
            session.set_status("stopping")
```

and then you need to update the `register_task` line for `toggle_laser_power()` to include an `abort` argument, example [here](https://github.com/simonsobs/socs/blob/d4da326901d6c8b71a0b23e3f2431ac9be064ffa/socs/agents/rtsp_camera/agent.py#L591)  

On ocs_web, this will allow an abort button to appear for the `toggle_laser_power()` task.

You may also find lines 252-262 helpful in the [fake data agent](https://github.com/simonsobs/ocs/blob/3a1de8c0927772026fff829698627a9e97154466/ocs/agents/fake_data/agent.py#L252). You also see versions of these lines in the acu and camera agent.

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,368 @@
...
+
+    def read_all(self, decode=True):
+        """
+        Handles decoding anything read out from the DLC Smart.
+        """
+        data = b""
+        expect_prompt = False
+        while True:
+            try:
+                #                chunk = self.sock.recv(1024)
```
sanahabhimani:
```suggestion

```

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,368 @@
...
+                      'scan_mode': scan_mode,
+                      'scan_min_frequency': float(scan_params[1]),
+                      'scan_max_frequency': float(scan_params[2]),
+                      'scan_step': abs(scan_params[3]),
+                      'scan_direction': int(np.sign(scan_params[3])),
+                      'integration_time': float(scan_params[4]),
+                      }
+        return value_dict
+
+    def get_scan_data(self):
```
sanahabhimani, line 331: If I recall correctly, you intentionally avoided calling this method from the agent because it was meant more as a helper/backup function for usage outside of OCS/SOCS workflows. In that case, I think it’s reasonable to keep it around for non-OCS users who may want to query data without accompanying timestamps. Though if you feel it no longer serves much purpose given the additional methods added since initially developing this driver code, it would also be fine to remove it. If you do keep it, it is worth adding a short note in the docstring clarifying that it exists in the driver layer for non-OCS usage and is not intended for SOCS/OCS workflows.

docs/agents/fls.rst
```diff
@@ -0,0 +1,149 @@
...
+
+  {'agent-class': 'FLSAgent',
+   'instance-id': 'fls',
+   'arguments': ['--ip', '169.254.18.24',
+                 '--port', '1998',
+                 '--mode', 'acq']}
+
+Each device requires configuration under 'agent-instances'. See the OCS site
+configs documentation for more details.
+
```
sanahabhimani, line 39: Will you also add the docker-compose configuration file setup? Example from galil axis agent [here](https://github.com/simonsobs/socs/blob/main/docs/agents/galil_axis.rst) (and obviously your own docker compose .yaml file you used for testing)

docs/agents/fls.rst
```diff
@@ -0,0 +1,149 @@
...
+     system is ready, it will produce a series of audible tones, and the
+     light under "System Ready" on the front of the unit will flash green.
+  2. Start the FLS Agent. This will initialize the connection. The DLC Smart
+     unit will produce a series of audible tones, and the Agent will begin
+     data acquisition.
+  3. To ensure that the voltage bias is set to zero::
+
+       client.set_bias(bias='zero')
+
+  4. Ensure that the U-shaped link is removed from the BNC breakout box,
```
sanahabhimani, line 82: I think it’s worth explicitly noting to the reader here that the code includes a countdown before it proceeds to toggle laser power. That helps further emphasize just how critical it is to have the U-shaped link removed before continuing. Once the aborter logic is added, it would also be good to mention that an abort button will appear in the task window for `toggle_laser_power` on OCS-web.

sanahabhimani (changes requested):
Thanks, Lauren! Mainly one "major" change --- adding an aborter to your `toggle_laser_power` task so that the countdown is accompanied by the ability to safely stop the task mid-execution without needing to crash/stop the entire agent more aggressively. I tried to make the instructions for implementing that as straightforward and approachable as possible to help you with the quick turnaround time.

Other changes are pretty superficial and include things like:
- removing some print statements
- minor docstring updates in the agent 
- adding 2 more details to the `fls.rst` file 
- a couple clarifying questions that might not have direct instructions on what to change (if anything) but we can correspond here/offline to be more efficient with our time

socs/agents/fls/agent.py
laurensaunders: We don't need the print statement, it was just for debugging (and this line only prints True/False, not the actual welcome)

socs/agents/fls/agent.py
laurensaunders: This was intentional. The case where we didn't get something that could be converted to a float typically involved not reading all of the characters in a previous request. The intention was to log the problem for debugging instead of having the Task fail. Should this be changed?

docs/agents/fls.rst
BrianJKoopman, line 1: Add this to [the index file](https://github.com/simonsobs/socs/blob/main/docs/index.rst) in alphabetical order. That way it'll show up in the sidebar in the compiled docs.

BrianJKoopman, line 1: You should also add an 'Agent API' section. This shows users what tasks/processes are availble. See [this ocs docs page](https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html#agent-reference-pages) (or any other agent docs page) for an example.

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        Handles decoding anything read out from the DLC Smart.
+        """
+        data = b""
+        expect_prompt = False
+        while True:
+            try:
+                chunk = self.recv(1024)
+            except self.timeout:
+                break
```
BrianJKoopman: This except doesn't look right to me. `self.timeout` is a float. If the socket isn't ready, or there's some other error, `self.recv(1024` will raise a `ConnectionError`. You should catch that here.

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,370 @@
...
+            data += chunk
+            if expect_prompt and data.endswith(b">"):
+                break
+            if not expect_prompt and b"\n" in chunk:
+                break
```
BrianJKoopman: It looks like `expect_prompt` is always `False`, so this first `if` never executes, and the second one's first condition `not False` is always `True`, so this could all be rewritten:

```suggestion
            if b"\n" in chunk:
                break
```

And `expect_prompt = False` can be removed up above. Is this the logic you were aiming for? When do you expect to get the prompt (`">"`), just at start up?

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,370 @@
...
+
+        Return
+        ------
+        resp: The response from the DLC Smart
+        """
+        msg = f"(param-set! '{param} {val})"
+        resp = self.send_msg(msg)
+        return resp
+
+    def command(self, param, decode=False, vals=[]):
```
BrianJKoopman: You shouldn't use empty lists or dicts as default arguments. See: https://www.hackerone.com/blog/python-pitfalls-perils-using-lists-and-dicts-default-arguments

I'd change this to:
```suggestion
    def command(self, param, decode=False, vals=None):
```

And then below check for `if vals is not None`, rather than `if len(vals)`. You might also check it is, in fact, a `list`.

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,370 @@
...
+                msg += " " + str(val)
+        msg += ")"
+        resp = self.send_msg(msg, decode=decode)
+        return resp
+
+    # network operations and checks
+
+    def get_ip(self):
+        resp = self.param_ref("net-conf:ip-addr")
+        self.ip_address = resp
```
BrianJKoopman: This might get potentially confusing. `self.ip_address` is already set when we create the `DLCSmart` object. Presumably because we were able to query the device the IP hasn't changed. If the IP does change, we probably can't query the device any more and need to recreate the socket. How does this get used in practice?

socs/agents/fls/drivers.py
```diff
@@ -0,0 +1,370 @@
...
+          - Scan integration time (ms)
+        For general monitoring purposes.
+        """
+        self.command("lockin:lock-in-reset")
+        time.sleep(0.3)
+
+        # Photocurrent
+        photocurrent = self.param_ref("lockin:lock-in-value-nanoamp")
+        if '#t' in photocurrent:
+            photocurrent = photocurrent.strip('(').strip(' #t)')
```
BrianJKoopman: You might want this to either be:

```suggestion
            photocurrent = photocurrent.lstrip('(').rstrip(' #t)')
```

Or otherwise you could use:
```suggestion
            photocurrent = photocurrent.strip('( #t)')
```

`.strip()` will strip the set of provided characters from either end of the string. For example (I'm not sure what `photocurrent` typically looks like):
```Python
>>> photocurrent = '(9000 #t)'
>>> photocurrent.strip('( #t)')
'9000'
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+
+        self.agent.register_feed('sampling_data',
+                                 record=True,
+                                 agg_params=agg_params,
+                                 buffer_time=1)
+
+    @ocs_agent.param('auto_acquire', type=bool, default=False)
+    def initialize(self, session, params=None):
+        """
+        initialize()
```
BrianJKoopman:
```suggestion
        initialize(auto_acquire=False)
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+        with self.lock.acquire_timeout(0, job='initialize') as acquired:
+            if not acquired:
+                self.log.warn(f"Could not start initialize because {self.lock.job}"
+                              "is already running")
+                return False, "Could not acquire lock."
+
+            # Make the connection and read out the welcome message
+            try:
+                self.dlcsmart = DLCSmart(ip_addr=self.ip, port=self.port)
+                time.sleep(1)
```
BrianJKoopman: We just added this during our debug session, it shouldn't be needed.
```suggestion
```

socs/agents/fls/agent.py
BrianJKoopman: I'm not sure I follow this. If you get something that can't be cast to float you will get a ValueError on line 109 here (when DLCSmart.get_actual_frequenc() does its cast to float) and this try/except won't be hit anyway.
Is the response to self.param_ref("frequency:frequency-act") in the drivers ever something that can't be cast?

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+        if params['auto_acquire']:
+            resp = self.agent.start('acq', params={})
+            self.log.info(f'Response from acq.start(): {resp[1]}')
+
+        return True, "FLS agent initialized"
+
+    @ocs_agent.param('test_mode', default=False, type=bool)
+    def acq(self, session, params):
+        """
+        acq()
```
BrianJKoopman:
```suggestion
        acq(test_mode=False)
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+                     'bias_voltage': 0.999834227,
+                     'bias_offset': -0.498235892,
+                     'lasers_on': True,
+                     'scan_mode': 'fast',
+                     'scan_min_frequency': 120.0,
+                     'scan_max_frequency': 180.0,
+                     'scan_step': 0.05,
+                     'scan_direction': 1,
+                     'integration_time': 299.3421
+                     'timestamp': 1771277799.562098}
```
BrianJKoopman: (I'm going to comment here on the structure, and make the required modifications in a separate comment below.)

I would suggest getting rid of the extra layer of nesting under the 'fields' key and just having the flat structure:
```suggestion
                {'set_frequency': 110.0,
                 'actual_frequency': 109.3425,
                 'photocurrent': 0.1124,
                 'bias_voltage': 0.999834227,
                 'bias_offset': -0.498235892,
                 'lasers_on': True,
                 'scan_mode': 'fast',
                 'scan_min_frequency': 120.0,
                 'scan_max_frequency': 180.0,
                 'scan_step': 0.05,
                 'scan_direction': 1,
                 'integration_time': 299.3421
                 'timestamp': 1771277799.562098}
```

This makes the items easier to access: `session.data['set_frequency']` instead of `session.data['fields']['set_frequency']`.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+                self.scan_direction = data['scan_direction']
+                self.lasers_on = data['lasers_on']
+                self.integration_time = data['integration_time']
+
+                sampling_data = {}
+                for key, val in data.items():
+                    sampling_data[key] = val
+
+                session.data = {"sampling_data": sampling_data,
+                                "timestamp": time.time()}
```
BrianJKoopman: To match the suggestion I made above:
```suggestion
                data['timestamp'] = time.time()
                session.data = data
```

socs/agents/fls/agent.py
BrianJKoopman: Just setting `session.status('stopping')` via the aborter here isn't enough. You need to check that status during this countdown and if you're trying to abort, do so, i.e. `return False, "operation aborted"`.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+
+        Parameters:
+            integration_time (float): The integration time in milliseconds.
+
+        """
+        int_time = params['integration_time']
+        self.dlcsmart.param_set("lockin:integration-time", int_time)
+        return True, f"Set integration time to {int_time}."
+
+    @ocs_agent.param('frequency', type=float)
```
BrianJKoopman: You should be able to use the `param` decorator to check the frequency is within range:
```suggestion
    @ocs_agent.param('frequency', type=float, check=lambda x: MIN_FREQ <= x < MAX_FREQ)
```

That way the client gets the error immediately.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+                self.log.warn(f"Could not start Task because "
+                              f"{self.lock.job} is already running")
+                return False, "Could not acquire lock"
+
+            # Set the new frequency
+            response = self.dlcsmart.set_frequency(set_frequency)
+            if response == '0':
+                return True, f"Frequency set to {set_frequency} in the DLC Smart."
+
+#        return True, f"Set frequency to {set_frequency} GHz"
```
BrianJKoopman:
```suggestion
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,575 @@
...
+
+        assert min_freq < max_freq, "max_freq must be greater than min_freq!"
+        assert min_freq >= MIN_FREQ, f"min_freq must be at least {MIN_FREQ} GHz."
+        assert min_freq < MAX_FREQ, f"min_freq must be less than {MAX_FREQ} GHz."
+        assert max_freq >= MIN_FREQ, f"max_freq must be at least {MIN_FREQ} GHz."
+        assert max_freq < MAX_FREQ, f"max_freq must be less than {MAX_FREQ} GHz."
+        assert freq_step >= 0.01, "minimum step size is 0.01 GHz."
+        assert start_dir in (-1, 1), "Choose start_dir=1 (increasing) or -1 (decreasing)"
+        assert int_time > 0.5, "integration time must be greater than 0.5 ms."
+        assert int_time <= 3000, "integration time must be no more than 3000 ms."
```
BrianJKoopman: Make these into checks in the [param decorators](https://ocs.readthedocs.io/en/main/developer/agent_references/params.html#validation-using-param).

BrianJKoopman (changes requested):
Thanks for the PR! A few comments below. I know this is in the field right now, so I'd say implement what you can. Happy to discuss any of these, of course.

One general request, can you `chmod -x` the agent and driver files. I think they're set with that because they were written on Windows. We don't need that here though.

socs/agents/fls/agent.py
sanahabhimani: looks like this was a forgotten suggested change

socs/agents/fls/agent.py
sanahabhimani: also pointing this out as a forgotten change in the re-review

socs/agents/fls/agent.py
```diff
@@ -434,8 +402,8 @@ def set_frequency(self, session, params):
                 return False, "Could not acquire lock"
 
             # Set the new frequency
-            set_the_freq = self.dlcsmart.set_frequency(set_frequency)
-            if set_the_freq == '0':
+            response = self.dlcsmart.set_frequency(set_frequency)
+            if response == '0':
                 return True, f"Frequency set to {set_frequency} in the DLC Smart."
```
sanahabhimani: change to

`return True, f"Commanded frequency to be set to {set_frequency} in the DLC Smart"`

socs/agents/fls/drivers.py
laurensaunders: I think this function is a relic from development that isn't used anymore (we were trying to use it in switching between DHCP and static IP).

sanahabhimani (dismissed):

socs/agents/fls/drivers.py
laurensaunders: I think this check with `expect_prompt` is a relic of some debugging. We thought the prompt `>` was expected after every returned value (based on section 2 of the command reference), but we found that to not be the case. I will remove the check.

socs/agents/fls/agent.py
BrianJKoopman: Good catch! Release and reacquire docs here for reference: https://ocs.readthedocs.io/en/main/developer/agent_references/timeoutlock.html#release-and-reacquire

docs/agents/fls.rst
```diff
@@ -0,0 +1,178 @@
...
+  {'agent-class': 'FLSAgent',
+   'instance-id': 'fls',
+   'arguments': ['--ip', '169.254.18.24',
+                 '--port', '1998',
+                 '--mode', 'acq']}
+
+Each device requires configuration under 'agent-instances'. See the OCS site
+configs documentation for more details.
+
+Docker-compose
```
BrianJKoopman:
```suggestion
Docker Compose
```
(Just for consistency with other pages.)

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+        """
+        acq(test_mode=False)
+
+        ***Process*** - Starts the 'sampling' data acquisiton from the DLC Smart.
+
+        Parameters:
+            test mode (bool): If True, the acquisition loop breaks after one iteration.
+
+        Notes:
+            The data collected are stored in session data in the structure:
```
BrianJKoopman:
```suggestion
            The data collected are stored in session data in the structure::
```

Second `:` is needed to render the example block as code.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+                                 record=True,
+                                 agg_params=agg_params,
+                                 buffer_time=1)
+
+    @ocs_agent.param('auto_acquire', type=bool, default=False)
+    def initialize(self, session, params=None):
+        """
+        initialize(auto_acquire=False)
+
+        ***Task*** - Initialize the connection to the DLC Smart
```
BrianJKoopman:
```suggestion
        **Task** - Initialize the connection to the DLC Smart
```

Too many `*`'s. Only two are needed to make it bold.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+        if session.status == "running":
+            session.set_status("stopping")
+
+    @ocs_agent.param('bias', type=str, choices=['default', 'zero'])
+    def set_bias(self, session, params):
+        """
+        set_bias(bias)
+
+        ***Task*** - Set the bias amplitude and offset of the lasers to a preset
+                     condition.
```
BrianJKoopman:
```suggestion
        **Task** - Set the bias amplitude and offset of the lasers to a preset
        condition.
```

That extra whitespace makes this render incorrectly.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+                    return False, "Bias not successfully set."
+        return True, f"Bias successfully set to {bias_to_set}."
+
+    @ocs_agent.param('integration_time', type=float)
+    def set_integration_time(self, session, params):
+        """
+        set_integration_time(integration_time)
+
+        ***Task*** - Set the integration time of the laser system. Time is in
+                     milliseconds.
```
BrianJKoopman:
```suggestion
        **Task** - Set the integration time of the laser system. Time is in
        milliseconds.
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+        self.dlcsmart.param_set("lockin:integration-time", int_time)
+        return True, f"Commanded integration time to be set to {int_time}."
+
+    @ocs_agent.param('frequency', type=float, check=lambda x: MIN_FREQ <= x < MAX_FREQ)
+    def set_frequency(self, session, params):
+        """
+        set_frequency(frequency)
+
+        ***Task*** - Set the frequency of the laser system. Frequency must be
+                     between 20 GHz and 880 GHz.
```
BrianJKoopman:
```suggestion
        **Task** - Set the frequency of the laser system. Frequency must be
        between 20 GHz and 880 GHz.
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+    @ocs_agent.param('max_frequency', type=float, check=lambda x: MIN_FREQ <= x < MAX_FREQ)
+    @ocs_agent.param('start_direction', type=int, choices=[-1, 1])
+    @ocs_agent.param('frequency_step', type=float, default=0.05, check=lambda x: x >= 0.01)
+    @ocs_agent.param('int_time', type=float, default=300., check=lambda x: 0.5 < x <= 3000)
+    def run_frequency_sweeps(self, session, params):
+        """
+        run_frequency_sweeps(min_frequency, max_frequency, start_direction,
+                             frequency_step, num_of_sweeps)
+
+        ***Task*** - Run frequency sweeps between the two frequency values.
```
BrianJKoopman:
```suggestion
        **Task** - Run frequency sweeps between the two frequency values.
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+            self.dlcsmart.start_scan()
+            return True, f"Started scan from {min_freq} GHz to {max_freq} GHz with step size {freq_step} and direction {start_dir}."
+
+    @ocs_agent.param("_")
+    def stop_frequency_sweep(self, agent, params):
+        """
+        stop_frequency_sweep()
+
+        ***Task*** - Send a stop command to the DLC Smart to stop running a frequency
+                     sweep. This command may be run during or at the end of a sweep.
```
BrianJKoopman:
```suggestion
        **Task** - Send a stop command to the DLC Smart to stop running a frequency
        sweep. This command may be run during or at the end of a sweep.
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+            return True, 'requested to stop taking sampling data.'
+        else:
+            return False, 'acq is not currently running.'
+
+    @ocs_agent.param('state', type=str, choices=['on', 'off'])
+    def toggle_laser_power(self, session, params):
+        """
+        toggle_laser_power(state)
+
+        ***Task*** - Enable or disable emission from both lasers
```
BrianJKoopman:
```suggestion
        **Task** - Enable or disable emission from both lasers
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+            self.log.info(f'Response from acq.start(): {resp[1]}')
+
+        return True, "FLS agent initialized"
+
+    @ocs_agent.param('test_mode', default=False, type=bool)
+    def acq(self, session, params):
+        """
+        acq(test_mode=False)
+
+        ***Process*** - Starts the 'sampling' data acquisiton from the DLC Smart.
```
BrianJKoopman:
```suggestion
        **Process** - Starts the 'sampling' data acquisition from the DLC Smart.
```

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+
+    @ocs_agent.param('min_frequency', type=float, check=lambda x: MIN_FREQ <= x < MAX_FREQ)
+    @ocs_agent.param('max_frequency', type=float, check=lambda x: MIN_FREQ <= x < MAX_FREQ)
+    @ocs_agent.param('start_direction', type=int, choices=[-1, 1])
+    @ocs_agent.param('frequency_step', type=float, default=0.05, check=lambda x: x >= 0.01)
+    @ocs_agent.param('int_time', type=float, default=300., check=lambda x: 0.5 < x <= 3000)
+    def run_frequency_sweeps(self, session, params):
+        """
+        run_frequency_sweeps(min_frequency, max_frequency, start_direction,
+                             frequency_step, num_of_sweeps)
```
BrianJKoopman:
```suggestion
        run_frequency_sweeps(min_frequency, max_frequency, start_direction, \
                             frequency_step, num_of_sweeps)
```

Fixes a rendering issue.

socs/agents/fls/agent.py
```diff
@@ -0,0 +1,565 @@
...
+            min_frequency (float): Minimum frequency for the sweeps (GHz).
+            max_frequency (float): Maximum frequency for the sweeps (GHz).
+            start_direction (int): Indicates increasing or decreasing frequency. Use
+                                   start_direction = 1 for increasing frequency, or
+                                   start_direction = -1 for decreasing frequency
+            frequency_step (float): Step size between frequencies during the sweep (GHz).
+                                    Must be at least 0.01 GHz.
+            int_time (float): Integration time for each step of the sweep (ms). Default
+                              chooses the last set integration time.
+        """
```
BrianJKoopman, line 414: I think it's worth adding a note in the docstring here (in new a `Note:` section) that the task returns quickly and the frequency sweep is left running. I was a bit confused by the split of the `stop_frequency_sweep` task, thinking it was meant to be an ocs 'aborter'.

socs/agents/fls/agent.py
BrianJKoopman: This is still unresolved. In this instance I think this should be changed to:
```Python
            self.actual_freq = self.dlcsmart.get_actual_frequency()
```
dropping all of the casting and error handling on lines 109-113.

The same also seems true for the params retrieved with `DLCSmart.check_scan_params()`, since that also already is casting responses as `float`.

BrianJKoopman (changes requested):
Thanks for all the updates! I went through and marked past suggestions as resolved where changes were made.

There's [one lingering comment](https://github.com/simonsobs/socs/pull/1051#discussion_r3220777168) related to error handling of values when casting on initialization that I think would be good to address since it will really reduce the amount of code in the initialize task.

The rest of this is just cleaning up some issues with rendering the docs. I tried to suggest changes everywhere so you can batch and commit them easily.

BrianJKoopman (approved):
Thanks for all the updates! Latest changes look good!

PR #1053, opened 2026-05-07
acu: Add gentle stop gen for free_form scans

mjrand:
Added a generator for gentle stops for free_form scans. Now free_form scans should have a smooth list of TrackPoints to follow to v=0 instead of the last point having a non-zero az.

## Description

Added extra utility to the PointProvider class in the form of a new `stop()` function. Old `stop()` function is preserved through the `abort()` function.

The new stop function will try to gently stop the telescope. If not free_form, the PointProvider will set the last point az_vel=0 and el_vel=0 before clearing the stash and gen. If free_form, the PointProvider will use `turnarounds.gen_free_form_stop` to generate a gentle track to v=0. This new function uses `_gen_trajectory` to solve a simple track from the current az to the same az with v=0, like the ACU does when not in free_form mode.

## Motivation and Context

Two_leg turnaround scans were routinely failing to az_summary_faults when trying to end scans and stop telescope movement. My hypothesis is that because the ACU is in free_form mode and the last az_vel is non-zero, the ACU has no track to follow and throws the summary fault because its not at v=0. This implementation should rest the platform at v=0 before it runs out of points.

## How Has This Been Tested?

Needs to be tested on a platform.

## Types of changes

## Checklist:


socs/agents/acu/drivers.py
```diff
@@ -134,10 +134,57 @@ def pop(self):
         self._request(1)
         return self._stash.pop(0)
 
-    def stop(self):
+    def abort(self):
+        """
+            Nuclear option to clear stash and gen immediately.
```
mhasself: De-indent 4 spaces (here and `stop`).

socs/agents/acu/drivers.py
```diff
@@ -134,10 +134,57 @@ def pop(self):
...
+            If free_form, then a graceful free_form stop is generated using
+            the timestamp, az, and az_vel of the final point.
+
+            args:
+                free_form (bool): Arg tracking if the scan is a done with free_form=True.
+        """
+
+        self._request(1)
+
+        final_point = self._stash[0]
```
mhasself: It is not guaranteed that _stash[0] exists.  All edge cases must be handled.  One option is to keep an internal cache of the last point yielded to the consumer.  That's nice because you can use it instead of stash[0], in any case.  (And if no points have ever been yielded to the consumer, then this function should do nothing, anyway.)

socs/agents/acu/drivers.py
```diff
@@ -134,10 +134,57 @@ def pop(self):
...
+
+        if free_form:
+            STOP_TIME = 5  # seconds
```
mhasself: So as implemented here it will solve for a solution that has the position return to the starting point, in 5 seconds, with vel -> 0.

The 5 seconds seems arbitrary, I suspect it is generated via "surely this will always be long enough", which I interpret as "usually too long".

I think the behavior where it comes back to settle at the last commanded point is actually not useful.  It would be better to just steadily decelerate to v=0.  That should occur "as quickly as possible", which sort of means we need to know the characteristic acceleration (or max accel) for this scan (this is basically soft-coding the "5 seconds").

socs/agents/acu/turnarounds.py
```diff
@@ -70,6 +70,64 @@ def gen_turnaround(turnaround_method, t0, az0, el0, v0, turntime, az_flag, el_fl
...
+        az_flag (int): The az flag used by the ACU. Inherited from the scan generation function and not changed. Used for TrackPoints.
+        el_flag (int): The el flag used by the ACU. Inherited from the scan generation function and not changed. Used for TrackPoints.
+        point_group_batch (int): the point group batch used by the ACU. Inherited from the scan generation function and not changed. Used for TrackPoints.
+        step_time (float): The step time between points in the turnaround. Defaults to 0.1 seconds (10Hz).
+
+    Returns:
+        free_form_stop_track (list of TrackPoints): A list of generated TrackPoints for the platform to follow for a gentle stop.
+    """
+
+    from sim_scan import TrackPoint
```
mhasself: Fix import

socs/agents/acu/drivers.py
```diff
@@ -134,10 +134,57 @@ def pop(self):
...
+            final_el = final_point.el
+            final_az_vel = final_point.az_vel
+            final_az_flag = final_point.az_flag
+            final_el_flag = final_point.el_flag
+            final_point_group_batch = 1  # Give a non-zero point_batch_group to ensure full upload.
+
+            final_point.point_group_batch = final_point_group_batch
+
+            stop_point_track = turnarounds.gen_free_form_stop(t0=final_timestamp, v0=final_az_vel,
+                                                              az0=final_az, el0=final_el, stop_time=STOP_TIME,
```
mhasself: Apparently it's stoptime, not stop_time.

socs/agents/acu/turnarounds.py
```diff
@@ -70,6 +70,64 @@ def gen_turnaround(turnaround_method, t0, az0, el0, v0, turntime, az_flag, el_fl
...
+    az_start = 0
+    az_target = az_start
+    v_start = v0
+    v_target = 0
+    a_start = 0
+    a_target = 0
+
+    # Uses standard turnaround method as that method solves using t, az, v.
+    ts, azs, vs = _gen_trajectory(t_start, t_target, 0,
+                                  az_start, az_target, v_start,
```
mhasself: It would be cleaner to just pass all your things, above, as kwargs to _gen_trajectory.  E.g.
```
ts, azs, vs = _gen_trajectory(
    t_start=0,
    t_target=stoptime,
    ...
```
That reduces the number of local variables that are zero and are only used once, without losing any self-documentation power.

mhasself (changes requested):
Prelim review -- this is a good feature to have!

socs/agents/acu/turnarounds.py
mjrand: Fixed

socs/agents/acu/drivers.py
mjrand: Fixed

socs/agents/acu/drivers.py
mjrand: The 5 seconds is admittedly arbitrary. I modeled the "return to the last commanded position" after what the ACU seemed to do when it ended a scan not in free_form mode. This isn't exactly required to stop in this way as you say, it was just really easy to implement it in this way. This hard coding of 5seconds of stop time (which could probably be reduced by a few seconds) meant I didn't have to pass any extra arguments through `generate_scan` to `_run_track` and into `PointProvider`. 

If we want to have a not hardcoded STOP_TIME based on the max accel I think I could use this equation?
```
# Do we need to limit the az_accel?  This limit comes from a
        # maximum jerk parameter; the equation below (without the
        # empirical 0.85 adjustment) is stated in the SATP ACU ICD.
        min_turnaround_time = (0.85 * az_speed / 9 * 11.616)**.5
        max_turnaround_accel = 2 * az_speed / min_turnaround_time
```

where `min_stop_time = min_turnaround_time / 2` (or just use `min_turnaround_time` as the STOP_TIME)

socs/agents/acu/drivers.py
mjrand: Fixed

socs/agents/acu/drivers.py
mjrand: Looks like the `min_turanround_time` for a `v=0.5` scan is ~0.74s (~0.93s for `v=0.8`). This gives ~7 (or ~9) TrackPoints for a stop. I'm sorta worried this number of points would end up maybe too resolution, especially at higher az velocities? I can test this implementation though, maybe it wouldn't be an issue. Here's what a "get to v->0 as quick as possible" while steadily accelerating to 0 looks like for `v=0.5`
image

and for `v=0.8`
image

(Note these stops aren't started at the same az or az_vels, they're arbitrary depictions)

socs/agents/acu/drivers.py
mjrand: Added a `last_yielded_point` variable into PointProvider. Now this function takes the last yielded point and builds the stop off of that.

socs/agents/acu/turnarounds.py
mjrand: Done

socs/agents/acu/drivers.py
```diff
@@ -132,12 +133,61 @@ def _request(self, n):
...
+
+        If free_form, then a graceful free_form stop is generated using
+        the timestamp, az, and az_vel of the final point.
+
+        args:
+            free_form (bool): Arg tracking if the scan is a done with free_form=True.
+        """
+        stop_time = 5 if stop_time is None else stop_time  # seconds
+
+        self._request(1)
```
mhasself: You don't need this now ... but you do need to clear the _stash.  And since you later also set _gen = None, you might as well just call self.abort() here and that will take care of both.

socs/agents/acu/drivers.py
```diff
@@ -132,12 +133,61 @@ def _request(self, n):
...
+
+        final_point = self._last_yielded_point
```
mhasself: Check if this is None, and if so just return, because platform was never in motion.

socs/agents/acu/agent.py
```diff
@@ -2396,7 +2396,7 @@ def generate_scan(self, session, params):
         return ret_val
 
     @inlineCallbacks
-    def _run_track(self, session, point_gen, step_time, track_axes=['az'],
+    def _run_track(self, session, point_gen, step_time, min_stop_time, track_axes=['az'],
```
mhasself: Ok ... so we need to change this to "max_stop_accel".  And it should be optional.

socs/agents/acu/drivers.py
```diff
@@ -132,12 +133,61 @@ def _request(self, n):
...
+        Stop point gen gracefully by clearing stash + gen and
+        setting the last point's velocities to 0.
+
+        If free_form, then a graceful free_form stop is generated using
+        the timestamp, az, and az_vel of the final point.
+
+        args:
+            free_form (bool): Arg tracking if the scan is a done with free_form=True.
+        """
+        stop_time = 5 if stop_time is None else stop_time  # seconds
```
mhasself: Ok, so instead of stop_time we should be passing in a max_accel parameter.

Quick profiling of your turnaround function suggests that setting stop_time = 2 * abs(final_az_vel / max_accel) should be safe enough.

mhasself (changes requested):
Great ... a couple things from testing, and request to change time control parameter to max accel instead of stop_time.

socs/agents/acu/drivers.py
mjrand: Changed

socs/agents/acu/drivers.py
mjrand: Added Check

socs/agents/acu/agent.py
mjrand: Changed to max_stop_accel. From `turnarounds.py`:

```Peak acceleration is = 1.5*average_acceleration = 3 * v0/turn_time```

so

```
v0 = final_point.az_vel
stop_time = 3 * v0 / max_stop_accel
```

File: socs/agents/acu/drivers.py (hunk 10 above)
Comment (mjrand) [reply socs/agents/acu/drivers.py]: The `gen_turnaround` functions all accept `turntime`, so I've precalc'd the desired stop_time from the `max_stop_accel` parameter in passed through `Point_Provider.stop(free_form, max_stop_accel)` call.

**Comment by mjrand** [comment]
Ok I changed a bunch of the `stop_time` passes in the agent to pass `max_stop_accel` to the PointProvider instead.

There was an interesting edge case of stopping in a turnaround where the last az acceleration wasn't actually 0 that could possibly have made a jagged velocity profile for the ACU to follow. It may have not been a big deal, but I wrote some code in `PointProvider.stop()` that deals calcs the last az acceleration for the final two points to make the start of the stop slightly smoother. This _does_ ignore the jerk of the last points not being 0 and sorta overcomplicates PointProvider past its initial scope probably... but it may help make stops somewhat smoother in these edge cases (especially if acceleration is high).

File: socs/agents/acu/agent.py [hunk 11]
```diff
@@ -2384,7 +2384,7 @@ def generate_scan(self, session, params):
                                     'data': scan_params_bundle})
 
         ret_val = (yield self._run_track(
-            session=session, point_gen=g, step_time=step_time, min_stop_time=min_turnaround_time,
+            session=session, point_gen=g, step_time=step_time, max_stop_accel=max_turnaround_accel,
```
mhasself: Pass in `az_accel` instead of `max_turnaround_accel`. Then the stop will be more consistent with the user's current preferred accel.

socs/agents/acu/drivers.py
```diff
@@ -134,7 +134,11 @@ def _request(self, n):
     def pop(self):
         self._request(1)
         if len(self._stash):
-            self._last_yielded_point = self._stash[0]
+            if len(self._last_yielded_points) == 2:
+                # Ensure we only ever have 2 points in the last yielded points.
```
mhasself: This is nitpicking comments but it's clearer to say "Ensure we _never have more than 2_ points..."

socs/agents/acu/drivers.py
```diff
@@ -158,13 +162,24 @@ def stop(self, free_form, stop_time=None):
         args:
             free_form (bool): Arg tracking if the scan is a done with free_form=True.
```
mhasself, line 165: Add new arg to docstring.

socs/agents/acu/drivers.py
```diff
@@ -158,13 +162,24 @@ def stop(self, free_form, stop_time=None):
...
+        self.abort()
+
+        if not len(self._last_yielded_points):
+            return  # Platform was never in motion because we never yielded a point!
 
-        final_point = self._last_yielded_point
+        final_point = self._last_yielded_points[-1]
 
+        # We'll use a standard turnaround to stop the telescope.
+        # For standard turnarounds, stop_time = 3 * v0 / max_acceleration
```
mhasself: Comment mentions factor 3 ... but that's not used below.  It does seem like if a turnaround takes 3, then a stop should only take 1.5 ... but it my (very quick) study of this, it seemed more like 1.8.  So 2 is safe, in my view.  It's just confusing to mention 3 here.

socs/agents/acu/drivers.py
```diff
@@ -158,13 +162,24 @@ def stop(self, free_form, stop_time=None):
...
+        stoptime = 2
+        if max_stop_accel is not None:
```
mhasself: Instead of these, how about:
```
if max_stop_accel is None:
    max_stop_accel = 0.5
```

socs/agents/acu/drivers.py
```diff
@@ -173,21 +188,30 @@ def stop(self, free_form, stop_time=None):
             final_el_flag = final_point.el_flag
             final_point_group_batch = 1  # Give a non-zero point_batch_group to ensure full upload.
 
-            stop_point_track = turnarounds.gen_free_form_stop(t0=final_timestamp, v0=final_az_vel,
-                                                              az0=final_az, el0=final_el, stoptime=stop_time,
+            # Calculate the final stop acceleration to make the stop smooth in case we stop in a turnaround.
+            if len(self._last_yielded_points[-1]) == 1:
```
mhasself: Remove [-1]

socs/agents/acu/turnarounds.py
```diff
@@ -70,14 +70,15 @@ def gen_turnaround(turnaround_method, t0, az0, el0, v0, turntime, az_flag, el_fl
     return turnaround_track
 
 
-def gen_free_form_stop(t0, v0, az0, el0, stoptime, az_flag, el_flag, point_group_batch, step_time=0.1):
+def gen_free_form_stop(t0, a0, v0, az0, el0, stoptime, az_flag, el_flag, point_group_batch, step_time=0.1):
```
mhasself: I think you can drop the "point_group_batch" arg, and just always set group_flag=1 in your TrackPoints.  That is designed to ensure that the points all get uploaded together.  Such should be unnecessary, but will also be harmless and probably slightly helpful, in the present case.

I just don't want copy-pasted "point_group_batch" logic to stick around in an inscrutable form where it's not necessary.

mhasself (changes requested):
A few minor requests.  I am excited to see this work on the platform!

I did test on simulator (including plotting some of the stop profiles) and it's looking really good.

socs/agents/acu/agent.py
mjrand: Ah, yeah I can do that. It may end up with the stop being somewhat slow though. For instance SAT1 scans are have v=0.8 , a=0.25 so it could take ~>3 seconds to stop rather than ~1 second with the max acceleration. If we're not worried about the time to stop I think this consistency is good.

socs/agents/acu/drivers.py
mjrand: Ah, true. I was concerned that "never have more than 2 points" is sorta not totally correct all the time because I have no guard rails for ensuring there's any corrections if we SOMEHOW end up over 2 points. I guess really if somehow `_last_yielded_points` has more than 2 points we should really pop points until we have 2.

socs/agents/acu/agent.py
mjrand: Ok rewrote this to pass in `stop_accel = min(az_accel, max_turnaround_accel)` from `generate_scan`

mjrand: actually, as I write this I realize az_accel is already checked against max_turnaround_accel. I'll just pass it straight through then instead of checking twice!

socs/agents/acu/drivers.py
mjrand: Updated the comment and the `_last_yielded_points` updated to ensure only ever 2 points or less

socs/agents/acu/drivers.py
mjrand: Done!

socs/agents/acu/drivers.py
mjrand: I took the equation for the peak turnaround accel from my documentation of in `turnarounds.py`... which may have been wrong (or I took it from the wrong place I guess. 

I think really `stop_time = abs(v0) / stop_accel` because its not going from v to -v.

socs/agents/acu/drivers.py
mjrand: Changed!

socs/agents/acu/drivers.py
mjrand: whoops! Fixed!

socs/agents/acu/turnarounds.py
mjrand: Ah, I see what you mean. I've updated the function to always set `group_flag=1`. Thanks!

socs/agents/acu/drivers.py
mjrand: I just realized we could just... put this in the default args of the function...

socs/agents/acu/drivers.py
mhasself: My point was just about the wording. In normal operation the list could be empty, have 1 point, or 2 points.

In this situation I would tend towards
```
while len(self._last_yielded_points) >= 2:
  self._last_yielded_points.pop(0)
```

That's overly cautious; but sometimes over-caution protects things from future changes.

socs/agents/acu/drivers.py
mhasself: Ok.  Just so we're clear -- now stop_accel is treated as an average accel.  The peak acceleration will be higher than this.  I think it's ok -- the same is true of turn-arounds.  And nothing in the docstring says "max" or "peak".

mhasself (approved):
This looks good to me!  I'll run a couple more tests in simulator but if you could merge in the main branch, that will trigger docker rebuild (I think last build failed on a random internet issue) and we can put this up for @BrianJKoopman sign-off.

socs/agents/acu/drivers.py
mhasself: Oh, it appears you implemented this exact thing already :)

mjrand:
Awesome, thanks for the review! I think we should test this on a real SAT once you've finished simulator testing. Just let know if it all looks good and I can commandeer a SAT to test.

BrianJKoopman:
> Awesome, thanks for the review! I think we should test this on a real SAT once you've finished simulator testing. Just let know if it all looks good and I can commandeer a SAT to test.

I'd say don't wait for me to test. And it seems like @mhasself gave his approval. Let me know how the testing goes. I'll take a look at this when I can, but also if it works and @mhasself is happy with it, then I'm happy too.

mjrand:
Ok I had to sorta change the implementation of the `free_form=False` scan stops generated by `PointProvider`. For whatever reason the ACU really did not like just duplicating the last point with `az_vel=0`. I'm GUESSING that its upset that there's something weird with the stop_time and v=0... not sure. However, I realized we could just use the stop generator we used for `free_form=True` scans and take the last point. This actually did work ok. It makes the standard user-requested stops look pretty goofy! (and maybe a little slow)... but they work!

I tested this latest commit on SATp1 with a end of planned track stop for standard and two_leg turnarounds as well as user-requested stops for standard and two_leg turnarounds. All worked ok. 

I ran real CMB scans with the `two_leg` turnarounds for over 4 hours with no issue!

mjrand:
Here's what a `free_form=False` stop looks like:
image

Here's a `free_form=True` stop
image

mhasself:
> This actually did work ok. It makes the standard user-requested stops look pretty goofy! (and maybe a little slow)... but they work!

Uh... let's not introduce any new goofiness.

Whatever was happening before, for free_form=False, let's get that back.

mjrand:
Ok, I think I have a few different things we could try to get the same behavior.

1. For free_form=False scans... just stop giving any points. This seemed to work before but it seems less than "gentle".
2. We could dupe a point with az += v0/1 second with az_vel=0 and el_vel=0. I think this is ~equivalent with how the end of track stop looks.

mjrand:
Here's what a `free_form=False` user-requested stop looks like now with these changes.
image

I think this mimics what the "end of planned track" stops were doing. I don't THINK there are any weird edge cases, I think this should also handle turnarounds ok. I think it was trying to make a new point with `az=final_point.az` and `az_vel=0` `el_vel=0` was what was killing the scan stop. I think the ACU was seeing something like " I need to fully accelerate in the other direction to make this point" and it was throwing fits. The end of the planned track sets the az_vel and el_vel of the last point to 0, it doesn't duplicate a point with the velocities set to 0 which was what I was getting errors from. This change emulates that behavior.

For what its worth, I also tested doing literally nothing for `free_form=False` stops and the ACU was perfectly happy with that as well. (i.e. just `if not free_form: pass` in `PointProvider.stop()` also worked perfectly fine). I THINK this is what the user-requested stops were doing before actually.

mhasself:
> For what its worth, I also tested doing literally nothing for `free_form=False` stops and the ACU was perfectly happy with that as well. (i.e. just `if not free_form: pass` in `PointProvider.stop()` also worked perfectly fine). I THINK this is what the user-requested stops were doing before actually.

So ya ... can you leave the `not free_from` stopping code in, but commented out?  The tests you've done on this are useful but I'm very hesitant to change the behavior of something that is working perfectly already.  We were not, to my knowledge, having any problems with user-requested stops.

mjrand:
Ok, I reverted `free_form=False` scan stops in `PointProvider.stop()` to do nothing but clear the stash and dump the generator. This is the current behavior of the agent.

PR #1055, opened 2026-05-08
ACU: use new persistent http connection in soaculib

mhasself:
## Description

Makes ACU agent use new backend.  This backend should be more stable as it opens a single http command connection to ACU and holds onto it.

Requires soaculib v0.4.3 or later; older versions will fail on startup, due to persistent=True argument.

## Motivation and Context

There is some evidence, in recent ACU communication failures, that rapid commanding (which would previously have required multiple connections) is causing the issues.  There was also some evidence that making such requests inside a twisted thread was causing some errors to not get caught properly.  The exact culprit was not found (i.e. why the thread locked up instead of calling back an error).  But in any case the new solution is what we wanted to, originally, and should be easier to debug if it shows similar comms issues.

## How Has This Been Tested?

I have run this at site on SATP3 and it works well.  So far the lockups that cause lots of trouble before have not been observed.  It's too early to say this fixed them, however.

I've also tested on simulator, where this backend works much better, as new connection overhead really seems to slow down simulator communications.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -287,7 +287,7 @@ class ACUAgent:
         tclient._HTTP11ClientFactory.noisy = False
 
         self.acu_control = aculib.AcuControl(
-            acu_config, backend=RetwistedHttpBackend(persistent=False))
+            acu_config, backend=RetwistedHttpBackend(persistent=True))
         self.acu_read = aculib.AcuControl(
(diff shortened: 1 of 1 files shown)


BrianJKoopman (approved):
Since we install `soaculib` from `master`, and this sounds like an improvement, I'm going to go ahead and merge this. If it causes issues we can revert/pin to a different tag for installation until that's worked out.

PR #1063, opened 2026-05-18
fls: Avoid publishing nan photocurrent value

laurensaunders:
- drivers: In sampling, changed 'nan' value for photocurrent to a non-nan float (999999.)
- drivers: In check_scan_params, changed handling of scan mode to remove excess '/'
- agent: Updated acq docstring in agent to reflect change to photocurrent handling

## Description
This is a bug fix for the FLS Agent acq process.

## Motivation and Context
The Influx agent was crashing due to the FLS Agent publishing a 'nan' value in its acq process.
We will handle the invalid data case by publishing an unrealistic non-nan float. Additional update
to publish a cleaner string for scan mode (remove an excess '/') when possible.

## How Has This Been Tested?
Currently untested.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/fls/drivers.py
+++ b/socs/agents/fls/drivers.py
@@ -233,10 +233,10 @@ class DLCSmart(TCPInterface):
         self.param_set("lockin:integration-time", int_time)
 
     def check_scan_params(self):
-        smode = self.param_ref("frequency:scan-mode-fast")
-        if 'fast' in smode:
+        fast = self.param_ref("frequency:scan-mode-fast")
+        if 'fast' in fast:
             fast = 'fast'
-        elif 'precise' in smode:
+        elif 'precise' in fast:
             fast = 'precise'
         smin = self.param_ref("frequency:frequency-min")
         smax = self.param_ref("frequency:frequency-max")
--- a/socs/agents/fls/drivers.py
+++ b/socs/agents/fls/drivers.py
@@ -233,7 +233,11 @@ class DLCSmart(TCPInterface):
         self.param_set("lockin:integration-time", int_time)
 
     def check_scan_params(self):
-        fast = self.param_ref("frequency:scan-mode-fast")
+        smode = self.param_ref("frequency:scan-mode-fast")
+        if 'fast' in smode:
+            fast = 'fast'
+        elif 'precise' in smode:
+            fast = 'precise'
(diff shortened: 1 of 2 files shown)


socs/agents/fls/drivers.py
```diff
@@ -234,6 +234,10 @@ def set_scan_params(self, freq_min, freq_max, freq_step, direction, int_time):
 
     def check_scan_params(self):
         fast = self.param_ref("frequency:scan-mode-fast")
+        if 'fast' in fast:
+            fast = 'fast'
+        elif 'precise' in fast:
+            fast = 'precise'
```
BrianJKoopman: > drivers: In check_scan_params, changed handling of scan mode to remove excess '/'

Is this for the `\` in the logs here?
```
2026-05-15T23:03:48+0000 InfluxDB Client Error: 400: {"error":"partial write: unable to parse 'lat.fls-dlc-smart,feed=sampling_data set_frequency=100.0,actual_frequency=99.9998544297996,photocurrent=nan,bias_voltage=1.00006103515625,bias_offset=-0.4999389573922598,lasers_on=True,scan_mode=\"fast\",scan_min_frequency=60.0,scan_max_frequency=500.0,scan_step=0.05,scan_direction=-1i,integration_time=499.9640615384615 1778886224558587136': invalid number dropped=0"}
```

That's just escaping the `"` in this print. The `\` isn't actually included in the publish call.

If that's the reason, I don't think this change is needed.

BrianJKoopman (commented):
Thanks for the change! It seems like the valid upper range for photocurrent is ~a few hundred, so this should be plenty out of the way. It might make your photocurrent plots kind of wonky in Grafana though, just a warning.

> The Influx agent was crashing due to the FLS Agent publishing a 'nan' value in its acq process.

I dug into the logs more. The InfluxDB agent wasn't actually crashing, just not publishing the FLS numbers due to the write error.

Anyway, one question below about the other change.

socs/agents/fls/drivers.py
laurensaunders: Oh okay, I had thought that it was included (we don't watch that value on grafana, so hadn't seen it before). We can remove this change then.

BrianJKoopman (approved):

PR #1072, opened 2026-06-09
Acu add sin el nod

mjrand:
Added the `gen_sin_el_nod` scan generator function into the ACU drivers.

## Description

Added a new generator function for generating sinusoidal elevation nods into the ACU agent. This generator will allow us to generate sine wave el nods that do not move in azimuth. This is basically a replica of type3 scan behavior with no azimuth movement.

## Motivation and Context

LAT ISO review requested sinusoidal elevation nods with no azimuth movement.

## How Has This Been Tested?

I've simmed the `gen_sin_el_nod` driver function output with success. The generator function should work as intended.
image

I have not yet tested the changes to the agent `generate_scan` function or tested an movement on the LAT. This needs to be tested before PR approval.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -2417,7 +2417,7 @@ class ACUAgent:
                                              'event': 2}})
         return ret_val
 
-    @ocs_agent.param('el_depth', type=float, default=None)
+    @ocs_agent.param('el_depth', type=float)
     @ocs_agent.param('el_freq', type=float, default=None)
     @ocs_agent.param('num_nods', type=int, default=None)
     @ocs_agent.param('start_time', type=float, default=None)
@@ -2428,27 +2428,27 @@ class ACUAgent:
                            num_nods=None, start_time=None, \
                            scan_upload_length=None)
 
-        **Process** - Scan generator, currently only works for
-        constant-velocity az scans with fixed elevation.
+        **Process** - Scan generator. Generates sinusoidal elevation nods
+        at the current azimuth and elevation of the platform. Nods are generated
+        with frequency equal to el_freq and amplitude equal to el_depth.
+        For a given el_depth, the platform will nod from the current elevation to
+        current elevation + el_depth.
+
+        This function publishes scan_params to the feed with scan_type = 4 .
 
         Parameters:
             el_depth (float): The number of degrees from the current
-            el to nod to. Can be negative.
-            Positive el_depth nods updwards, downwards for negative el_depth.
-            el_freq (float): frequency of the elevation nods.
+                el to nod to. Can be negative.
+                Positive el_depth nods updwards, downwards for negative el_depth.
+            el_freq (float or None): frequency of the elevation nods.
+                If None, scan_params['el_freq'] is used.
             num_nods (int or None): if not None, limits the nods to
                 the specified number of sinusoidal nods. The
                 process will exit without error once that has
                 completed.
             start_time (float or None): a unix timestamp giving the
                 time at which the scan should begin.  The default is
-                None, which means the scan will start immediately (but
-                taking into account the value of wait_to_start).
-            wait_to_start (float): number of seconds to wait before
-                starting a scan, in the case that start_time is None.
-                The default is to compute a minimum time based on the
-                scan parameters and the ACU ramp-up algorithm; this is
-                typically 5-10 seconds.
+                None, which means the scan will start immediately.
             step_time (float): time, in seconds, between points on the
                 constant-velocity parts of the motion.  The default is
                 None, which will cause an appropriate value to be
@@ -2492,15 +2492,14 @@ class ACUAgent:
         el_endpoint1, _untweaked_el = _f(el_endpoint1), el_endpoint1
         if abs(el_endpoint1 - _untweaked_el) > 0.1:
             return False, "Current elevation (%.4f) is well outside limits." % _untweaked_el
-        init_el = el_endpoint1
 
         scan_params = {k: params.get(k) for k in [
             'num_nods', 'start_time', 'step_time']
             if params.get(k) is not None}
+        step_time = scan_params['step_time']
 
         self.log.info('The scan_params: {scan_params}', scan_params=scan_params)
 
-        # Before any motion, check for sun safety. WE GOTTA MAKE THIS FOR NODS UGH
         ok, msg = self._check_nod_sunsafe(az, el_endpoint1, el_endpoint2)
         if ok:
             self.log.info('Sun safety check: {msg}', msg=msg)
@@ -2518,30 +2517,12 @@ class ACUAgent:
         if not ok:
             return False, msg
 
-        # Seek to starting position.  Note "legs" will always include
-        # at least 2 points; first point being current (az, el).
-        self.log.info(f'Moving to start position, az={az}, el={init_el}')
-        legs, msg = yield self._get_sunsafe_moves(az, init_el)
-        if msg is not None:
-            self.log.error(msg)
-            return False, msg
-        hwp_safe, msg = yield self._check_hwpsafe_legs(legs)
-        if not hwp_safe:
-            msg = f'Move to start position not permitted: {msg}'
-            self.log.info('{msg}', msg=msg)
-            return False, msg
-
         # Also validate the scan generally -- need to be movable in el.
         hwp_safe, msg = yield self._check_hwpsafe(el_endpoint1, el_endpoint2, axes=['el'])
         if not hwp_safe:
             msg = f'El-nod not permitted: {msg}'
             self.log.info(msg)
             return False, msg
-
-        for leg_az, leg_el in legs[1:]:
-            ok, msg = yield self._go_to_axes(session, az=leg_az, el=leg_el)
-            if not ok:
-                return False, f'Start position seek failed with message: {msg}'
 
         # Prepare the point generator.
         free_form = True
@@ -2576,7 +2557,7 @@ class ACUAgent:
                                     'data': scan_params_bundle})
 
         ret_val = (yield self._run_track(
-            session=session, point_gen=g, step_time=16 * el_freq,
+            session=session, point_gen=g, step_time=step_time,
             track_axes=track_axes, free_form=free_form, unabort_failure=True))
 
         self.agent.publish_to_feed('scan_params',
@@ -2691,7 +2672,7 @@ class ACUAgent:
             last_mode = None
             last_upload_az = None
             start_time = time.time()
-            got_progtrack = False
+            got_progtrack = {ax: False for ax in track_axes}
             faults = {}
             got_points_in = False
             first_upload_time = None
@@ -2713,7 +2694,7 @@ class ACUAgent:
                 # points but ACU is quietly dumping them because the
                 # vel is too high.
                 got_points_in = got_points_in \
-                    or (got_progtrack and free_positions < FULL_STACK)
+                    or (any(got_progtrack.values()) and free_positions < FULL_STACK)
 
                 if last_mode != mode:
                     self.log.info(f'scan mode={mode}, line_buffer={len(point_prov)}, track_free={free_positions}')
@@ -2733,15 +2714,15 @@ class ACUAgent:
                         # Convert between track axes names and status field names.
                         track_axes_names = {'az': 'Az', 'el': 'El'}  # There's probably a better way to convert these.
                         if current_modes[track_axes_names[ax]] == 'ProgramTrack':
-                            got_progtrack = True
+                            got_progtrack[ax] = True
                         else:
-                            if got_progtrack:
-                                self.log.warn('Unexpected exit from ProgramTrack mode!')
+                            if got_progtrack[ax]:
+                                self.log.warn(f'Unexpected exit from ProgramTrack mode on {ax} axis!')
                                 if mode == 'stop':
                                     prog_track_err = True
                                 mode = 'abort'
                             elif now - start_time > MAX_PROGTRACK_SET_TIME:
-                                self.log.warn('Failed to set ProgramTrack mode in a timely fashion.')
+                                self.log.warn(f'Failed to set ProgramTrack mode in a timely fashion on {ax} axis.')
                                 mode = 'abort'
 
                     # If we've attempted to upload points does the ACU actually have them?
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -355,6 +355,11 @@ class ACUAgent:
                                self._simple_process_stop,
                                blocking=False,
                                startup=False)
+        agent.register_process('generate_sin_el_nod',
+                               self.generate_sin_el_nod,
+                               self._simple_process_stop,
+                               blocking=False,
+                               startup=False)
         agent.register_process('idle_reset',
                                self.idle_reset,
                                self._simple_process_stop,
@@ -2073,7 +2078,7 @@ class ACUAgent:
                      choices=['end', 'mid', 'az_endpoint1', 'az_endpoint2',
                               'mid_inc', 'mid_dec'])
     @ocs_agent.param('az_drift', type=float, default=None)
-    @ocs_agent.param('scan_type', default=1, choices=[1, 2, 3, 'sin_el_nod'])
+    @ocs_agent.param('scan_type', default=1, choices=[1, 2, 3])
     @ocs_agent.param('az_vel_ref', type=float, default=None)
     @ocs_agent.param('turnaround_method', default=None,
                      choices=[None, 'standard', 'standard_gen',
@@ -2147,7 +2152,6 @@ class ACUAgent:
                 Type 1 is a constant elevation scan.
                 Type 2 includes a variation in az speed that scales as sin(az).
                 Type 3 is a Type 2 with an sinusoidal el nod.
-                'sin_el_nod' is a sinusoidal el nod with no az movement.
             az_vel_ref (float or None): azimuth to center the velocity profile at.
                 If None then the average of the endpoints is used.
             turnaround_method (str): The method used for generating turnaround.
@@ -2211,10 +2215,6 @@ class ACUAgent:
         if el_mode is None:
             el_mode = self.scan_params['el_mode']  # ... which may also be None.
 
-        # Check our az endpoints for sin_el_nods. They should be the same!
-        if params['scan_type'] == 'sin_el_nod' and az_endpoint1 != az_endpoint2:
-            raise ValueError("Cannot run type 4 scans with different az endpoints!")
-
         # Check if the turnaround method is usable for the called scan type.
         # This should never happen with the above turnaround_method setting.
         if turnaround_method == "standard" and params['scan_type'] != 1:
@@ -2266,8 +2266,6 @@ class ACUAgent:
             az_cent = az_vel_ref - 90
             az_edge = np.max(np.abs((az_endpoint1 - az_cent, az_endpoint2 - az_cent)))
             az_edge_speed = az_speed / np.sin(az_edge)
-        if params['scan_type'] == 'sin_el_nod':
-            az_edge_speed = 0
 
         plan = sh.plan_scan(az_endpoint1, az_endpoint2,
                             el=el_endpoint1, v_az=az_edge_speed, a_az=az_accel,
@@ -2381,18 +2379,8 @@ class ACUAgent:
                                        az_vel_ref=az_vel_ref,
                                        az_first_pos=plan['init_az'],
                                        **scan_params)
-
-        elif params['scan_type'] == 'sin_el_nod':
-            free_form = True
-            track_axes = ['el']
-            g = sh.generate_sin_el_nod(az=az_endpoint1,
-                                       el_endpoint1=el_endpoint1,
-                                       el_endpoint2=el_endpoint2,
-                                       el_freq=el_freq,
-                                       **scan_params)
-
         else:
-            raise ValueError("Scan type must be 1, 2, 3, or 'sin_el_nod'")
+            raise ValueError("Scan type must be 1, 2, or 3")
 
         scan_params_bundle = {'session_id': session.session_id,
                               'schema': 1,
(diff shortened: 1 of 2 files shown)


mhasself:
Thanks Michael.  I recommend that you move this to "draft" PR until we finalize design and test it on a real system.

Initial reactions:
- Why should this be a new "scan type" for generate_scan rather than a new, stand-alone process / task?  The arguments for this functionality are a tiny subset of the full az scan functionality.  It's not like someone would decide, one day, to do an el-nod everywhere that they previously were doing a scan!
- In modern application, num_batches and batch_size are deprecated. Let's not introduce new usages ... i.e., they can be removed from your driver function (and just hard-code the current default at the top of the function).
- Is there a control on step_time relative to the el_freq?  Make sure step_time gives you "enough" points.  Fourier limit is 4 but I would think we'd want more like 16.
- The generator sets group_flag=0, which means a "stop" can cause the track to terminate at ~any point. That might "work" but it would be best to tune group_flag such that stops will always occur at the bottom of the dip (which is also the starting point).
- To support the preceding point, it might be helpful to tweak step_time such that it divides the oscillation period evenly.  Then just generate 1 cycle at a time and yield it (another reason to drop the "batch_size" notion).  Can discuss more elsewhere.  But basically set `new_step_time = (el_freq*round((el_freq*step_time)**-1))**-1`.

mjrand:
Aaaah I didn't even know there was a draft PR feature... that's super helpful! (I am not great at git... lol). Changed.

1. There's really no reason other than I wanted to reuse the `generate_scan` architecture and make it really easy to call from with sorunlib. It should be somewhat easy to make its own process/task. I'll write that in. I wasn't super attached to keeping it in `generate_scan` instead of making its own process. 
2. Yeah the notes on batches makes sense to me. I can rewrite this to always yield out whole number nods. That should also prevent us from stopping the elevation on a non-zero velocity.... which opens up a question of if we need to also write a smooth-stop generator for elevation (I think we probably do), but that's outside the scope of this PR.
3. I think step_size = 0.1 is probably good enough for all nods that aren't hilariously high frequency. I think there _is_ a check for this in `plan_scan`... but it should probably just be in this function calculated like you say. 
4. Agreed we should probably yield whole number nods.
5. Agreed.

I think rewriting the driver to yield whole number of nods and then rewriting the agent process will be good. One of the worries I had with the current implementation into `generate_scan` was how `plan_scan` would interact with az_vel=0. I _think_ it would work... but I think if we make a `generate_el_nod` function or whatever it could just start at the current az instead of having to run-up to `az_vel=0`. I'll try this implementation and see how it would work.

mjrand:
Alright I reverted `generate_scan` and made `generate_sin_el_nod`. I started it as a copy of `generate_scan` and then tried to prune back as many things weren't used or weren't necessary. This meant pruning `plan_scan` and some other misc variables like `wait_to_start` (which I couldn't find used in `generate_scan` either...). I also fixed up the `generate_sin_el_nod` drivers function to only yield full el nods and to always divide them into 16 points (which can be changed...). This means step_time is determined in the driver and doesn't need to be callable in the agent.

socs/agents/acu/agent.py
```diff
@@ -355,6 +355,11 @@ def __init__(self, agent, acu_config='guess',
                                self._simple_process_stop,
                                blocking=False,
                                startup=False)
+        agent.register_process('generate_sin_el_nod',
```
mhasself: Let's just call this "generate_elnod".  If we invent alternatives, we'll likely jam them into this function.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,172 @@ def generate_scan(self, session, params):
                                              'event': 2}})
         return ret_val
 
+    @ocs_agent.param('az', type=float)
+    @ocs_agent.param('el_endpoint1', type=float, default=None)
+    @ocs_agent.param('el_endpoint2', type=float, default=None)
```
mhasself: Instead of having these as args, let's just make this function run them at the current position.  This makes for less bookkeeping in sorunlib.  And then you'd just have "el_depth", which could be negative, as the parameter that determines how far apart the endpoints are.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,172 @@ def generate_scan(self, session, params):
...
+                              'schema': 1,
+                              'event': 1,
+                              'init_time': init_time,
+                              }
+        scan_params_bundle.update({
+            'az1': az,
+            'el1': el_endpoint1,
+            'el2': el_endpoint2,
+            'el_freq': el_freq,
+            'type': 'sin_el_nod',
```
mhasself: The data here need to be exactly the same (same fields with same data types) as the ones in generate_scan.  Otherwise the feed will complain (or, at best, data loading of this info will be totally messed up).  I think you should indeed use 4 for the sin_el_nod -- and register it in the main list of scan types.  Any irrelevant stuff should be set to 0 or nan or something like that.

socs/agents/acu/agent.py
```diff
@@ -3188,6 +3359,30 @@ def _check_scan_sunsafe(self, az1, az2, el, v_az, a_az):
...
+        planned El-nod seems to currently be sun-safe.
+
+        """
+        if not self._get_sun_policy('sunsafe_moves'):
+            return True, 'Sun-safety checking is not enabled.'
+
+        if not self._get_sun_policy('map_valid'):
+            return False, 'Sun Safety Map not computed or stale; run the monitor_sun process.'
+
+        n = max(2, int(np.ceil((el2 - el1) / 1.)))
```
mhasself: humor me and throw an "abs" in front of that (el2 - el1) ...

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,95 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
...
+    Yields:
+        points (list): a list of TrackPoint objects.  Raises
+          StopIteration once exit condition, if defined, is met.
+
+    """
+    POINTS_PER_NOD = 16
+    MIN_STEP_TIME = 0.05  # in seconds.
+
+    # Get el throw
+    el_throw = abs(el_endpoint2 - el_endpoint1) / 2
```
mhasself: I think we want to allow downward nods; so remove the abs().

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,95 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
...
+    t = 0
+    el = el_endpoint1
+
+    # We want to divide each nod into at least 16 equadistant points.
+    # This gives a max frequency of el_freq < 1/(0.05*16) -> el_freq < 1.25.
+    if el_freq >= 1 / (POINTS_PER_NOD * MIN_STEP_TIME):
+        raise ValueError("El Freq is too high! El Freq must be < 1.25Hz to keep "
+                         "step_time > 0.05!")
+
+    step_time = POINTS_PER_NOD * el_freq  # Divide nod into 16 points
```
mhasself: Ya I don't think we should force step_time like this.  Rather, respect what the user asked for but:
- round it, so it evenly divides the nod period. 
- make sure there are at least 16 points per period.

That's what I described in previous comment.  We'll likely need to experiment with this a little to see what step_times work well.

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,95 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
...
+
+    while check_completed_nods():
+        point_block = []
+        for j in range(POINTS_PER_NOD):
+            el, el_vel = get_el(t + t0)
+            point_block.append(TrackPoint(
+                timestamp=t + t0,
+                az=az, el=el, az_vel=0, el_vel=el_vel,
+                az_flag=0, el_flag=1,
+                group_flag=0))
```
mhasself: set group_flag=1 except for the final point in each cycle -- that is what will actually force only complete cycles to be uploaded.

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,95 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
...
+
+            t += step_time
```
mhasself: Consider restructuring the loop like this:
```
# Templates for one cycle
template_t = step_time * np.arange(points_per_nod)
template_el, template_el_vel = get_el(template_t)  # vectors

while not done():
   point_block = [PointBlock(timestamp=t0 + t, ..., group_flag=1)
                       for (t, el, el_v) in zip(template_t, template_el, template_el_vel)]
   point_block[0].group_flag = 0  # Ok to stop on the zero vel starting point!
   yield point_block
   t0 += 1 / el_freq

# And one last zero-vel point.
yield [PointBlock(timestamp=t0, ... vels=0, group_flag=0)]
```

This reduces floating point errors, but also just makes it crystal clear that we're doing one cycle at a time, and they're always identical in el shape.

mhasself (commented):
Thanks! More notes ...

socs/agents/acu/agent.py
mjrand: Done. Now just called `generate_el_nod` in the agent. I kept the `generate_sin_el_nod` in the driver

socs/agents/acu/agent.py
mjrand: Done, changed to `el_freq` and `el_depth` as callable parameters

socs/agents/acu/agent.py
mjrand: Copied the format from `generate_scan` and filled it with the available data. 

Do you mean register scan_type 4 in the agent? Like in an enum? or is there another place to register the scan_type?

socs/agents/acu/agent.py
mjrand: Done!

socs/agents/acu/drivers.py
mjrand: removed!

socs/agents/acu/drivers.py
mjrand: Reworked the step_time in this function. It'll now take the step_time and find the closest step_time that divides the number of points per nod into the closest multiple of four (so that we get clean looking, regular sinewaves).

socs/agents/acu/drivers.py
mjrand: Done

socs/agents/acu/drivers.py
mjrand: Reworked the loop to make the cycle more obvious.

mjrand:
I had to change part of the `_run_track` function to get el nods to work properly. The previous iteration was _only_ and explicitly checking that the azimuth axis was in ProgramTrack mode. This meant that these el nods that only put the el axis into `track_axes` was failing because the az axis was in in stop mode. Now `_run_track` should check that every axis in `track_axes` (which are the axes we expect to be in ProgramTrack mode) are in ProgramTrack mode. This should benefit type2 and 3 scans too which _should_ be checking that the el axis didn't fail but currently don't.

I also added some comments into the abort checks to make it clear what we're checking for abort mode.

mjrand:
I've tested this current commit on satp5 and generated a few nods with varying `num_nods` and it worked as expected. Here is `num_nods=10` `el_freq=0.1` , `el_depth=-3` starting at el=60.

image

The free upload positions plot shows that we're uploading full nods at once too:
image

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
...
+                completed.
+            start_time (float or None): a unix timestamp giving the
+                time at which the scan should begin.  The default is
+                None, which means the scan will start immediately (but
+                taking into account the value of wait_to_start).
+            wait_to_start (float): number of seconds to wait before
+                starting a scan, in the case that start_time is None.
+                The default is to compute a minimum time based on the
+                scan parameters and the ACU ramp-up algorithm; this is
+                typically 5-10 seconds.
```
mhasself: not a declared arg

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
                                              'event': 2}})
         return ret_val
 
+    @ocs_agent.param('el_depth', type=float, default=None)
```
mhasself: remove `default=None` -- this param isn't optional.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
...
+                           num_nods=None, start_time=None, \
+                           scan_upload_length=None)
+
+        **Process** - Scan generator, currently only works for
+        constant-velocity az scans with fixed elevation.
+
+        Parameters:
+            el_depth (float): The number of degrees from the current
+            el to nod to. Can be negative.
+            Positive el_depth nods updwards, downwards for negative el_depth.
```
mhasself: fix indentation.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
...
+            el_freq (float): frequency of the elevation nods.
```
mhasself: float or None. "If None, scan_params['el_freq'] is used."

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
...
+            return False, "Current elevation (%.4f) is well outside limits." % _untweaked_el
+        init_el = el_endpoint1
+
+        scan_params = {k: params.get(k) for k in [
+            'num_nods', 'start_time', 'step_time']
+            if params.get(k) is not None}
+
+        self.log.info('The scan_params: {scan_params}', scan_params=scan_params)
+
+        # Before any motion, check for sun safety. WE GOTTA MAKE THIS FOR NODS UGH
```
mhasself: Clean up the comment

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
...
+        self.log.info('Clearing faults to prepare for motion.')
+        yield self.acu_control.clear_faults()
+        yield dsleep(1)
+
+        # Verify we're good to move
+        ok, msg = yield self._check_ready_motion(session)
+        if not ok:
+            return False, msg
+
+        # Seek to starting position.  Note "legs" will always include
```
mhasself: I don't think you have to do this.  The starting position is the current position.

socs/agents/acu/agent.py
```diff
@@ -2551,25 +2725,37 @@ def _run_track(self, session, point_gen, step_time, stop_accel=0.5, track_axes=[
...
-                            mode = 'abort'
+                    # Reasons we might decide to abort or stop ...
+
+                    # First we'll check that each axis we expect to be in ProgramTrack mode,
+                    # is indeed still in ProgramTrack mode.
+                    for ax in track_axes:
+                        # Convert between track axes names and status field names.
+                        track_axes_names = {'az': 'Az', 'el': 'El'}  # There's probably a better way to convert these.
+                        if current_modes[track_axes_names[ax]] == 'ProgramTrack':
+                            got_progtrack = True
```
mhasself: I think we also need separate got_progtrack vars, for each tracked axis.  I.e. `got_progtrack[ax] = True`.  I forget what else this is used for... maybe it gets complicated.  But I'm not comfortable assuming that, if more than 1 axis is in program track mode, then they all switch to progtrack mode at the exact same moment.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,175 @@ def generate_scan(self, session, params):
...
+
+        scan_params_bundle.update({
+            'az1': az,
+            'az2': az,
+            'az_vel': 0,
+            'az_accel': 0,
+            'el1': el_endpoint1,
+            'el2': el_endpoint2,
+            'el_freq': el_freq,
+            'type': 4,
```
mhasself, line 2549: Ah ... I guess we don't have an enum for "scan_type".  Please add to docstring that this function publishes to scan_params feed with scan type=4.

mhasself (changes requested):
Thanks -- those plots are quite convincing!  Looking good but a few comments.

socs/agents/acu/agent.py
mjrand: I've removed it because I can't seem to see if its used for anything beyond the initial plan and move to scan start (which isn't used in this el nod function).

socs/agents/acu/agent.py
mjrand: removed.

socs/agents/acu/agent.py
mjrand: Fixed!

socs/agents/acu/agent.py
mjrand: Added

socs/agents/acu/agent.py
mjrand: Whoops! removed

socs/agents/acu/agent.py
mjrand: I've removed everything in this function that calculated and executed the initial legs

socs/agents/acu/agent.py
mjrand: I've updated all the `got_progtrack` var calls to now be a dict with keys equal to the values in `track_axes`. Every `got_progtrack` check now checks the value for the axis.

`got_points_in` now checks if any axis has been set to ProgramTrackMode (A valid assumption if we're just checking we've given points to the ACU for ANY axis I think).
```
# Use this var to detect case where we're uploading
# points but ACU is quietly dumping them because the
# vel is too high.
got_points_in = got_points_in \
    or (any(got_progtrack.values()) and free_positions < FULL_STACK)
```

checks for exit from ProgramTrack Mode are done axis by axis:
```
for ax in track_axes:
        # Convert between track axes names and status field names.
        track_axes_names = {'az': 'Az', 'el': 'El'}  # There's probably a better way to convert these.
        if current_modes[track_axes_names[ax]] == 'ProgramTrack':
            got_progtrack[ax] = True
        else:
            if got_progtrack[ax]:
                self.log.warn(f'Unexpected exit from ProgramTrack mode on {ax} axis!')
                if mode == 'stop':
```

mhasself: Well done -- thanks.

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,112 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
...
+    points_per_nod = round(points_per_nod / 4) * 4
+
+    # Ensure we have the minumum number of points
+    if points_per_nod < MIN_POINTS_PER_NOD:
+        points_per_nod = MIN_POINTS_PER_NOD
+
+    step_time = nod_period / points_per_nod  # Divide nod into points_per_nod points.
+
+    def check_completed_nods():
+        return num_nods is None or t * el_freq < num_nods
```
mhasself, line 1074: This is vulnerable to rounding error.  Instead of using t, just count the nods and check that.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,156 @@ def generate_scan(self, session, params):
...
+                constant-velocity parts of the motion.  The default is
+                None, which will cause an appropriate value to be
+                chosen automatically (typically 0.1 to 1.0).
+
+        Notes:
+          Note that all parameters are optional except for
+          el_depth and el_freq.  If only those two parameters
+          are passed, the Process will nod with that depth and frequency,
+          with the azimuth axis held in Stop, indefinitely (until
+          Process .stop method is called)..
```
mhasself, line 2462: In fact only el_depth is mandatory.  The indefinitely bit is accurately communicated through the arg num_nods.  The write-up here makes it sound like "will nod ... held in Stop" is only the case if one only passes those two parameters.  Which is not the case.  I think you should just mention here that the az axis is held in Stop.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,156 @@ def generate_scan(self, session, params):
...
+        el_endpoint1 = el
+        el_endpoint2 = el + el_depth
+
+        # Params with defaults configured ...
+        el_freq = params['el_freq']
+        if el_freq is None:
+            el_freq = self.scan_params['el_freq']
+
+        if el_depth == 0:
+            raise ValueError("El depth must not be equal to 0 for el nod!")
```
mhasself, line 2486: Why not?  If zero is not ok ... is 0.0001 ok?  Replace with a minimum depth.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,156 @@ def generate_scan(self, session, params):
...
+
+        # Could probably use a condition for if the el endpoints are too close?
```
mhasself, line 2488: Yes, see above...

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,156 @@ def generate_scan(self, session, params):
...
+                the specified number of sinusoidal nods. The
+                process will exit without error once that has
+                completed.
+            start_time (float or None): a unix timestamp giving the
+                time at which the scan should begin.  The default is
+                None, which means the scan will start immediately.
+            step_time (float): time, in seconds, between points on the
+                constant-velocity parts of the motion.  The default is
+                None, which will cause an appropriate value to be
+                chosen automatically (typically 0.1 to 1.0).
```
mhasself, line 2455: Default is not None.

socs/agents/acu/agent.py
```diff
@@ -2412,6 +2417,156 @@ def generate_scan(self, session, params):
...
+        # If requested el is just outside acceptable range, tweak it in.
+        _f, _ = self._get_limit_func('elevation')
+        el_endpoint1, _untweaked_el = _f(el_endpoint1), el_endpoint1
+        if abs(el_endpoint1 - _untweaked_el) > 0.1:
+            return False, "Current elevation (%.4f) is well outside limits." % _untweaked_el
+
+        scan_params = {k: params.get(k) for k in [
+            'num_nods', 'start_time', 'step_time']
+            if params.get(k) is not None}
+        step_time = scan_params['step_time']
```
mhasself, line 2499: Careful here -- if indeed params['step_time'] could be None, then this bit will fail with KeyError.

socs/agents/acu/agent.py
```diff
@@ -2551,25 +2706,37 @@ def _run_track(self, session, point_gen, step_time, stop_accel=0.5, track_axes=[
...
-                            self.log.warn('Failed to set ProgramTrack mode in a timely fashion.')
-                            mode = 'abort'
+                    # Reasons we might decide to abort or stop ...
+
+                    # First we'll check that each axis we expect to be in ProgramTrack mode,
+                    # is indeed still in ProgramTrack mode.
+                    for ax in track_axes:
+                        # Convert between track axes names and status field names.
+                        track_axes_names = {'az': 'Az', 'el': 'El'}  # There's probably a better way to convert these.
+                        if current_modes[track_axes_names[ax]] == 'ProgramTrack':
```
mhasself, line 2716: You could just use `current_modes[ax.capitalize()]`.

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,112 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
...
 
 
+def generate_sin_el_nod(az, el_endpoint1, el_endpoint2,
+                        el_freq=.15,
+                        num_batches=None,
+                        num_nods=None,
+                        start_time=None,
+                        wait_to_start=10.,
+                        step_time=0.05,
+                        batch_size=500):
```
mhasself, line 1009: drop batch_size arg

socs/agents/acu/drivers.py
```diff
@@ -999,6 +999,112 @@ def generate_type2_scan(az_endpoint1, az_endpoint2, az_speed,
                                turnaround_method=turnaround_method)
 
 
+def generate_sin_el_nod(az, el_endpoint1, el_endpoint2,
+                        el_freq=.15,
+                        num_batches=None,
```
mhasself, line 1004: drop num_batches arg

mhasself (changes requested):
Last batch of comments (?).  Mostly cosmetic but not all!

socs/agents/acu/drivers.py
mjrand: Changed! Now each loop of generating a nod counts with a `num_completed_nod` variable

socs/agents/acu/agent.py
mjrand: Fixed up wording to be a bit more clear w.r.t. the az axis. Removed reference to el_freq being mandatory.

socs/agents/acu/agent.py
mjrand: Replaced. I chose a super arbitrary value of `abs(el_depth) < 0.1` as the condition. I'm not really sure what the real physical minimum should be. Maybe values less than 0.1 work or maybe values a bit higher than 0.1 don't work? I'm not sure without testing. Maybe we need to check this.

socs/agents/acu/agent.py
mjrand: Removed comment as it is resolved by changing the condition

socs/agents/acu/agent.py
mjrand: Fixed. Default is properly commented as 0.1 seconds

socs/agents/acu/agent.py
mjrand: changed to `step_time = 0.1 if 'step_time' not in scan_params.keys() else scan_params['step_time']`

socs/agents/acu/agent.py
mjrand: Fixed!

socs/agents/acu/drivers.py
mjrand: Removed

socs/agents/acu/drivers.py
mjrand: Removed

mjrand:
Sorry I fell behind on this for a month due to some personal complications. I've resolved many of the lingering comments. One comment in particular about the minimum el_depth may need some further thinking. I set the minimum el_depth to 0.1 degrees rather arbitrarily. We may need to test what the minimum el_depth really is on the platforms (or just pick a value we know no one will ever try to nod smaller than...)

PR #1074, opened 2026-06-09
hwp-supervisor: Add option to disable shutdown

ykyohei:
## Description
Added a option to disable shutdown on startup
Added a task to disable and re-enable shutdown non-temporarily.

## Motivation and Context
Having these option is helpful for hwp expert to operate in irregular state.
For example, sometimes lakeshore or ups could have communication issue and we cannot monitor YBCO temperature or ups state, but hwp could be safe to operate.

## How Has This Been Tested?
Tested at satp3 on June 9th 2026.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/hwp_supervisor/agent.py
+++ b/socs/agents/hwp_supervisor/agent.py
@@ -1682,6 +1682,8 @@ class HWPSupervisor:
             session.data['actions'] = {
                 'pmx': self.hwp_state.pmx_action,
                 'gripper': self.hwp_state.gripper_action,
+                'shutdown_mode': self.shutdown_mode,
+                'shutdown_enabled': self.shutdown_enabled,
             }
 
             action = self.hwp_state.pmx_action
@@ -2168,7 +2170,7 @@ class HWPSupervisor:
         if params['enable'] is not None:
             self.shutdown_enabled = params['enable']
 
-        return True, 'Params updated.'
+        return True, f'Shutdown parameters updated: {self.shutdown_enabled}'
 
     def cancel_shutdown(self, session, params):
         """cancel_shutdown()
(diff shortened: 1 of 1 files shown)


mhasself (commented):
Definitely a useful thing.  Can you please add (if not already in there):

- the "shutdown_enabled" value should appear in session.data.
- the update_shutdown task logs what value the user passed in (I don't think we'll see it in the logs, otherwise).

Also, is there a feed where this can be published?  I vaguely remember that HWPSupervisor doesn't record much about the decisions it's making; consider starting this.

ykyohei:
@mhasself Thanks for the suggestions, updated as suggested and tested in daq-dev.
We recently updated hwp_supervisor to publish shutdown state so I also added "shutdown_enabled" as a new feed to publish.
Made similar updates to another PR of hwp-gripper as well https://github.com/simonsobs/socs/pull/1075

mhasself (approved):
Looks good to me -- one last minor thing:

If a user calls "update_shutdown" to set enable=false, that will *not* abort an active shutdown, right?  User needs to also call cancel_shutdown. That fact should be added to the Task docstring.

Sorry for the delay in getting back to this.

ykyohei:
@BrianJKoopman I'm not sure why the build is keep failing, it looks like it is related to latest pysmurf update? Could you take a look?

BrianJKoopman:
> @BrianJKoopman I'm not sure why the build is keep failing, it looks like it is related to latest pysmurf update? Could you take a look?

Thanks for the ping. This should fix things: https://github.com/simonsobs/socs/pull/1081 -- I expect to merge that when the build is complete in ~10 minutes. Then if you merge in the latest from `main` you should be good.

PR #1083, opened 2026-07-10
Create UKKID Controller agent

wenlockedge:
Creating a pull request as request by Brian Koopman on 10 July 2026.

## Types of changes

## Checklist:


socs/agents/ukkid_controller/example_full_band_sweep.csv
BrianJKoopman, line 1: Let's not commit a 49 MB `.csv` file here. That ~doubles the size of cloning this entire repo.

socs/agents/ukkid_controller/example_narrow_band_sweep.csv
BrianJKoopman, line 1: Drop this one too.

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+## Installation of prerequisites
+
+ On a standard Ubuntu linux installation (at the time of writing
+ Ubuntu 22.04 LTS) the following prerequisites are required.
+
+1. An Observatory Control System (OCS) installation.
+
+2. souk_readout_tools. This is the python code for the readout client
+described above (note this repository also contains the code for the
+readout server software that runs on the RFSoCs). **Use branch main**
```
BrianJKoopman, line 56: The "Use branch main" conflicts with "Be sure to select branch dev/1.3.0 from the pull down menu on github before cloning this repository." below. Which branch is recommended?

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+cd socs/socs/agents/ukkid_controller/
+ls
+```
+
+This will show the python code ukkid_controller.py, this README.md and various support files.
+
+## Making the UKKID_controller.py agent available to your OCS install
+
+The way I am doing this (there may be a more elegant way) is to create a soflink from the default "agents" directory in your default OCS installed to point at the
+ukkid_controller directory within the SOCS directory tree. On my system I had installed ocs in /home/leechj/ocs and socs in /home/leechj/socs, so I did.
```
BrianJKoopman, line 328: Yes, there is a more elegant way. `ocs` is capable of launching any agent registered with a given 'ocs plugin'. `socs` is an ocs plugin. All agents within `socs` are 'registered' in the file [`plugin.py`](https://github.com/simonsobs/socs/blob/main/socs/plugin.py).

So a few recommendations to match the standards already in place in `socs` are:
1. Rename `ukkid_controller.py` to `agent.py`. (Just for consistency really.)
2. Add a line in [`plugin.py`](https://github.com/simonsobs/socs/blob/main/socs/plugin.py) for the `UKKIDController`.
3. Create an `__init__.py` file in `socs/socs/agents/ukkid_controller/`. This makes a [regular package](https://docs.python.org/3/reference/import.html#regular-packages) that can be imported. This'll be a blank file.

Once you do this, remove this whole "Making the UKKID_controller.py agent available to your OCS install" section.

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+
+2. souk_readout_tools. This is the python code for the readout client
+described above (note this repository also contains the code for the
+readout server software that runs on the RFSoCs). **Use branch main**
+for compatibility with the ukkid_controller agent.
+
+3. The ukkid_controller agent itself - i.e. everything within the
+agents/UKKIDcontroller directory. This can be obtained by checking out
+the "ukkid_agent" branch of the official "socs" Simons Observatory
+repository.
```
BrianJKoopman, line 62: This might make sense for a guide on Confluence or similar, but 1 and 3 don't really make sense to include in this context. `ocs` is a requirement for `socs` (i.e. it gets installed by the `socs` installation automatically. And the `ukkid_controller` agent is getting packaged into `socs` -- of course it's included.

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+eb07461422d0   influxdb:1.7                    "/entrypoint.sh infl…"   12 days ago   Up 12 days   0.0.0.0:8086->8086/tcp, [::]:8086->8086/tcp    influxdb
+
+```
+
+The ocs-crossbar agent is essential for the OCS client-to-agent
+communication, so nothing will work without this. There is no
+communication between the ukkidcontroller agent and the influxdB or
+grafana agents, at the moment (although these will eventually be made
+to subscribe to the ukkidcontroller agent's feed and display
+diagnostic information.)
```
BrianJKoopman, line 102: Again this probably makes sense as a guide on Confluence, but doesn't really fit here. We can't have a guide for installing ocs within the documentation for every agent.

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+
+One should follow the "quick-start" installation instructions here...
+
+https://github.com/sr-cdf/souk_readout_tools
+
+**Be sure to select branch dev/1.3.0 from the pull down menu on github before cloning this repository.**
+
+### ukkid_controller agent
+
+For installation, see "Obtaining the UKKID_controller.py agent" section below.
```
BrianJKoopman, line 114: Drop this section too.

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+the hostname that gets returned by ```echo $HOSTNAME```, and the same
+thing with ```-docker``` appended. This section
+
+```
+{'agent-class': 'UKKIDController',
+       'instance-id': 'UKKIDController1',
+       'arguments': ['--mode', 'check_state',
+                     '--kid_stream_id', 'ufm_kid1']},
+      ]
+```
```
BrianJKoopman, line 210: This is a good section to keep -- there's a section in the reference pages for this kind of block, giving an example for how to configure the described agent. Check out [this page](https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html#agent-reference-pages) of the ocs docs.

Everything else in this section can go.

socs/agents/ukkid_controller/README.md
```diff
@@ -0,0 +1,571 @@
...
+Select the branch "ukkid_agent" from the pull down branch menu.
+Clone this branch onto the same machine as you have souk_readout_tools and your ocs system installed. The agent can be
+found via
+
+```
+cd socs/socs/agents/ukkid_controller/
+ls
+```
+
+This will show the python code ukkid_controller.py, this README.md and various support files.
```
BrianJKoopman, line 323: Drop this section. We're packaging the controller agent in the `socs` package, we don't need to fetch it separately to install.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+##################
+# Set to where you installed souk_readout_tools
+souk_readout_tools_install_dir = '/home/leechj/souk_readout_tools'
+#############
+# Set to True to run without RFSoc hardware i.e. in simulated mode for testing.
+mock = True
+#############
+# Where the site packages have been installed by the python virtual enivroment - should contain a dir called souk_readout_tools
+souk_readout_tools_path = os.path.join(souk_readout_tools_install_dir, 'client_venv/lib/python3.10/site-packages/')
```
BrianJKoopman, line 25: This is unnecessary -- the `pip install` step from the `souk_readout_tools` package puts this in your `PYTHONPATH`, so you should be able to import it without this path manipulation.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+# Directory where the config files specified in the config_file_dict dictionary below are located.
+config_file_root_dir = souk_readout_tools_install_dir
+# This next variable tells the agent which config file to push to the server depending on which kid_stream_id the agent was started with on the command line.
+# This will eventually have 28 entries for the 28 individual reduction pipelines.
+# For single pipline running, with one ukkid_controller.py agent running at a time, we only need the first entry, and it should the filename of
+# your souk_readout_tools config .yaml file which lives in config_file_root_dir specified above.
+config_file_dict = {'ufm_kid1': 'my_config_bun_p0.yaml', 'ufm_kid2': 'my_config_bun_p1.yaml'}
+##################
+
+sys.path.append(souk_readout_tools_path)
```
BrianJKoopman, line 39: See above note, but drop this path manipulation.

socs/agents/ukkid_controller/example_full_band_sweep.png
BrianJKoopman, line 1: If this isn't going to be included in the documentation somehow, let's also not commit image files here.

socs/agents/ukkid_controller/example_narrow_band_sweep.png
BrianJKoopman, line 1: If this isn't going to be included in the documentation somehow, let's also not commit image files here.

socs/agents/ukkid_controller/README.md
BrianJKoopman, line 1: Generally this is great! I've made a bunch of suggestions inline below on how to pare down the text here, as much of it is redundant with guides elsewhere in the docs.

It would also be great to convert this to an `.rst` file that Sphinx can process and incorporate into the readthedocs page. That'll live under `socs/docs/agents/` with the other agent reference pages. There are docs in ocs for standardized sections -- [check it out here](https://ocs.readthedocs.io/en/main/developer/agent_references/documentation.html#agent-reference-pages).

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+import txaio
+from ocs import ocs_agent, site_config
+from ocs.ocs_twisted import TimeoutLock
+
+##################
+# Set to where you installed souk_readout_tools
+souk_readout_tools_install_dir = '/home/leechj/souk_readout_tools'
+#############
+# Set to True to run without RFSoc hardware i.e. in simulated mode for testing.
+mock = True
```
BrianJKoopman, line 22: This should be configurable without needing to come here and modify this variable within the module. You should add it as an argparse argument, and probably default to `False`. Users who want to run in mock mode can then set it in their ocs site config file.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+#############
+# Where the site packages have been installed by the python virtual enivroment - should contain a dir called souk_readout_tools
+souk_readout_tools_path = os.path.join(souk_readout_tools_install_dir, 'client_venv/lib/python3.10/site-packages/')
+# Should contain the script client_scripts/receive_stream_g3.py
+stream_script_path = os.path.join(souk_readout_tools_install_dir, 'src/souk_readout_tools/client/')
```
BrianJKoopman, line 27: This is in the `souk_readout_tools` repot, but the `client_scripts/receive_stream_g3.py` also contains some unique path manipulation that needs to be removed:
https://github.com/sr-cdf/souk_readout_tools/blob/8901e45f022aa71b61bb6febd7f35f8bca3722d3/src/souk_readout_tools/client/client_scripts/receive_stream_g3.py#L12-L13

It seems like it might just be for the sake of the next import line:
```
from souk_readout_tools.client.readout_client import ReadoutClient
```

This shouldn't be necessary if `souk_readout_tools` was installed with `pip`. It'll be in the path and be importable.

That said, we should also consider how this is packaged in general. You could automatically determine the full path to this file, as installed by `souk_readout_tools` via:
```
>>> import souk_readout_tools.client.client_scripts.receive_stream_g3
>>> souk_readout_tools.client.client_scripts.receive_stream_g3.__file__
'/home/koopman/git/socs/.venv-ukkid/lib/python3.11/site-packages/souk_readout_tools/client/client_scripts/receive_stream_g3.py'
```

This would be better than requiring the user to configure this path, whether in this module directly or in an argument to the agent.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+# Output directory where you want your data files to get written to
+data_out_dir = os.path.join(souk_readout_tools_install_dir, 'src/souk_readout_tools/client/client_scripts/tmp/')
```
BrianJKoopman, line 29: You should make this configurable via an argument too.

I don't know where it gets set ultimately, but on the SMuRF systems they output to `/data`. Those agents all run in containers, and it's easy to map any directory onto `/data` within the container, so we don't ever change it. That said, since the SMuRF servers are dedicated to this readout, we just use `/data` on the host system too.

I know the context for this agent is on the host, but it should (in theory) run in a container running from the `socs` base image once this is merged, so a similar thing is possible here.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+# Directory where the config files specified in the config_file_dict dictionary below are located.
+config_file_root_dir = souk_readout_tools_install_dir
```
BrianJKoopman, line 31: For this we likely want to use a subdirectory of `$OCS_CONFIG_DIR`. On the SMuRF systems we use `$OCS_CONFIG_DIR/<hostname>/` as the equivalent. This is also aliased as `$SMURF_CONFIG_DIR`. Could do something similar here.

Are these config files static? Or does the system automatically write updates to them? (We have both types in the SMuRF system. The static ones we commit to https://github.com/simonsobs/ocs-deployment-configs. The dynamic ones aren't committed.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+# This next variable tells the agent which config file to push to the server depending on which kid_stream_id the agent was started with on the command line.
+# This will eventually have 28 entries for the 28 individual reduction pipelines.
+# For single pipline running, with one ukkid_controller.py agent running at a time, we only need the first entry, and it should the filename of
+# your souk_readout_tools config .yaml file which lives in config_file_root_dir specified above.
+config_file_dict = {'ufm_kid1': 'my_config_bun_p0.yaml', 'ufm_kid2': 'my_config_bun_p1.yaml'}
```
BrianJKoopman, line 36: Can you describe this more? Does each instance of the UKKID controller agent need this full list? Are all config files present and this is only selecting the on the agent needs? How often are these files changed?

Naively to me it seems like this sort of potentially changing information should go in a standard config file and the agent should be able to read that config file, rather than hardcoding these values in the agent. For instance, like the SMuRF [`sys_config.yml`](https://github.com/simonsobs/ocs-deployment-configs/blob/main/satp1/smurf-so3-satp1/sys_config.yml) files.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+from os import environ
+
+import numpy as np
+import txaio
+from ocs import ocs_agent, site_config
+from ocs.ocs_twisted import TimeoutLock
+
+##################
+# Set to where you installed souk_readout_tools
+souk_readout_tools_install_dir = '/home/leechj/souk_readout_tools'
```
BrianJKoopman, line 19: Once all of the below is addressed, this shouldn't be needed anymore.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+               p=self.client.get_sweep_progress()
+               self.log.info(f'full_band_sweep: Sweep progress: {100*p:.3f}%')
+               if p==1.0: break
+               else: time.sleep(1.0)
+
+            #################################
+            # NEW VER
+            # TO HERE
+            #################################
+            '''
```
BrianJKoopman, line 1045: What's this large commented out code block? Is it just old (and so we can drop it)? Or is it needed for the non-mocked case?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            self.log.info('get_tone_powers: complete.')
+            # JL write something to the feed here?
+            return True, 'get_tone_powers complete.'
+
+
+######################################################################
+# JL: We may not need command line arguments (other than kid_stream_id) when starting this agent
+# But I have left the infrasructure in at the moment (a simple --mode switch)
+# This should always be set to "check_state" in order to start up default "check_state" process.
+######################################################################
```
BrianJKoopman, line 1725: These argument are how things are passed through via the ocs site config file. I'd recommend keeping them for any configuration we need to do, i.e. for the paths you're hardcoding at the head of this module.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+    pgroup = parser_in.add_argument_group('Agent Options')
+    pgroup.add_argument('--mode', type=str, default='check_state',
+                        choices=['check_state', 'idle'],  # Command line params - either idle (does nothing) or start up the check_state process
+                        help="Starting action for the Agent.")
+    # Equivalent of stream_id in the SMURF world. Each option corresponds to a specific RFSoc board connected to a specific
+    # sub-array of KID detectors in the focal plane.
+    # Plan would be to have 7 instances per SAT of the ukkid_controller agent running
+    # Each with a different kid_stream_id.
+    pgroup.add_argument('--kid_stream_id', type=str, default='ufm_kid1',
+                        choices=['ufm_kid1', 'ufm_kid2', 'ufm_kid3', 'ufm_kid4', 'ufm_kid5', 'ufm_kid6', 'ufm_kid7', 'ufm_kid8'],  # Command line params - either idle (does nothing) check_state or stream process
```
BrianJKoopman, line 1739: Remove the duplicated comment at the end of the line here.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            os.makedirs(self.top_level_output_dir)
+
+        # Register OCS feed
+        # JL: Need to register feed names here.
+        # Not sure what, if anything, needs to go here
+        # just make one called UKKID_feed for the time being
+        agg_params = {
+            'frame_length': 10 * 60  # [sec]
+        }
+        self.agent.register_feed('UKKID_feed',
```
BrianJKoopman, line 92: Feed names should only use lowercase letters, numbers, and underscores. So let's change this to `ukkid` or similar.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+        None.
+
+        Notes
+        -------
+        The following data will be written to the session.data object::
+
+            >> response.session['data']
+            {
+            JL: Insert reasonable stuff here
+            }
```
BrianJKoopman, line 147: Add the `session.data` structure. This isn't standard anywhere, so needs to be documented so that when someone is writing a control program they know what is available.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            # So set the longest_subtask_duration to greater than the longest ever overiding task
+            # e.g. likely to be the stream task.
+            # No stream should be longer than, say, 12 hours so set to this
+            longest_subtask_duration = 12 * 60 * 60
+            while self._check_state:
+                if time.time() - last_release > 1.:
+                    last_release = time.time()
+                    if not self.lock.release_and_acquire(timeout=longest_subtask_duration):
+                        print(f"Could not re-acquire lock now held by {self.lock.job}.")
+                        return False
```
BrianJKoopman, line 175: This makes sense, but also I'm wondering if you can check state while a stream is happening. If you can, then you don't have to put a lock on this process and could just run it all the time.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+                # JL: Below we add the same json status string to both
+                # (1) the UKKID_feed (which you need to subscribe to or look at using the
+                # ocs-client-cli listen observatory.UKKIDController1.feeds.UKKID_feed
+                # command line line tool
+                # and
+                # (2) also the "session data" accessed by looking at
+                # client.check_state.status().session['data']
+                # from within the python agent interface
+                #  - not sure if this is appropriate - depends on the intended use of "session" data.
```
BrianJKoopman, line 187: The intention with `session.data` is to provide the 'latest' snapshot of data. So the full timestream goes to the feed, but the last point from the latest process loop goes into `session.data`. This facilitates control programs checking say the current temperature without needing to handle the whole timestream.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+            self.log.info("Getting RFSoc system information...")
+
+            now = time.time()
+            full_status = self.client.get_info()
+            json_status = json.dumps(full_status)
+
+            # Write result string to session.data, the feed and the log.
+            session.data = {"value": json_status,
+                            "timestamp": now}
```
BrianJKoopman, line 254: Should be documented in the docstring.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+            # Format message for publishing to Feed
+            message = {'block_name': 'status_string',
+                       'timestamp': now,
+                       'data': {'value': json_status}}
+            self.agent.publish_to_feed('UKKID_feed', message)
+
+            self.log.info(full_status)
```
BrianJKoopman, line 262: Similar to the above process, do you need to send the data to `session.data`, the feed, and the logs? We should decide what goes where.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            # Write result string to session.data, the feed and the log.
+            # Note, this also seems to turn up in the log.
+            session.data = {"value": json_status,
+                            "timestamp": now}
+
+            # Format message for publishing to Feed
+            message = {'block_name': 'status_string',
+                       'timestamp': now,
+                       'data': {'value': json_status}}
+            self.agent.publish_to_feed('UKKID_feed', message)
```
BrianJKoopman, line 313: If all that's coming back from this is a `dict` like: `'{"status": "success"}'`, then you should do a bit more to make this fit a data structure that makes sense for the various places you're sending it.

For the feed, it would make more sense to have something like:
```
message = {'block_name': 'status_string',
           'timestamp': now,
           'data': {'initialised': 1}}
```

It's more awkward to work with strings in Grafana. If 'initialised' is just a bit, you can plot it from the feed if it's an `int`.

For the session data the structure you have now is:
```
session.data = {"value": {"status": "success"},
                "timestamp": now}
```

But that means checking that like `session.data['value']['status']`. Flattening the dict to:
```
session.data = {"status": "success"},
                "timestamp": now}
```

would make more sense.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+            # Send stripped version to log.
+            self.log.info('initialise_server returned: ' + json_status_stripped)
```
BrianJKoopman, line 316: For the logs you might just make this:

```suggestion
            self.log.info('initialise_server returned: ' + json_status['status'])
```

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+            self.log.info("Sending RFSoc initialise_server...")
+
+            now = time.time()
+            full_status = self.client._initialise_server()
+            json_status = json.dumps(full_status)
+            # JL: This returns a string that looks like this
+            # '{"status": "success"}'
+            # However,  we get errors like this
+            # 2025-10-10T15-49-40.369218 Unable to format event {'log_logger': <Logger 'ocs.ocs_agent.OCSAgent'>, 'log_level': <LogLevel=info>, 'log_namespace': 'ocs.ocs_agent.OCSAgent', 'log_source': None, 'log_format': 'INITIALISE SERVER RETURNED JSON2:{"status": "success"}', 'log_time': 1760111380.36913, '                  #  message': (), 'time': 1760111380.36913, 'system': '-', 'format': '%(log_legacy)s', 'log_legacy': <twisted.logger._stdlib.StringifiableFromEvent object at 0x7aa852373c10>, 'isError': 0}: '"status"'
```
BrianJKoopman, line 297: This likely has to do with the formatting of your `self.log.info()` call. Check out the [txaio programming docs](https://txaio.readthedocs.io/en/latest/programming-guide.html#logging-1) for examples of the syntax.

I mess this up regularly too, since it differs from modern Python convention, but I think you're looking for something like:
```
self.log.info('initialise_server returned: {status}', status=json_status)
```

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            message = {'block_name': 'status_string',
+                       'timestamp': now,
+                       'data': {'value': json_status}}
+            self.agent.publish_to_feed('UKKID_feed', message)
+
+            # Send stripped version to log.
+            self.log.info('initialise_server returned: ' + json_status_stripped)
+            time.sleep(1)
+
+            return True, 'Sent initialise_server message to RFSocs'
```
BrianJKoopman, line 319: Does the status ever come back uninitialized? If so, check the return status when it comes back and do a:
```Python
return False, 'Failed to initialise the RFSoc'
```

socs/agents/ukkid_controller/ukkid_controller.py
BrianJKoopman, line 1: I'm skipping a bunch of docstring formatting comments for now, until the `README.md` gets converted to an `.rst` for Sphinx. Just a heads up.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+                # If we have found no suitable filename, throw error and return at this point.
+                if f_guess_filename is None:
+                    error_msg = 'No .json file with filename ' + prefix + '*.json found within directory ' + self.top_level_output_dir
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    message = {'block_name': 'det_res_freq_from_sweep_data_string',
+                               'timestamp': now,
+                               'data': {'value': error_msg}}
```
BrianJKoopman, line 389: This `message` block doesn't end up getting published to anything, since the next line returns with the error. Remove?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                        d = json.load(f)
+                        f_guess_list = d['f_guess_list']
+                except json.decoder.JSONDecodeError:
+                    error_msg = 'Problem with decoding json file: ' + params['f_guess_filename']
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    message = {'block_name': 'det_res_freq_from_sweep_data_string',
+                               'timestamp': now,
+                               'data': {'value': error_msg}}
```
BrianJKoopman, line 406: This `message` block doesn't end up getting published to anything, since the next line returns with the error. Remove?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                    return False, error_msg
+                except KeyError:
+                    error_msg = 'Expected key f_guess_list not found in json file: ' + params['f_guess_filename']
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    message = {'block_name': 'det_res_freq_from_sweep_data_string',
+                               'timestamp': now,
+                               'data': {'value': error_msg}}
```
BrianJKoopman, line 415: This `message` block doesn't end up getting published to anything, since the next line returns with the error. Remove?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                    return False, error_msg
```
BrianJKoopman, line 416: Great to see the error handling! You might also add one more block like:
```
except Exception as e:
```

For any unexpected error that comes up.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            self.log.info('Accurate resonant frequencies will be written out out to file ' + out_json_fname)
+            with open(out_json_fname, 'w', encoding='utf-8') as f:
+                json.dump(f_acc_dict, f, ensure_ascii=False, indent=4)
+            self.log.info('Accurate resonant frequencies written out to file ' + out_json_fname)
+
+            # Finally send data out to feed
+            feed_string = 'Guess freqs: ' + str(f_guess_list) + ' Accurate Freqs: ' + str(accurate_res.tolist())
+            now = time.time()
+            session.data = {"value": feed_string,
+                            "timestamp": now}
```
BrianJKoopman, line 461: I haven't run this yet myself, but is this `feed_string` potentially the full 200,000+ points from a full band sweep? It seems like it might be. That isn't really suitable for putting in the `session.data`.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            # JL TODO - makes this a json file.
+            self.log.info('Generating summary table...')
+            outfilename = out_full_path + '/' + 'fit_summary_table_' + self.kid_stream_id + '_' + now_string + '.txt'
+            table_str, summaries, filename = fit_summary_table(
+                fit_results,
+                outfilename=outfilename
+            )
+            self.log.info('...done.')
+
+            # JL TODO - makes this a json file.
```
BrianJKoopman, line 544: Looks like it is a json file? Remove old TODO comment.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            self.log.info('Generating histogram plots...')
+            outfilename = out_full_path + '/' + 'fit_histogram_plot_' + self.kid_stream_id + '_' + now_string + '.png'
+            fig_hist, axes_hist, histfile = fit_summary_histograms(
+                fit_results,
+                xscale_for={'Qi': 'log', 'Qc': 'log', 'Nonlinearity': 'log'},
+                outfilename=outfilename
+            )
+            self.log.info('...done.')
+            time.sleep(4)
+            self.agent.feeds['UKKID_feed'].flush_buffer()
```
BrianJKoopman, line 570: No need to sleep at the end of this task.

Also, nothing is published to the feed, so shouldn't need to flush the buffer at the end here either. Remove these two lines.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                       'data': {'value': feed_string}}
+            self.agent.publish_to_feed('UKKID_feed', message)
+
+            time.sleep(4)
+            self.agent.feeds['UKKID_feed'].flush_buffer()
+
+            return True, '"Determined accurate resonant frequencies from wideband sweep data.'
+
+    @ocs_agent.param("filename", default=None, type=str)
+    def fit_from_narrow_sweep(self, session, params=None):
```
BrianJKoopman, line 475: When does this get called in practice? Is it after every `narrow_band_sweep`? I'd suggest maybe just incorporating this into that task to run on the output `.csv` there. But maybe there is a reason you separated the two steps?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+            p = subprocess.Popen(executable_list, cwd=stream_script_path + '/client_scripts/', stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, shell=False)
+
+            self.log.info("stream receive process pid is " + str(p.pid))
+
+            # Communicate with the process for 2 seconds
+            # long enough for the process to send any obvious errors to stderr / stdout
+            try:
+                output, errors = p.communicate(timeout=2)
+            except BaseException:
```
BrianJKoopman, line 634: Use `Exception` instead of `BaseException`. Catching `BaseException` will prevent the agent from exiting properly if it needs to during this block. Even better would be to handle the expected [`TimeoutExpired` exception](https://docs.python.org/3/library/subprocess.html#subprocess.TimeoutExpired) explicitly, and have an `except Exception` for anything unexpected.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                    process_died = True
+
+                session.data = {"value": stream_status_string,
+                                "timestamp": now}
+
+                # Format message for publishing to Feed
+                message = {'block_name': 'stream_string',
+                           'timestamp': now,
+                           'data': {'value': stream_status_string}}
+                self.agent.publish_to_feed('UKKID_feed', message)
```
BrianJKoopman, line 675: We don't really need to publish this to a feed, in my opinion. This is mostly informative for the current state of the stream. I doubt people will care to have this archived in our TODs as a string forever. Correct me if I'm wrong.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+                # Also send this string to the log - discuss wether appropriate.
+                if not process_died:
+                    self.log.info(stream_status_string)
+                else:
+                    self.log.error(stream_status_string)
```
BrianJKoopman, line 681: Logging the status every 2 seconds is going to be quite verbose. So I probably wouldn't do that -- and that data is already available in the `session.data` (which doesn't clog up the logs). But logging the error string on unexpected return codes definitely makes sense.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+        # The streaming process should have concluded of its own accord after params['duration'] + TERMINATE_GRACE_PERIOD_SECONDS seconds.
+        # Check whether this is the case. If not, send SIGTERM to the process
+        # and send a warning that the streaming process needed to be manually SIGTERMed
+        self.log.info(f"Stream duration of {params['duration']} seconds has expired.")
+        ret_code = p.poll()
+        if ret_code is None:
+            self.log.warn("Streaming process unexpectedly still alive after the streaming duration has expired.")
+            self.log.warn("Will attempt to terminate - sending SIGTERM to the process...")
+            try:
+                os.kill(p.pid, signal.SIGTERM)
```
BrianJKoopman: This could be a `p.terminate()`. That sends a `SIGTERM` as well, and is maybe slightly more straightforward.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            except OSError as e:
+                self.log.error("Error received from os.kill:")
+                self.log.error(f"Failed to send signal: {os.strerror(e.errno)}")
```
BrianJKoopman, line 702: Consider sending a `SIGKILL` if `SIGTERM` doesn't work.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+        self.log.info('Disabling TCP/IP data streaming on RFSoc board.')
+        self.client.disable_stream()
+
+        self.log.info("Returning from stream function.")
+        message = {'block_name': 'stream_string',
+                   'timestamp': now,
+                   'data': {'value': "Streaming complete"}}
+
+        self.agent.publish_to_feed('UKKID_feed', message)
```
BrianJKoopman, line 716: Again, something to log and/or put in `session.data`. Doesn't really need to be published.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+        # JL: Presumably,I should return False here if process_died error or Error received from os.kill?
```
BrianJKoopman, line 718: Yes, certainly if you ran into an error and streaming failed in some way, `return False`. Control programs will see a failure status on the operation and can react accordingly.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+        return True, 'Streamed RFSoc data.'
+
+    def _abort_stream(self, session, params):
+        if session.status == 'running':
+            session.set_status('stopping')
+
+        if session.status != 'running':
+            return False, 'Aborted streaming RFSoc data.'
```
BrianJKoopman, line 726: This, by itself, won't abort the stream. The stream will just run as normal, but the session status will have changed to 'stopping'. You need to check this in the task and abort if this status change occurs for this to work. You could do that in the duration loop.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+    # Task to perform a wideband sweep across the full RF bandwidth.
+
```
BrianJKoopman, line 729: Drop redundant comment -- this is also in the docstring.
```suggestion
```

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            if not mock:
+                self.log.info("full_band_sweep: Will write sweep data to " + output_sweep_full_path)
+            else:
+                # Copy an example sweep file over to the output destination, then return from the function, without doing any real scan on the RFSoC.
+                self.log.info("MOCK: full_band_sweep: Will write sweep data to " + output_sweep_full_path)
+                loop_time_seconds = 10.0
+
+                shutil.copy('./example_full_band_sweep.csv', output_sweep_full_path)
+                self.log.info("MOCK: full_band_sweep: sweep data written to " + output_sweep_full_path)
```
BrianJKoopman, line 800: Ah, I see why you committed that file now. Still, I'd rather not commit a 50 MB `.csv` here. Is there a way we can generate a semi-realistic looking sweep? It'd be cheaper to write some code that generates a file than it would be to store the whole example file.

Open to other packaging ideas too.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                    sweep_status_string = "MOCK: Data has been sweeping for %.1f seconds." % (now - sweep_start_time)
+
+                    session.data = {"value": 'SESSION ' + sweep_status_string,
+                                    "timestamp": now}
+
+                    # Format message for publishing to Feed
+                    message = {'block_name': 'sweep_string',
+                               'timestamp': now,
+                               'data': {'value': 'FEED ' + sweep_status_string}}
+                    self.agent.publish_to_feed('UKKID_feed', message)
```
BrianJKoopman, line 819: Similar to other comments, this doesn't need to be published.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            phicat = np.angle(zcat)
+            slope = np.nanmedian(np.gradient(phicat,fcat))
+            zcat *= np.exp(-1j*(slope*fcat))
+
+            s['sweep_f'] = [fcat]
+            s['sweep_i'] = [np.real(zcat)]
+            s['sweep_q'] = [np.imag(zcat)]
+            s['sweep_ei'] = [np.ravel(s['sweep_ei'].T)]
+            s['sweep_eq'] = [np.ravel(s['sweep_eq'].T)]
+            '''
```
BrianJKoopman, line 1086: Same question here for this large block comment. Can we just drop this?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+            # SUSPECT???
```
BrianJKoopman, line 1088: ???

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                fig.supxlabel('Frequency (MHz)')
+                s1.set_ylabel('Power (dB)')
+                s2.set_ylabel('Phase (rad)')
+                s1.set_ylim(np.min(logmag),np.max(logmag))
+                s2.set_ylim(np.min(uphase),np.max(uphase))
+                plt.savefig(plot_filename,dpi=300)
+                self.log.info('full_band_sweep: Plot saved to ' + plot_filename)
+                '''
+
+            ################################################
```
BrianJKoopman, line 1162: More old code to drop?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+        self.log.info('full_band_sweep: Sweep complete.')
+        # JL: write something to the feed here?
+        return True, 'full_band_sweep: Sweep complete.'
+
+    def _abort_full_band_sweep(self, session, params):
+        if session.status == 'running':
+            session.set_status('stopping')
+
+        if session.status != 'running':
+            return False, 'Aborted full band sweep data.'
```
BrianJKoopman, line 1173: This doesn't really abort unless the task is built for it, as described above.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            sweep_start_time = time.time()
+            now_string = str(int(sweep_start_time))
+
+            # Make an output filename string
+            if params['filename'] is None:
+                output_sweep_filename = now_string + '_' + 'narrow_band_sweep' + '.csv'
+                output_sweep_plot_filename = now_string + '_' + 'narrow_band_sweep' + '.png'
+            else:
+                output_sweep_filename = params['filename']
+                output_sweep_plot_filename = params['filename'] + '.png'
```
BrianJKoopman, line 1227: Won't `output_sweep_filename` end in `.csv` or similar? So `output_sweep_plot_filename` is going to be `filename.csv.png`?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                # These example files derive from
+                # 1763660769_narrow_band_sweep.png
+                # and
+                # 1763660769_narrow_band_sweep.csv
+                shutil.copy('./example_narrow_band_sweep.csv', output_sweep_narrow_path)
+                self.log.info("narrow_band_sweep: sweep data written to " + output_sweep_narrow_path)
+                if plot_data:
+                    plot_filename = output_sweep_narrow_path.replace(filetype, '') + 'png'
+                    shutil.copy('./example_narrow_band_sweep.png', plot_filename)
+                    self.log.info("full_narrow_sweep: sweep data PNG file written to " + plot_filename)
```
BrianJKoopman, line 1250: Same thing here. Can we generate this semi-realistic narrow band sweep on the fly instead of loading it from the example file to avoid storing the large `.csv` in the package?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                    sweep_status_string = "Data has been sweeping for %.1f seconds." % (now - sweep_start_time)
+
+                    session.data = {"value": 'SESSION ' + sweep_status_string,
+                                    "timestamp": now}
+
+                    # Format message for publishing to Feed
+                    message = {'block_name': 'sweep_string',
+                               'timestamp': now,
+                               'data': {'value': 'FEED ' + sweep_status_string}}
+                    self.agent.publish_to_feed('UKKID_feed', message)
```
BrianJKoopman, line 1265: Don't publish this to the feed.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                # Send error message to feed, and log and return from function with error state.
+                stream_in_progress_error_msg = f'narrow_band_sweep: There is already a sweep in progress ({p * 100:.3f}%), please wait for it to finish before starting a new one.'
+                self.log.error(stream_in_progress_error_msg)
+                session.data = {"value": stream_in_progress_error_msg,
+                                "timestamp": now}
+                message = {'block_name': 'status_string',
+                           'timestamp': now,
+                           'data': {'value': stream_in_progress_error_msg}}
+                self.agent.publish_to_feed('UKKID_feed', message)
+                self.agent.feeds['UKKID_feed'].flush_buffer()
```
BrianJKoopman, line 1292: Don't publish this.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+
+                # If we have found no suitable filename, throw error and return at this point.
+                if f_accurate_filename is None:
+                    error_msg = 'No .json file with filename ' + prefix + '*.json found within directory ' + self.top_level_output_dir
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    message = {'block_name': 'narrow_band_sweep_string',
+                               'timestamp': now,
+                               'data': {'value': error_msg}}
```
BrianJKoopman, line 1330: This gets assembled to publish, but isn't published. We shouldn't publish it anyway. Remove.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                        d = json.load(f)
+                        f_accurate_list = d['f_accurate_list']
+                except json.decoder.JSONDecodeError:
+                    error_msg = 'Problem with decoding json file: ' + params['f_accurate_filename']
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    message = {'block_name': 'narrow_band_sweep_string',
+                               'timestamp': now,
+                               'data': {'value': error_msg}}
```
BrianJKoopman, line 1347: This gets assembled to publish, but isn't published. We shouldn't publish it anyway. Remove.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                    return False, error_msg
+                except KeyError:
+                    error_msg = 'Expected key f_accurate_list not found in json file: ' + params['f_accurate_filename']
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    message = {'block_name': 'narrow_band_sweep_string',
+                               'timestamp': now,
+                               'data': {'value': error_msg}}
```
BrianJKoopman, line 1356: This gets assembled to publish, but isn't published. We shouldn't publish it anyway. Remove.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                # Send error message to feed, and log and return from function with error state.
+                runtime_error_msg = str(e)
+                self.log.error(runtime_error_msg)
+                session.data = {"value": runtime_error_msg,
+                                "timestamp": now}
+                message = {'block_name': 'status_string',
+                           'timestamp': now,
+                           'data': {'value': runtime_error_msg}}
+                self.agent.publish_to_feed('UKKID_feed', message)
+                self.agent.feeds['UKKID_feed'].flush_buffer()
```
BrianJKoopman, line 1396: Don't publish this to the feed.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            num_points = params['num_points']
+            samples_per_point = params['samples_per_point']
+
+            self.client.perform_sweep(f_accurate_list, spans, num_points, samples_per_point, 'up')
+
+            # while self.client.get_server_status()['message']['latest_sweep_data_valid'] == False:
+            # while self.client.get_info(['server'])['server']['latest_sweep_data_valid'] == False:
+            while self.client.get_info(['server'])[0]['latest_sweep_data_valid'] is False:
+                self.log.info('Waiting for scan to complete, sleeping for 1 second.')
+                time.sleep(1)
```
BrianJKoopman, line 1410: What if the server never reports a valid sweep? Seems like we could get stuck here if there's an issue on the server end.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            self.log.info('narrow_band_sweep: Sweep complete.')
+            # JL write something to the feed here?
+            return True, 'narrow_band_sweep: Sweep complete.'
+
+    def _abort_narrow_band_sweep(self, session, params):
+        if session.status == 'running':
+            session.set_status('stopping')
+
+        if session.status != 'running':
+            return False, 'Aborted narrow band sweep data.'
```
BrianJKoopman, line 1459: Like the others, this aborter doesn't do anything unless the task handles the change in status appropriately.

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+                except KeyError:
+                    error_msg = 'Expected key f_accurate_list not found in json file: ' + params['f_accurate_filename']
+                    self.log.error(error_msg)
+                    session.data = {"value": error_msg,
+                                    "timestamp": now}
+                    return False, error_msg
+                self.log.info('Accurate Frequencies extracted from ' + f_accurate_filename)
+
+            # At this point. f_accurate_list should defined.
+            self.log.info('set_tone_frequencies: Will use tone frequencies of ' + str(f_accurate_list))
```
BrianJKoopman, line 1556: How long is this list in practice?

socs/agents/ukkid_controller/ukkid_controller.py
```diff
@@ -0,0 +1,1793 @@
...
+            self.log.info('set_tone_frequencies: Setting tone frequences on RFSoc.')
+            self.client.set_tone_frequencies(f_accurate_list)
+
+            self.log.info('set_tone_frequencies: complete.')
+            # JL write something to the feed here?
+            return True, 'set_tone_frequencies: complete.'
+
+    def get_tone_frequencies(self, session, params=None):
```
BrianJKoopman, line 1564: How are these 'get_X' tasks used in practice?

socs/agents/ukkid_controller/ukkid_controller.py
BrianJKoopman, line 1: I've left some of these as detailed comments below, but some general comments:
* You don't need to/shouldn't publish log style messages to the feed. Things like "Expected key f_guess_list not found in json file..." don't need to be kept forever in the .g3 file archive. That's more appropriate as a log message. They're useful when that message *just* occurred, but less so on data that's years old.
* You probably don't need to put those types of log messages into `session.data` either. Some sort of 'status' key/value like `'streaming: True'` makes sense, but long log messages fit better in the stdout log.
* There's a lot of 'go find the latest relevant file' code in this agent, which would probably make more sense to abstract away to either a 'driver module' or to the `souk_readout_tools` package. This doesn't have to happen now, but it would clean up the agent a bit to do it eventually.

BrianJKoopman (changes requested):
Thanks for PRing this @wenlockedge, and apologies it took so long to get to. Lots of comments below. I'm happy to help addressing some of them, like the packaging bits.

socs/agents/ukkid_controller/example_full_band_sweep.csv
wenlockedge: Hi,
 Yes, it makes sense to only allow actual code, rather than ancillary data files, in the repo. However, this file is required for the code to work properly in mocked mode. Do you have a recommendation for the best place for it can go such that it is publicly accessible and can be downloaded by someone following the install procedure in README.md? Something like a personal google drive/dropbox/ or git repo doesn't seem like the right solution...

socs/agents/ukkid_controller/example_narrow_band_sweep.csv
wenlockedge: See above.

socs/agents/ukkid_controller/example_full_band_sweep.csv
BrianJKoopman: My thought at the moment is a web server somewhere would be a good place. Either at a computing center, or I could host it at Yale.

socs/agents/ukkid_controller/README.md
wenlockedge: I have now removed references to dev/1.3.0 branch as those have been merged into main of souk_readout_tools, so we can use main for everything.

PR #1085, opened 2026-07-14
ACU: support commanding of HVAC systems

mhasself:
## Description

Adds a task to change states of Fans, Boosters, and Heater.

## Motivation and Context

Need to change these.  Due to changes in commanding (#1055), it's not easy to run shell scripts that command the ACU for this purpose, as the Agent effectively locks out other http command connections.
 
## How Has This Been Tested?

Tested extensively in simulator and also on the LAT a few weeks ago. Commanding this fields is somewhat unreliable, which is not completely understood; but the present changes are unlikely to degrade normal performance of the Agent, even if a few tweaks are going to be needed going forward.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/acu/agent.py
+++ b/socs/agents/acu/agent.py
@@ -3593,6 +3593,17 @@ class ACUAgent:
           targets: a list of targets.
           values: a list of values (or a single value, which will be broadcast).
 
+        Here are examples of target names:
+        - "Fan Yoke Traverse A Electronic Space"
+        - "Booster Yoke Traverse M Servo Space"
+        - "Heater" (to turn heater on/off)
+        - "Temperature Yoke Traverse" (the heater setpoint)
+
+        If a value matches "on" or "off", then it will be treated as a
+        request to turn the device on or off.  Otherwise, the value
+        will be interpreted as a "setpoint" specification (i.e. a fan
+        speed (%), or a thermostat temperature (degrees C).
+
         """
         error_count = 0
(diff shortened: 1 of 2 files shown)


BrianJKoopman:
Just a quick comment here that if you merge in the latest `main` (or rebase before review) that should fix the failing docs build.

socs/agents/acu/agent.py
```diff
@@ -3574,6 +3581,79 @@ def log(msg):
...
+
+        Args:
+          targets: a list of targets.
+          values: a list of values (or a single value, which will be broadcast).
+
+        Here are examples of target names:
+        - "Fan Yoke Traverse A Electronic Space"
+        - "Booster Yoke Traverse M Servo Space"
+        - "Heater" (to turn heater on/off)
+        - "Temperature Yoke Traverse" (the heater setpoint)
```
BrianJKoopman, line 3600: How does one usually get these target names if they aren't in this list? Or do HVAC experts just know them all?

BrianJKoopman (approved):
Looks good, just one question about the target names. Will wait until the group of #1085, #1086, and #1087 are all ready before merging.

socs/agents/acu/agent.py
mhasself: `ocs-web` knows them, and that's how they'll be applied 99.9% of the time.  They're parsed out of the HVAC dataset, which is included in the session.data for `monitor`.

PR #1093, opened 2026-07-22
wg-actuator: Increase initial movement distance to 10 mm

sadachi5:
Increase the initial movement distance from 5 to 10 mm to avoid failure after the initial movement due to the limit switch not turning off after the movement.

## Description

Increase the initial movement distance from 5 to 10 mm in `wiregrid_actuator`.
## How Has This Been Tested?

I did not test it. But it is a very small change, which should not have any issues.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/wiregrid_actuator/agent.py
+++ b/socs/agents/wiregrid_actuator/agent.py
@@ -176,7 +176,7 @@ class WiregridActuatorAgent:
                 return False, msg
 
         # Initial slow & small moving
-        ret, msg, LSonoff = move_func(5, speedrate=0.2)
+        ret, msg, LSonoff = move_func(10, speedrate=0.2)
         # Check the status of the initial moving
         if not ret:
             msg = '_insert_eject()[{}]: ERROR!: in the initail moving | {} '\


sadachi5:
@ykyohei Please approve it. This is to avoid recent sapt2 actuator issue.

ykyohei (approved):
Looks good to me. 
I think it would be nice to modify agent to be able to configure parameters so that we only need to change config file when we need adjustments, but fine for this time.

BrianJKoopman:
> I think it would be nice to modify agent to be able to configure parameters so that we only need to change config file when we need adjustments, but fine for this time.

+1 on the suggestion to make this configurable.

We're sure this won't negatively affect the other platforms' wiregrids?

sadachi5:
@BrianJKoopman Yes. This modification does not affect other sats negatively. 
In the insertion, the actuator moves a small distance as a first step (now 5mm -> 10mm) to confirm that the actuator works correctly.
This is much smaller than the actual actuator length to insert completely (~1m).
The distance of the small movement is not important.

PR #1102, opened 2026-09-01
PysmurfController: fix uxm_setup arg bug and default to modify_attens…

### LINKED ISSUE #1098: PysmurfController: typo likely prevents uxm_setup from running
https://github.com/simonsobs/socs/blob/d00c687bf0cb4048795836e2404aaccfbd618fc5/socs/agents/pysmurf_controller/agent.py#L489

Not sure how this could work.

mhasself:
## Description

In uxm_setup:

1. Typo bugfix.  Probably was broken.
2. Set a default value for modify_attens.  This is rarely needed.

## Motivation and Context

Running uxm_setup through OCS is desirable because then we can more easily enforce that PysmurfMonitor is running.  That's beyond the scope of this PR but here we at least help make it possible to do so.

Resolves #1098.

## How Has This Been Tested?

Untested, but unlikely to make anything worse than it was before.  And there's some chance it will work.

## Types of changes

## Checklist:

### DIFF
--- a/socs/agents/pysmurf_controller/agent.py
+++ b/socs/agents/pysmurf_controller/agent.py
@@ -486,7 +486,7 @@ class PysmurfController:
 
     @ocs_agent.param('bands', default=None)
     @ocs_agent.param('kwargs', default=None)
-    @ocs_agent.param('run_in_main_procsess', default=False)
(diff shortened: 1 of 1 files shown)


mhasself:
Draft while waiting for confirmation on best default args.  Also would be good if someone tested this somehow.

